PR #34, opened 2015-05-20
Mac

vick08:


jcsteh:
Thanks Vic. As I understand it, at present, this will break the Windows build because the path to db2val.h changes. I'm happy to tweak this myself later, but it might make more sense to just merge this along with the commits for the Mac build, as I'll probably have to tweak a few other things as well. Note also that I ended up putting the ifdefs for _WIN32 in master (but without the Mac code comments), as this way, I can make sure I ifdef Windows specific code as I develop, rather than forgetting and having to chase it later.

vick08:
Makes sense.

jcsteh:
I've actually just gone and made this change in master, as it'll be needed there eventually anyway.

PR #36, opened 2015-07-15
Support for VoiceOver on Mac OS X

vick08:
Adds a wrapper for NS accessibility API to enable posting notifications to AT. With this commit it is now possible to use OSARA on Mac OS X with VoiceOver.


jcsteh:
Thanks heaps Vic! I rebased this and tweaked it slightly so that it compiles again on Windows. :) Please double check that this still builds for you, but I don't think I could have broken anything.

PR #37, opened 2015-07-15
Eliminates Clang warnings regarding C++11 extensions

vick08:


jcsteh:
This -std=c++11 change is already in the NV Access OSARA master. I think you probably need to reset your branch to the NV Access master, as I've rebased your commits.

PR #97, opened 2016-10-31
Mac keymap

mulcahy1000:
This adds a separate keymap for Mac based on the current Windows keymap, and adds supports to the installation scripts to install the separate keymaps on each platform.


jcsteh:
Thanks so much for this. Sorry it took me so long to get to it.

I made a few tweaks myself before merging. See the last few commits on the branch if you're curious.

PR #163, opened 2018-05-29
Update readme.md

GarthHum:
Updated actions with keyboard shortcuts 
changed existing keyboard shortcuts inline with proposed key map
Fixed typo yu > you in two locations.
Added contributor

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -34,13 +34,13 @@ However, on Windows, OSARA uses Microsoft Active Accessibility (MSAA) to communi
 You can download the latest OSARA installer from the [OSARA Development Snapshots](https://osara.reaperaccessibility.com/snapshots/) page.
 
 ### Windows
-Once yu have downloaded the installer, simply run it and follow the instructions.
+Once you have downloaded the installer, simply run it and follow the instructions.
 
 Note that if you previously copied the OSARA extension into REAPER's program directory manually (before the installer became available), you must remove this first.
 The installer installs the extension into your user configuration, not the program directory.
 
 By default, the OSARA key map will be installed, completely replacing your existing key map.
-If yu do not wish this to occur, you can uncheck the "Replace existing key map with OSARA key map" option.
+If you do not wish this to occur, you can uncheck the "Replace existing key map with OSARA key map" option.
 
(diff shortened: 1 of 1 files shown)


jcsteh:
Thanks Garth! For future reference:

1. I tend to camel case key names containing multiple words (UpArrow, DownArrow) so they are pronounced better by speech synths. I made this change here.
2. You probably want to create a new branch based off master, rather than submitting a pull request from master itself. The other problem you're going to hit in future is that this branch needs to be based off the latest official master from me, not the copy in your fork, which will be out of date. Unfortunately, I don't see any way to sync your fork from the web interface, which kind of sucks.

GarthHum:
Thanks Jamie
Noted re the camel case
I will try to branch again rather than fork next time. I Forked this time as I couldn’t work out how to branch, thought it might be a permissions thing but it might be a web interface thing. 
Thanks again 
Garth 
 
 

On 5 Jun 2018, at 8:49 am, James Teh <notifications@github.com> wrote:

Thanks Garth! For future reference:

I tend to camel case key names containing multiple words (UpArrow, DownArrow) so they are pronounced better by speech synths. I made this change here.
You probably want to create a new branch based off master, rather than submitting a pull request from master itself. The other problem you're going to hit in future is that this branch needs to be based off the latest official master from me, not the copy in your fork, which will be out of date. Unfortunately, I don't see any way to sync your fork from the web interface, which kind of sucks.

jcsteh:
You need to fork and then create a branch within your fork. So forking was
the right thing to do; you can't do anything without creating a fork.
However, you then also need to branch. Think of a fork as "Garth's work",
then think of a branch as "Garth's work on this particular thing". :)

GarthHum:
Cheers Jamie 
Thanks for your patience 

On 7 Jun 2018, at 10:06 am, James Teh <notifications@github.com> wrote:

You need to fork and then create a branch within your fork. So forking was
the right thing to do; you can't do anything without creating a fork.
However, you then also need to branch. Think of a fork as "Garth's work",
then think of a branch as "Garth's work on this particular thing". :)

PR #178, opened 2018-10-31
Focus MIDI event filter dialog from actions which can open it

AZSlow3:
Based on window text and use Windows specific function. So Win32/English only.


src/midiEditorCommands.cpp
```diff
@@ -500,4 +500,23 @@ void cmdFocusNearestMidiEvent(Command* command) {
...
+	if ((GetWindowTextW(hWnd, szText, 128) != 0) &&
+		!wcsncmp(szText, L"Filter Events", 13)) {
+		*((HWND *)lParam) = hWnd;
+		return false;
+	}
+	return true;
+}
+
+void cmdMidiFilterWindow(Command *command) {
+	HWND editor = MIDIEditor_GetActive(), filter = NULL;
```
jcsteh: nit:

1. Please split this onto two separate lines, one variable per line.
2. Please use nullptr instead of NULL.

src/midiEditorCommands.cpp
```diff
@@ -500,4 +500,23 @@ void cmdFocusNearestMidiEvent(Command* command) {
...
+	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
+	// TODO: we could also check the command state was "off", to skip searching otherwise
+	EnumWindows(cbFindFilterWindow, (LPARAM)&filter);
```
jcsteh: I think you can simplify (and get rid of the callback) to this:

```
HWND filter = FindWindow(WC_DIALOG, L"Filter Events");
```

src/midiEditorCommands.cpp
```diff
@@ -500,4 +500,23 @@ void cmdFocusNearestMidiEvent(Command* command) {
...
+	if (filter && (filter != GetFocus()))
+		SetFocus(filter); // focus the window
```
jcsteh: nit: Please brace this. I know it's only a single line block, but bracing even single lines prevents hard to trace mistakes in future. Also, I'm aware a lot of existing OSARA code doesn't do this; this is a change I've made to my coding style recently. :)

src/midiEditorCommands.cpp
```diff
@@ -500,4 +500,23 @@ void cmdFocusNearestMidiEvent(Command* command) {
...
+}
```
jcsteh: nit: Blank line before this, since it now encloses multiple functions.

src/midiEditorCommands.cpp
AZSlow3: Done

src/midiEditorCommands.cpp
AZSlow3: That I have not understood...

src/midiEditorCommands.cpp
AZSlow3: I have not seen any nullptr in this file ;)
BTW: when compiling using default VS2015 settings, several  "deprecated" options are required, to avoid failures from indirect conversions.

src/midiEditorCommands.cpp
AZSlow3: Original idea was to make it correct way. There can be several MIDI editors open in REAPER, and so several open "Filter Events" dialogs. But I have not found a good way to detect which dialog is for which editor. And in this case you are right, FindWindow produce the same result without any extra code.

AZSlow3:
Thank you for detailed comments.

ScottChesworth:
Hey Alexey,

When testing your mod, I noticed an omission in the OSARA key map.
Currently, the F key is assigned to display the filter window in the
MIDI Editor, but that assignment hasn't been done in the MIDI Event
List Editor category. It should be assigned to "Filter: Show/hide
filter window..." ID 40762.

Is there a way to add that to your pull request without causing a huge
headache? If so, it'd help to cut back on the amount of questions
we're gonna get when this is released.

Cheers.

AZSlow3:
The shortcut is added, for Windows only

src/midiEditorCommands.cpp
jcsteh: Err, my comment was actually on the wrong line. Sorry. I meant please add a blank line before the #endif.

src/midiEditorCommands.cpp
jcsteh: > I have not seen any nullptr in this file ;)

That's correct. :) I've become a lot more familiar with modern C++ since I started working on OSARA. I'm ensuring newer code uses modern conventions, but some of the older code doesn't.

> BTW: when compiling using default VS2015 settings, several "deprecated" options are required, to avoid failures from indirect conversions.

Happy to take a PR if those changes are within OSARA. If the issues are in REAPER or WDL headers, though, that's out of scope for OSARA.

src/midiEditorCommands.cpp
jcsteh: I made the change myself. :)

PR #181, opened 2018-11-06
Bugfix for switching to plug-in window inside FX chain

AZSlow3:
Extended to float FX windows.

I have quite some thoughts about it.

Originally it was one character bugfix, 4 characters buffer was insufficient for "FX: " string including ending zero. So F6 could not work.

But I have extended to support floating FX windows. May be not so important before, Sibiac is working with screen images. And there are big plug-ins like AD2 which have hard time to fit into screen, especially at lower resolutions and when wrapped inside FX chain.

Finally, once F6 has started to work, I was confused that in case plug-in is already in focus, F6 start switching Solo. Not accessible plug-ins and in some cases Sibiac enabled plug-ins  produce no special audition. So people probably will try to press F6 more then once, and then get Solo switch. I have removed the focus change check and always consume F6. Since that is after FX check, I can not imagine the situation when it is unexpectedly consumed.

But... It can be practical to solo/unsolo the track when operating FXes. And default key for that is F6. So I propose to find better key for the focusing operation. I am not experienced to propose concrete  key, you know that better.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1189,25 +1189,37 @@ bool maybeSwitchToFxPluginWindow() {
 	) {
 		return false;
 	}
-	// Descend.
-	if (!(window = GetWindow(window, GW_CHILD)))
+	// Descend. Observed as the first or as the last.
+	if (!(window = FindWindowExA(window, nullptr, "#32770", nullptr))) {
 		return false;
+	}
 	// This is a property page containing the plugin window among other things.
-	if (!(window = GetWindow(window, GW_HWNDLAST)))
-		return false;
-	HWND dialog = window;
-	// Descend.
+	// set property page name, to avoid CPU/PDC label audition after switching
+	if (GetWindowText(window, name, sizeof(name)) == 0) {
+		SetWindowText(window, " ");
+	}
+	// Descend. Observed as the first or as the last. 
+	// Can not just search, we do not know the class nor name.
 	if (!(window = GetWindow(window, GW_CHILD)))
 		return false;
-	// This should get the plugin window.
-	if (!(window = GetWindow(window, GW_HWNDLAST)))
+	char classname[16];
+	if (!GetClassName(window, classname, sizeof(classname))) {
 		return false;
-	// set property page name, to avoid CPU label audition after switching
-	if (GetWindowText(dialog, name, sizeof(name)) == 0) {
-		SetWindowText(dialog, " ");
+	}
+	if (!strcmp(classname, "ComboBox")) {
+		// Plugin window should be the last.
+		if (!(window = GetWindow(window, GW_HWNDLAST))) {
+			return false;
+		}
+	} // else it is the first
+	// We have found plug-in window or its container
+	HWND plugin = window;
+	// if focus is already inside plug-in window, let F6 work as usual
+	HWND focus = GetFocus();
+	if ((focus == plugin) || (IsChild(plugin, focus))) {
+		return false;
 	}
 	// Try to focus the first child in Z order
-	HWND plugin = window; // remember plug-in window
 	HWND child;
 	while ((child = GetWindow(window, GW_CHILD))) {
 		window = child;
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1179,12 +1179,14 @@ bool maybeSwitchToFxPluginWindow() {
 	char name[8];
 	if (GetWindowText(window, name, sizeof(name)) == 0)
 		return false;
-	if (strncmp(name, "FX: ", 4) != 0 &&
+	if (strncmp(name, "FX: ", 4) != 0 && // FX chain window
+		// floating FX window, for different plug-in types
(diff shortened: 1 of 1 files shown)


jcsteh:
> Originally it was one character bugfix, 4 characters buffer was insufficient for "FX: " string including ending zero. So F6 could not work.

Oh. Oops. You're right, of course, but... I did test this and it previously worked. I don't understand how it could have worked at all.

> But... It can be practical to solo/unsolo the track when operating FXes. And default key for that is F6. So I propose to find better key for the focusing operation.

F6 is the normal keyboard command for switching to major panes in a window, which is why I chose it. Users are likely to try this (if they try anything at all) if they can't get to some area in a window. Choosing another key would be less intuitive, but I totally follow your concern about the conflict here. The key is documented anyway, so it's probably not too much of a concern. I'll wait to hear back from @ScottChesworth on this.

src/reaper_osara.cpp
```diff
@@ -1176,11 +1176,17 @@ void clickIoButton(MediaTrack* track, bool rightClick=false) {
 
 bool maybeSwitchToFxPluginWindow() {
 	HWND window = GetForegroundWindow();
-	char name[4];
+	char name[8];
 	if (GetWindowText(window, name, sizeof(name)) == 0)
 		return false;
-	if (strncmp(name, "FX: ", 4) != 0)
+	if (strncmp(name, "FX: ", 4) != 0 &&
+		strncmp(name, "DX: ", 4) != 0 &&
```
jcsteh: Can you please add a comment above this line noting that the following checks are for floating FX windows?

src/reaper_osara.cpp
```diff
@@ -1176,11 +1176,17 @@ void clickIoButton(MediaTrack* track, bool rightClick=false) {
...
+		strncmp(name, "VST: ", 5) != 0 &&
+		strncmp(name, "VSTi: ", 6) != 0 &&
+		strncmp(name, "VST3: ", 6) != 0 &&
+		strncmp(name, "VST3i: ", 7) != 0) {
```
jcsteh: nit: These if condition lines are indented to the same level as the code inside the if block. Please either indent them an extra level:

```
	if (strncmp(name, "FX: ", 4) != 0 &&
			...
			strncmp(name, "VST3i: ", 7) != 0) {
```

or keep them as they are but have the closing brackets at the same level as the if:

```
	if (strncmp(name, "FX: ", 4) != 0 &&
		...
		strncmp(name, "VST3i: ", 7) != 0
	) {
```

AZSlow3: Aligned.

AZSlow3:
During testing with iZotope, I have found the solution is not working perfectly. F6 is switching to the plug-in, but just to the Pane. I have added the code to focus deepest child. That works for iZotope, Sibiac and REAPER own plug-ins.

MatejGolian:
Hi,
I don't have anything technical to bring to the discussion, but I want to take this opportunity and thank all the contributors for keeping the Osara ball rolling. I think we all agree that it has become indispensable to us at this point and therefore I hope that its development will continue well into the future.
Special thanks goes out to Alexey for testing and improving features like this one.
I also have one question. Based on Alexey's last comment, do I understand correctly that this function does not guarantee  that one always ends up in the plugin's GUI and that focusing the GUIs of certain plugins may require specific coding?

AZSlow3:
The last change is a dirty hack... I personally is annoyed that every time the plug-in property page get focus, including F6, there was no label for that dialog and so the last available label is used for announcement.   That label has CPU consumption and PDC info. So I have added the space.

AZSlow3:
One more related comment. Sometimes the focus in FX dialogs is "stuck" on the dialog itself. Tab does not work then. Switching from/back to REAPER helps.
That should be fixed in REAPER, alternatively in OSARA.
At the moment I do this on NVDA/Sibiac level

AZSlow3:
Some problems spotted... please do not accept yet.

AZSlow3:
With latest changes, the mod is working fine. Still unclear with F6. While I have re-enabled it, OSARA auditions are rarely working when the focus is inside plug-ins. And so "Solo/Unsolo", while working, keep silence.

ScottChesworth:
FYI, if reassigning the key clears up this silent speech when
soloing/unsoloing issue, the consensus that users came to was to use
the G key. G as in GUI, I guess.

AZSlow3:
I have understood the reason for silence from F6. That should work fine with Sibiac 0.19b1 
The reason was in the way OSARA triggers audition and re-naming method used in NVDA app examples, that does not work together.

With G or any other normal keys, the implementation in OSARA will be more tricky. There are text field and drop box in the FX chain window, so people may want to use G as a part of text or search.

Since F6 should clearly announce what it does now, we can keep things as thy are.  I mean accept this PR in its current state.

ScottChesworth:
Confirmed, Sibiac 0.19b1 fixes the silence here. Awesome!

AZSlow3:
Current design is to do nothing in case focus is already inside plug-in or the window is closed after plug-in window is found.

PR #184, opened 2018-12-27
Note names and octave shift are used when audition notes in the MIDI editor

### LINKED ISSUE #183: MIDI Editor: Add Support for Named Notes
Hello,
I'm not sure if the current REAPER API allows implementing this or not, but it would be useful if support for reading note names could be added to Osara.
Currently loading note names from a file does not make any difference - Osara doesn't report the imported note names not even if one switches to the Named notes mode.
If my understanding of how this works is correct, custom note names (for example imported from note name maps) should be visible when switched to the Named notes view. I don't know how sighted people see it, but I think that it would be ideal if note names also appeared inside the event properties dialog. As far as I understand, the event properties dialog does not change no matter the mode and so this might be an all or nothing situation in a sense, because it could presumably result in the note names being there even in the editor modes that don't make use of them, but having the note names there is rather crucial in case of drum editing - it being probably the most common reason why'd one want to use named notes in the first place.

AZSlow3:
Fixes #183 .

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -202,13 +202,25 @@ void cmdMidiMoveCursor(Command* command) {
 		outputMessage(s);
 }
 
-const string getMidiNoteName(int pitch) {
+const string getMidiNoteName(MediaItem_Take *take, int pitch, int channel) {
 	static char* names[] = {"c", "c sharp", "d", "d sharp", "e", "f",
 		"f sharp", "g", "g sharp", "a", "a sharp", "b"};
-	int octave = pitch / 12 - 1;
-	pitch %= 12;
+	MediaTrack *track = GetMediaItemTake_Track(take);
+	int tracknumber = static_cast<int> (GetMediaTrackInfo_Value(track, "IP_TRACKNUMBER")); // one based
+	const char *note_name = GetTrackMIDINoteName(tracknumber - 1, pitch, channel); // track number is zero based
 	ostringstream s;
-	s << names[pitch] << " " << octave;
+	if (note_name) {
+		s << note_name;
+	} else {
+		int octave = pitch / 12 - 1;
+		int szOut = 0;
+		int *octave_offset = (int *)get_config_var("midioctoffs", &szOut);
+		if (octave_offset && (szOut == sizeof(int))) {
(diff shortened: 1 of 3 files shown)


src/midiEditorCommands.cpp
```diff
@@ -202,13 +202,25 @@ void cmdMidiMoveCursor(Command* command) {
...
 
-const string getMidiNoteName(int pitch) {
+const string getMidiNoteName(MediaItem_Take *take, int pitch, int channel) {
 	static char* names[] = {"c", "c sharp", "d", "d sharp", "e", "f",
 		"f sharp", "g", "g sharp", "a", "a sharp", "b"};
-	int octave = pitch / 12 - 1;
-	pitch %= 12;
+	MediaTrack *track = GetMediaItemTake_Track(take);
+	int tracknumber = static_cast<int> (GetMediaTrackInfo_Value(track, "IP_TRACKNUMBER")); // one based
+	const char *note_name = GetTrackMIDINoteName(tracknumber - 1, pitch, channel); // track number is zero based
```
jcsteh: nit: Asterisk after the type before the space and camel case the name:

`const char* noteName = ...`

src/midiEditorCommands.cpp
```diff
@@ -202,13 +202,25 @@ void cmdMidiMoveCursor(Command* command) {
...
 	ostringstream s;
-	s << names[pitch] << " " << octave;
+	if (note_name) {
+		s << note_name;
+	} else {
+		int octave = pitch / 12 - 1;
+		int szOut = 0;
+		int *octave_offset = (int *)get_config_var("midioctoffs", &szOut);
```
jcsteh: nit: Same as above for noteName. Also, just `(int*)` for the cast, no space before the asterisk.

src/midiEditorCommands.cpp
AZSlow3: Changes

src/midiEditorCommands.cpp
AZSlow3: Changed

AZSlow3:
I have unintentionally modified .gitignore and tried revert the change. Not sure I was successful, at least I have not managed to make the changes invisible in github...

PR #212, opened 2019-07-05
Cursor movement when pressing home / end on midi events listview

### LINKED ISSUE #211: Home / end keys don't work in midi events list
Currently, pressing home / end key doesn't move the cursor to the top / end of the midi events list.
(1) Make a midi item that contains 3 or 4 midi events
(2) Select the midi item and press ctrl+alt+e
(3) Press alt+3 to go to the events list
(4) Try pressing home /  end key
Any workarounds about this?
I built the latest master branch myself and confirmed this issue still existed.

yncat:
The keystrokes work and you can now quickly go to top / end as you would expect for the normal list views, but combining them with the shift key clears selection, rather than selecting everything inbetween. I have not figured out how to make it work.
Also, the keyboard hook callback had an if statement to return early, but it didn't seem to be working. I have fixed it, too.
This is my first pr. I apologize if I missed coding conventions or whatever for this project.

Fixes #211.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1242,7 +1242,11 @@ bool maybeSwitchToFxPluginWindow() {
 // Handle keyboard keys which can't be bound to actions.
 // REAPER's "accelerator" hook isn't enough because it doesn't get called in some windows.
 LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
-	if (code != HC_ACTION && wParam != VK_APPS && wParam != VK_RETURN && wParam != VK_F6) {
+	if (code != HC_ACTION){
+		// Return early if this event does not have key code info
+		return CallNextHookEx(NULL, code, wParam, lParam);
+	}
+	if(wParam != VK_APPS && wParam != VK_RETURN && wParam != VK_F6 && wParam != VK_HOME && wParam != VK_END && wParam != VK_SHIFT) {
 		// Return early if we're not interested in the key.
 		return CallNextHookEx(NULL, code, wParam, lParam);
 	}
@@ -1298,7 +1302,19 @@ LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
 	} else if (wParam == VK_F6 && !(lParam & 0x80000000)) {
 		if (maybeSwitchToFxPluginWindow())
 			return 1;
-	}
+	}else if(wParam == VK_HOME || wParam == VK_END || wParam == VK_SHIFT){
+		//Pressing home / end keys in midi events view moves the cursor to top / bottom respectively, by passing the needed key events to the listview
+		if(wcscmp(className, L"SysListView32") == 0){
+			HWND parent = GetParent(focus);
+			WCHAR parentClassName[22] = L"\0";
+			GetClassNameW(parent, parentClassName, ARRAYSIZE(parentClassName));
+			if(wcscmp(parentClassName, L"REAPERmidieditorwnd") == 0){
+				int msg = WM_KEYDOWN;
+				if((lParam >> 31) == 1) msg=WM_KEYUP;
+				SendMessage(focus, msg, wParam, 0);
+			}//if reaper midi editor window is foreground
+		}//If the focused on sysListView
+	}//if VK_HOME or VK_END
 	return CallNextHookEx(NULL, code, wParam, lParam);
 }
 HHOOK keyboardHook = NULL;


src/reaper_osara.cpp
```diff
@@ -1242,7 +1242,11 @@ bool maybeSwitchToFxPluginWindow() {
 // Handle keyboard keys which can't be bound to actions.
 // REAPER's "accelerator" hook isn't enough because it doesn't get called in some windows.
 LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
-	if (code != HC_ACTION && wParam != VK_APPS && wParam != VK_RETURN && wParam != VK_F6) {
```
jcsteh: Oops. The `code` check probably should have been an or, not an and. Anyway, the way you've done it is probably easier to read.

src/reaper_osara.cpp
```diff
@@ -1242,7 +1242,11 @@ bool maybeSwitchToFxPluginWindow() {
...
+	if (code != HC_ACTION){
+		// Return early if this event does not have key code info
```
jcsteh: This comment is not strictly correct. `code` can also be `HC_NOREMOVE`, which still provides key code info but isn't something we want to intercept.

src/reaper_osara.cpp
```diff
@@ -1298,7 +1302,19 @@ LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
 	} else if (wParam == VK_F6 && !(lParam & 0x80000000)) {
 		if (maybeSwitchToFxPluginWindow())
 			return 1;
-	}
+	}else if(wParam == VK_HOME || wParam == VK_END || wParam == VK_SHIFT){
```
jcsteh: Is it useful to pass shift at the moment given that it's not working as expected?

src/reaper_osara.cpp
```diff
@@ -1298,7 +1302,19 @@ LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
...
+		//Pressing home / end keys in midi events view moves the cursor to top / bottom respectively, by passing the needed key events to the listview
+		if(wcscmp(className, L"SysListView32") == 0){
+			HWND parent = GetParent(focus);
+			WCHAR parentClassName[22] = L"\0";
+			GetClassNameW(parent, parentClassName, ARRAYSIZE(parentClassName));
+			if(wcscmp(parentClassName, L"REAPERmidieditorwnd") == 0){
+				int msg = WM_KEYDOWN;
+				if((lParam >> 31) == 1) msg=WM_KEYUP;
+				SendMessage(focus, msg, wParam, 0);
```
jcsteh: So we're just sending the key message to the ListView? It'd be good if you can verify whether the home/end keys work in the events list without OSARA. I'm concerned that OSARA is actually the problem here and is preventing those keys from getting through. Otherwise, I guess REAPER's own key hook is eating home/end? Sounds like something that should ideally be fixed in REAPER. The Cockos devs are usually pretty responsive to reports like these.

src/reaper_osara.cpp
```diff
@@ -1298,7 +1302,19 @@ LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
...
+			}//if reaper midi editor window is foreground
```
jcsteh: If we're consuming the key, we should return 1 so it doesn't get passed further. We don't want to do this for shift, though, since that might be shift + something else.

src/reaper_osara.cpp
yncat: I'm going to fix the comment if we decide to continue this pull request.

src/reaper_osara.cpp
yncat: It might be something I forgot to delete after failing to make the "select to end" keystroke work. Currently, it is not necessary as you pointed out.

src/reaper_osara.cpp
yncat: I re-tested the following.
1) Checkout the latest master 94f67edff6081d5e1f892f4f969b5b5ae15c5886
2) build
3) home / end do not work
If further tests are needed, please let me know. I'm still not sure if this test is enough.
I don't know what is eating the key event, but something is definitely doing so.
I guess it is one of the parents at least, since directly sending the key event to the listview itself changes the cursor position.
Are we going to continue working on this PR? Or should we report this issue to Reaper devs? What do you think about this?

src/reaper_osara.cpp
yncat: Thanks for pointing that out! I will fix it if we will continue on this PR.

src/reaper_osara.cpp
jcsteh: Yeah, I guess that would indeed be tough. Perhaps just comment out the SetWindowsHookEx for the keyboard hook. I don't see how anything else in OSARA could be interfering here. You won't be able to use the applications key with that hook disabled, but you should be able to get into the MIDI editor with control+alt+e (I think; it's something like that).

jcsteh:
If we do determine this is a REAPER bug (and I suspect it is), I think it's worth reporting it to Cockos in the REAPER forum. You could try doing this yourself... or maybe @ScottChesworth might be interested in reporting it. If they don't want to fix it for whatever reason (maybe sighted users like this behaviour or something?), we can continue with this PR.

src/reaper_osara.cpp
yncat: Sure, thanks. I will post another comment after testing is done. Then, let's decide whether we report it to Cockos or not.

yncat:
I tested the keyboard event behavior. I think Reaper is eating home/end keys. Here's what I got:
1) Commented out SetWindowsHookEx and UnhookWindowsHookEx.
2) Reloaded Reaper with the compiled plugin.
3) Tried pressing home / end keys in the events list view in which about 10 items are listed. Nothing happened.
4) To confirm that the modified plugin is loaded, pressed the applications key on the track. As expected, nothing happened.

If @jcsteh agrees, I'm going to report this to Cockos with the PR's URL attached.

jcsteh:
Your testing seems sound to me. Please go ahead and report to Cockos.
Thanks.

yncat:
I've just posted about this issue.
https://forum.cockos.com/showthread.php?p=2197698#post2197698

yncat:
This PR is getting old and the problem itself should ideally be fixed from the Reaper side. For this reason, I'm going to close this.

PR #215, opened 2019-09-18
Fix reporting of volume and pan when automating

### LINKED ISSUE #112: Feedback for volume and pan when automating
Hi, 

As we've discussed privately, this might be more something to address to Cockos, but would it be possible for some or all of the following three things. Either to have a way to input exact values when automating volume and pan in in say latch preview. Currently, one can enter exact values for send volume in the routing dialogue but not track volume and pan. this would be the most preferable step for me. 

Or could Osara somehow provide feedback when nudging track volume and pan in latch preview. Currently, because of the way latch preview works, feedback stays locked on what the parameter value was before you changed automation modes. This makes changing volume from 0 to -inf very tedious though, which brings me to my third possibility. 

It's true that, while there's currently no feedback when controls are nudged, you can hear where things are going by ear, even if they move very slowly. Especially if the first method isn't possible, either by you or by Cockos, could we at least have a means to move automatable volume and pan straight to the extremes or to the middle, 0 db in volume's case, since I'm not sure if that is the precise middle of the slider, probably not. 

Hope this makes sense, 

justin

RDMurray:
The actions for nudge volume and pan, bound to alt+arrow keys, reported
the wrong value if the automation mode was anything but trim/read. this
commit reports the correct value.

Fixes #112.

### DIFF
--- a/src/osara.h
+++ b/src/osara.h
@@ -134,6 +134,7 @@
 #define REAPERAPI_WANT_EnumProjects
 #define REAPERAPI_WANT_GetProjectName
 #define REAPERAPI_WANT_plugin_register
+#define REAPERAPI_WANT_GetTrackUIVolPan
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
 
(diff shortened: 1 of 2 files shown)


jcsteh:
Thanks! Also, don't know what I did there to cause GitHub to lose your commit description in the merge commit... but it wasn't intentional. 😳

RDMurray:
Thank you.  Don't worry about it.  It was a very small change, so the commit message probably doesn't matter.

PR #216, opened 2019-09-19
Add reporting of global automation override

RDMurray:
Add feedback for the following actions that set global automation override:
Global automation override: All automation in latch mode
Global automation override: All automation in latch preview mode
Global automation override: All automation in read mode
Global automation override: All automation in touch mode
Global automation override: All automation in trim/read mode
Global automation override: All automation in write mode
Global automation override: Bypass all automation
Global automation override: No override (set automation modes per track)


src/reaper_osara.cpp
```diff
@@ -331,6 +331,31 @@ bool isItemSelected(MediaItem* item) {
 	return *(bool*)GetSetMediaItemInfo(item, "B_UISEL", NULL);
 }
 
+char* automationModeAsString(int mode) {
```
jcsteh: Please make this `const char*`. We definitely don't want callers being able to modify these static strings. :)

src/reaper_osara.cpp
```diff
@@ -331,6 +331,31 @@ bool isItemSelected(MediaItem* item) {
...
+	// this works for track automation mode and global automation override.
+	switch (mode) {
+	case -1:
+		return "no override";
```
jcsteh: Consider making this string shorter. Right now, this is going to result in an announcement like "Global automation override no override". I wonder if we could just say "none" here?

src/reaper_osara.cpp
```diff
@@ -331,6 +331,31 @@ bool isItemSelected(MediaItem* item) {
...
+	case 0:
+		return "Trim/Read";
```
jcsteh: Some of these messages are capitalised, others aren't. I'd prefer they were consistent. I guess lower case makes sense, since they're being appended to the end of a message.

src/reaper_osara.cpp
```diff
@@ -331,6 +331,31 @@ bool isItemSelected(MediaItem* item) {
...
+	case 1:
+		return "Read";
+	case 2:
+		return "Touch";
+		break;
```
jcsteh: nit: No need for break statement, since we're returning before it.

src/reaper_osara.cpp
```diff
@@ -869,6 +894,14 @@ void postCycleRecordMode(int command) {
 	}
 }
 
+void postChangeGlobalAutomationOverride(int command) {
+	ostringstream s;
+	s << "Global Automation Override ";
```
jcsteh: Consider shortening this. Maybe we could say "Global auto override" or just "Auto override" (unless you think it needs to be clearly differentiated by saying global)?

src/reaper_osara.cpp
```diff
@@ -869,6 +894,14 @@ void postCycleRecordMode(int command) {
...
+	s << automationModeAsString(GetGlobalAutomationOverride());
+	outputMessage(s);
+}
+
+
```
jcsteh: nit: Extraneous blank line.

src/reaper_osara.cpp
RDMurray: I don't know how I forgot that, sorry.

src/reaper_osara.cpp
RDMurray: Changed to "none"

src/reaper_osara.cpp
RDMurray: For some reason eloquence says "trim/red" if it is lower case.

src/reaper_osara.cpp
RDMurray: I've changed it to just "override"

src/reaper_osara.cpp
```diff
@@ -331,6 +332,30 @@ bool isItemSelected(MediaItem* item) {
 	return *(bool*)GetSetMediaItemInfo(item, "B_UISEL", NULL);
 }
 
+const char* automationModeAsString(int mode) {
+	// this works for track automation mode and global automation override.
+	switch (mode) {
+	case -1:
```
jcsteh: The entire block inside this switch statement needs to be indented one more tab. I'm pretty sure you had this right in your first patch... or maybe I missed it?

src/reaper_osara.cpp
```diff
@@ -273,7 +273,8 @@ const char* getFolderCompacting(MediaTrack* track) {
 }
 
 void reportActionName(int command, KbdSectionInfo* section=NULL, bool skipCategory=true) {
-	const char* name = kbd_getTextFromCmd(command, section);
+	
+	r* name = kbd_getTextFromCmd(command, section);
```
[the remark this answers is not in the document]
RDMurray: Wow, that is embarrassing!  Note to self: always compile before committing, oh and check the diff. :)

RDMurray:
Sorry about that, should all be fixed now.

Robbie

PR #218, opened 2019-09-26
Add action to cycle through midi recording modes on selected tracks

### LINKED ISSUE #188: Request: feedback and key map assignment for new SWS action "SWS/NF: Cycle through MIDI recording modes"
In SWS 2.10.0.1, they've also added an action to cycle through MIDI record modes as per my request. Can we get feedback for this and have it added to the key map?

Action: SWS/NF: Cycle through MIDI recording modes
ID: _NF_CYCLE_MIDI_RECORD_MODES
Key: Shift+\

Note: This key map assignment will overwrite "Track: Set track record mode to MIDI replace". once it's assigned, "Track: Set "track record mode to MIDI overdub" will no longer be needed (currently bound to the Backslash key).

Cheers.

RDMurray:
Add the action "OSARA: Cycle through midi recording modes of selected
tracks" which cycles through the midi recording modes, insuring that all
selected tracks have the same mode.

fixes #188.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -486,6 +486,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Cycle automation mode of selected tracks
 - OSARA: Report global / Track Automation Mode
 - OSARA: Toggle global automation override between latch preview and off
+- OSARA: Cycle through midi recording modes of selected tracks
 
 #### MIDI Event List Editor
 - OSARA: Focus event nearest edit cursor: control+f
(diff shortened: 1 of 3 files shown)


RDMurray:
rebased on upstream

RDMurray:
Okay, I made them all lower case.  

Btw the function recordingModeAsString() does not work for all possible recording modes.  It does work for all midi modes, which is all that it is used for at this time.  Are you happy leaving it like this at the moment or would you prefer me to find out the other modes? the Reaper API documentation is incomplete.

jcsteh:
> Btw the function recordingModeAsString() does not work for all possible recording modes. It does work for all midi modes, which is all that it is used for at this time. Are you happy leaving it like this at the moment or would you prefer me to find out the other modes? the Reaper API documentation is incomplete.

I'd prefer to have all the modes given the name of the function, but that's a bit tricky when the API documentation is incomplete. This is fine for now.

PR #220, opened 2019-10-18
Removed support for 32-bit architecture on Mac OS since it is no longer valid for 10.15 (Catalina)

vick08:


jcsteh:
Won't this break things for earlier versions of Mac OS? I'm concerned we might have Mac users using older versions who can't upgrade for whatever reason (e.g. outdated Mac).

vick08:
If they compile for 64-bit arch, then it should work on pre-Catalina versions of OS as well.
I do wonder if anyone is really using 32-bit version of Reaper on Mac OS anyway.
What do you think?
I don't know how to check Mac version from inside Scons.

jcsteh:
Cockos still offer a 32 bit build of REAPER, so I'm not comfortable
dropping this just yet.

To clarify, is the problem that Catalina won't even allow you to build 32
bit; i.e. it won't even cross-compile? That seems odd to me, since that
would cause problems for anyone trying to build apps for older versions of
Mac OS. Or does it fail to load universal binaries including 32 bit; i.e.
are current OSARA snapshots broken on Catalina?

Anyway, we can use the platform.mac_ver() function to get the mac OS
version:
import platform # near top of script
macReleaseStr = platform.mac_ver()[0]
if macReleaseStr:
    macReleaseNum = [int(part) for part in macReleaseStr.split(".")]
    if macReleaseNum >= (10, 15):
        # Catalina code here...

PR #222, opened 2019-10-23
Add feedback in the MIDI editor when moving to next/previous track/item

### LINKED ISSUE #160: Requesting action feedback when navigating to previous and next MIDI tracks and items from within the MIDI Editor and MIDI Event List Editor
Action feedback has been requested for the following actions. I can definitely see the UX and productivity benefits if these are possible to implement.

Here are the actions we're using (all in the MIDI Editor and MIDI Event List Editor categories):
Activate next MIDI track (40835)
Activate previous MIDI track (40836)
Custom: Go to next midi item (_75564b5383268f4e8701b2465bb5e198)
Custom: Go to previous midi item (3d82f1008d23dd4c9849c028325d3743)

Also just noticed when filing this request that in the MIDI Event List Editor category, the CTRL+Down and CTRL+Up keystrokes which should be bound to "Activate next MIDI track" and "Activate previous MIDI track" are incorrectly assigned. It's a carry over from the previous key map. Surprised that got missed in testing. Apologies!

RDMurray:
Add feedback for:
* Activate next MIDI track
* Activate previous MIDI track

Add the following new actions to replace the custom actions previously assigned to Ctrl+Left and Ctrl+Right:
* OSARA: Move to previous midi item on track
* OSARA: Move to next midi item on track

Fixes #160

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -506,7 +506,9 @@ void midiMoveToItem(int direction) {
 	HWND editor = MIDIEditor_GetActive();
 	MIDIEditor_OnCommand(editor, ((direction==1)?40798:40797)); // Contents: Activate next/previous MIDI media item on this track, clearing the editor first
 	MIDIEditor_OnCommand(editor, 40036); // View: Go to start of file
-	MIDIEditor_OnCommand(editor, NamedCommandLookup("_FNG_ME_SELECT_NOTES_NEAR_EDIT_CURSOR")); // SWS/FNG: Select notes nearest edit cursor
+	int cmd = NamedCommandLookup("_FNG_ME_SELECT_NOTES_NEAR_EDIT_CURSOR");
+	if(cmd>0)
+		MIDIEditor_OnCommand(editor, cmd); // SWS/FNG: Select notes nearest edit cursor
 	MediaItem_Take* take = MIDIEditor_GetTake(editor);
 	MediaItem* item = GetMediaItemTake_Item(take);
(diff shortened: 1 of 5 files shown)


RDMurray:
This fixes #160

src/midiEditorCommands.cpp
```diff
@@ -502,6 +502,65 @@ void cmdMidiSelectNotes(Command* command) {
 	outputMessage(s);
 }
 
+void midiMoveToItem(int direction) {
+	HWND editor = MIDIEditor_GetActive();
+	MIDIEditor_OnCommand(editor, ((direction==1)?40798:40797)); // Contents: Activate next/previous MIDI media item on this track, clearing the editor first
+	MIDIEditor_OnCommand(editor, 40036); // View: Go to start of file
+	MIDIEditor_OnCommand(editor, NamedCommandLookup("_FNG_ME_SELECT_NOTES_NEAR_EDIT_CURSOR")); // SWS/FNG: Select notes nearest edit cursor
```
jcsteh: While I realise the custom action relied on an SWS command and while it's generally recommended that SWS is installed, I try not to "break" things if the user doesn't have SWS installed. Can you confirm this doesn't crash or the like if SWS isn't installed? If this is a problem, working around this is probably as simple as checking the return from NamedCommandLookup and just not doing this part if it doesn't return an id. Unfortunately, the return value isn't explicitly documented for this case. I'm guessing 0, but it might also be -1.

RDMurray:
Okay, I'm checking for the returned value from NamedCommandLookup().  BTW it is 0 if the command does not exist.

PR #225, opened 2019-11-11
report position when moving item edges

RDMurray:


src/osara.h
```diff
@@ -159,6 +161,7 @@ typedef struct Command {
 	void (*execute)(Command*);
 } Command;
 extern int lastCommandRepeatCount;
+extern Command* lastCommand ;
```
jcsteh: I don't think this is needed in this patch? You're accessing lastCommand from within reaper_osara.cpp where it is defined.

src/reaper_osara.cpp
```diff
@@ -1713,6 +1713,48 @@ void cmdMoveItems(Command* command) {
...
+	// try to provide information based on the last item spoken by osara if it is selected
+	if (currentItem && ValidatePtr((void*)currentItem, "MediaItem*")
+		&& IsMediaItemSelected(currentItem)
+	) 
+	item = currentItem;
+	else if(CountSelectedMediaItems(0)>0)
+		item = GetSelectedMediaItem(0, 0);
+	else {
+		outputMessage("no items selected");
+		return;
```
jcsteh: I guess it's true that these commands always do nothing when no items are selected. Even so, I'm uncomfortable with not passing the command to REAPER. If this assumption or code is wrong in some way and the command does actually do something, this will completely prevent users from being able to do it. Of course, if they do it, it'll be reported incorrectly, but I think that's better than not passing the command at all. We can still return here; just pass the command to REAPER first.

src/reaper_osara.cpp
```diff
@@ -1713,6 +1713,48 @@ void cmdMoveItems(Command* command) {
...
+	}
+	const char* name;
```
jcsteh: This can be declared inside the `if` block below, since it's only used therein.

RDMurray:
Hi.  Okay, I made those changes.  I moved the declaration of lastCommandTime and lastCommandRepeatCount to near the top of reaper_osara.cpp along with lastCommand.  Is this okay?

Thanks

PR #227, opened 2019-11-11
Add feedback for time selection nudge commands

RDMurray:


src/reaper_osara.cpp
```diff
@@ -2261,6 +2261,25 @@ void cmdCycleMidiRecordingMode(Command* command) {
...
+	bool first= (lastCommand!=command);
+	double oldStart, oldEnd, newStart, newEnd;
+	GetSet_LoopTimeRange(false, false, &oldStart, &oldEnd, false);
+	Main_OnCommand(command->gaccel.accel.cmd, 0);
+	GetSet_LoopTimeRange(false, false, &newStart, &newEnd, false);
+	if(first) 
+		resetTimeCache();
+	ostringstream s;
+	if(newStart!=oldStart) {
+		s<<(first?"time selection start ":"");
```
jcsteh: 1. nit: Using a conditional expression here with an empty string as one of the possibilities makes this harder to read. Please use a normal if statement for this; i.e. don't output to `s` at all unless `first` is true.
2. nit: This won't matter any more when you fix 1), but I'd prefer spaces around symbols, particularly symbols like `?` and `:`.

PR #230, opened 2019-11-14
Add feedback when toggling item mute and solo

### LINKED ISSUE #172: requesting action feedback for item solo state
Right now, Osara supports item mute state, but not the solo state.
It would be very useful if it could provide the feedback for the Item properties: Toggle solo action, and say soloed where it currently says muted when going through the items.
Soloing items is something I do often, and it can currently be hard to find the soloed item if you forget which one it was that was soloed, or if you switched the focus from it, so I often have to either try to toggle solo on an item I think it was, but sometimes that doesn't work, and I have to go back and undo the solo.

RDMurray:
This reports the number of items muted and the number of items unmuted.  

It also adds feedback for item solo.  The problem with item solo is that it seams to be implemented by muting every other item in the project, not as a separate solo flag.  

Partial fix for #172 

Item solo state can't be reported as items don't appear to have a solo flag.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -894,8 +894,12 @@ void postToggleItemMute(int command) {
 		outputMessage(isItemMuted(GetSelectedMediaItem(0,0))?"muted":"unmuted");
 		return;
 	}
-	for (int i=0; i<count; i++)
-		isItemMuted(GetSelectedMediaItem(0, i))?muteCount++:unmuteCount++;
+	for (int i=0; i<count; ++i) {
+		if(isItemMuted(GetSelectedMediaItem(0, i)))
+			++muteCount;
+		else
+			++unmuteCount;
+	}
 	ostringstream s;
 	if(muteCount>0){
 		s<<muteCount<<" item"<<((muteCount==1)?"":"s") <<" muted";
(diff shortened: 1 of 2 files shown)


src/reaper_osara.cpp
```diff
@@ -871,12 +871,50 @@ void postRenameTrack(int command) {
...
+	int itemCount=CountMediaItems(0);
+	int selectedCount = CountSelectedMediaItems(0);
+	if(selectedCount==0)
 		return;
-	outputMessage(*(bool*)GetSetMediaItemInfo(item, "B_MUTE", NULL) ?
-		"muted" : "unmuted");
+	for(int i=0; i<itemCount; i++) {
+		MediaItem* item = GetMediaItem(0, i);
+		if((!isItemSelected(item)) && (!isItemMuted(item))) 
+			soloed=false;
```
jcsteh: I don't think there's any reason to continue the loop once we set this to false, so we should break here and skip pointless extra checks.

src/reaper_osara.cpp
```diff
@@ -871,12 +871,50 @@ void postRenameTrack(int command) {
...
+	}
```
jcsteh: I worry this might be slow in large projects with thousands of items. Still, I guess that might be better than no feedback. We'll have to wait and see if this is a problem in the real world.

src/reaper_osara.cpp
```diff
@@ -1033,6 +1071,7 @@ PostCommand POST_COMMANDS[] = {
 	{40118, postMoveEnvelopePoint}, // Item edit: Move items/envelope points down one track/a bit
 	{40696, postRenameTrack}, // Track: Rename last touched track
 	{40175, postToggleItemMute}, // Item properties: Toggle mute
+		{41561, postToggleItemSolo}, // Item properties: Toggle solo
```
jcsteh: Extraneous level of indentation.

src/reaper_osara.cpp
RDMurray: I tested it on a project with a ridiculous number of items, and it doesn't seam too slow.  everything else slows down about the same amount.

PR #251, opened 2020-02-28
When reviewing notes in the midi editor, play them with their actual length rather than an absolute length

### LINKED ISSUE #250: Midi editor: make note preview length reflect the length of the actual note
The osara note preview is now around 300ms. It would be helpful if the preview is equal to the actual length. Should be possible to calculate that based on bpm and note length.

Of course, chord previews would be more difficult, as notes in chords could have multiple lengths. I assume using the shortest length here would be fine. Alternatively, we need a done timer per note that turns off that particular note, but that adds much complexity with minimal effort.

Related to #249

leonardder:
When playing notes when reviewing them in the midi editor, they were played for a total time of 300ms. This pr fixes #250 by doing the following:

* Note start and end can be saved on the MidiNote struct. It also has a const method to calculate the length, simply by doing end - start.
* When playing chords or notes, we calculate start and end. For chords, we calculate the minimum length for the notes in the chord

### Known issue
When changing the pitch cursor or inserting a note, we don't have a length. Therefore, use the default length of 0.3ms. I assume we should be able to fix this for inserting a note as the note is inserted when we preview it, but we need a way to get the actual note. For pitch cursor, there's no way to do this.


leonardder:
I figured that when you can hear the length of the notes, and you have absolute pitch, reported information is mostly redundant. So, I added config parameters to enable disable reporting/previewing of notes as well. @jcsteh If you think that's a separate subject that needs to go in a separate pr, I can do that.

src/midiEditorCommands.cpp
```diff
@@ -455,6 +484,9 @@ void cmdMidiMovePitchCursor(Command* command) {
 	int chan = MIDIEditor_GetSetting_int(editor, "default_note_chan");
 	int vel = MIDIEditor_GetSetting_int(editor, "default_note_vel");
 	previewNotes(take, {{chan, pitch, vel}});
```
jcsteh: Perhaps we could use the configured insertion length as the preview length for the pitch cursor? We don't have to do that in this PR, though.

src/midiEditorCommands.cpp
```diff
@@ -455,6 +484,9 @@ void cmdMidiMovePitchCursor(Command* command) {
...
+	if (!shouldReportNotes) {
+		return;
```
jcsteh: nit: I'm a big fan of early returns where it makes things simpler, but in this case, I think it's more readable to reverse this and put the outputMessage inside the if block.

src/reaper_osara.cpp
```diff
@@ -2509,11 +2509,16 @@ map<pair<int, int>, Command*> commandsMap;
 
 const char CONFIG_SECTION[] = "osara";
 
+extern bool shouldReportNotes;
+extern bool shouldPreviewNotes;
```
jcsteh: I'd ideally prefer these options were split into a separate PR, but it's not a big deal and it's probably a bit of work to split it out now, so I'll live with it. :)

I can understand the desire to turn off note reporting. That way, you can use speech for note position and audio for the notes themselves, increasing efficiency. However, I'm not convinced we have a use case for disabling preview? Configurability can be a good thing, but only if every single option adds real value. Otherwise, it adds unnecessary UI complexity.

I'm also concerned about the possibility of someone disabling both note preview and note reporting. That seems like something the UI shouldn't allow. We'd have to use a radio button to avoid that, though: Indicate midi notes by: reporting/previewing/both... or similar.

src/midiEditorCommands.cpp
```diff
@@ -18,17 +18,30 @@
...
 typedef struct {
 	int channel;
 	int pitch;
 	int velocity;
 	int index;
+	double start;
+	double end;
+	double GetLength() const {
+		double length = end - start;
+		if (length <= 0) {
```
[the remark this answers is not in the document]
leonardder: It could be negative when providing a higher start than end. That should never happen, but just in case as you say, better safe than sorry. Note that SetTimer doesn't take negative values and therefore the interval of the timer would be uncertain.

src/midiEditorCommands.cpp
leonardder: I looked into that, but the configured insertion length is in midi ticks (ppq), and we can only convert a start and end to project time values, converting a length fails. This is what I tried:

* Converting the length from ppq to time directly
* Getting start as the cursor position, convert it to ppq, add the length ppq and convert the end back to project time

All resulted in a too short preview, probably around USER_TIMER_MINIMUM (0x0000000A) in length. If you have alternative suggestions, I'd be glad to know them.

src/reaper_osara.cpp
leonardder: I think you're true. I also noticed that Reaper also has its own note previewing in some cases, having Osara disable note preview and then Reaper preview notes would be confusing.
I removed the option for now. I'm a big fan of disabling reporting, though.

src/reaper_osara.cpp
```diff
@@ -2544,12 +2549,14 @@ INT_PTR CALLBACK config_dialogProc(HWND dialog, UINT msg, WPARAM wParam, LPARAM
 	return FALSE;
 }
 
+
```
jcsteh:
```suggestion
```

src/midiEditorCommands.cpp
jcsteh: I can understand why converting the raw length might not work, but the second surprises me. I think we'd need to get a look at the raw values we're getting back from each function to know what's going on there.

Is the insertion length from MIDIEditor_GetSetting_int even correct?

Failing that, I wonder whether MIDI_GetPPQPosFromProjTime expects a time relative to the take or something like that? That would make the two functions asymmetric, but this wouldn't be the first time I've been confused by REAPER's lack of documentation/consistency on the time base for various functions.

Alternatively, we could convert to qn time and then back, which is absolute to the project instead of per take. We'd need to calculate the length in quarter notes ourselves. Instead of using MIDIEditor_GetSetting_int, we could use GetToggleCommandState with the various insertion length action ids, converting 1/4 to 1 qn, 1/2 to 2 qn, etc.

leonardder: I'm afraid that whatever we do that relates to project time will at least always fail if it is the first note we're inserting, as the project is empty at that point.

ScottChesworth:
Hi @leonardder , just tested the new preview and reporting, great stuff! One minor UX nitpick: at the moment, disabling the new "Report MIDI notes in MIDI editor" option also disables reporting of note length. Was that by design? If so, it would be good to have that reflected in the wording of the new check box, perhaps something like "Report notes and note lengths in piano roll" would add clarity?

jcsteh:
I don't think that's necessarily true for projec time (as opposed to times
relative to a take). Certainly, when a project is empty, you can still move
to bar 10 and get a valid time in measures/beats.

Issue #252, opened 2020-03-02
Add feedback to several midi editor note parameter actions


leonardder:
I propose adding feedback to the following actions:

* Edit: Lengthen/Shorten notes one grid unit
* Edit: Move notes up/down one semitone/octave
* Edit: Note velocity +-01/10

The problem with these actions is that we need to know the note to which the action was applied. As far as I know, there is no reliable way to find out the current note, especially when the cursor is at a chord. @jcsteh: any ideas how to handle this?
Note that some of these commands already have note preview, therefore I propose additional reporting feedback for them that can be toggled with the new report note feedback option.

jcsteh:
I think those commands operate on selected notes and OSARA selects notes
when you move to a chord or note. So, you should be able to report based on
selected notes.

ScottChesworth:
This seems to have been implemented already, so closing the issue. Feel free to reopen if that's incorrect.

PR #254, opened 2020-03-03
Respect note length when previewing inserted notes

leonardder:
Follow up of #251 
Starting point for #252 

When inserting a note, we now do the following after inserting:

1. Get the selected notes in the current take. This will be the new note, along with possible other notes. If you selected a chord before inserting a note, the chord will be selected including the new note.
2. As you couldn't insert multiple notes with the same pitch, pitch is something that makes a note unique. We also know the pitch at point of inserting. Therefore, we get the newly inserted note from the selected notes by comparing it with the current pitch.
3. Then, the inserted note is played. As a bonus, it will be announced when reporting notes is on.

This pr will be the fundament of other subsequent pull requests that report action feedback for several actions that act on selected notes (#252)

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -521,7 +521,7 @@ void cmdMidiInsertNote(Command* command) {
 	// Play the inserted note.
 	previewNotes(take, {note});
 	ostringstream s;
-		if (shouldReportNotes) {
+	if (shouldReportNotes) {
 		s << getMidiNoteName(take, note.pitch, note.channel) << " ";
 		s << formatTime(note.getLength(), TF_MEASURE, true, false, false);
 		s << ", ";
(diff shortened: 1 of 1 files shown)


src/midiEditorCommands.cpp
```diff
@@ -492,12 +510,24 @@ void cmdMidiInsertNote(Command* command) {
...
+	vector<MidiNote> selectedNotes = getSelectedNotes(take);
+	// Find the just inserted note based on its pitch, as that makes it unique.
+	auto note = *(find_if(
+		selectedNotes.begin(), selectedNotes.end(),
+		[pitch](MidiNote n) { return n.pitch == pitch; })
+	);
+	// Play the inserted note.
+	previewNotes(take, {note});
+	ostringstream s;
+		if (shouldReportNotes) {
```
jcsteh: nit: Extraneous tab.
```suggestion
	if (shouldReportNotes) {
```

jcsteh:
Hmm. GitHub seems to have messed up the authorship info with this merge, perhaps because I committed a nit fix. Sorry about that.

leonardder:
No problem. I think it could also have to do with my work e-mail ending up in one of te commits.

PR #260, opened 2020-03-11
Add new Osara actions: Move to next/previous CC

leonardder:
This pr is the first in a series of prs that will improve editing of control changes. This first pull request adds four new actions:

- OSARA: Move to next CC
- OSARA: Move to previous CC
- OSARA: Move to next CC and add to selection
- OSARA: Move to previous CC and add to selection

These actions are currently unbound, though I suggest the following:

- OSARA: Move to next CC: '
- OSARA: Move to previous CC: ;
- OSARA: Move to next CC and add to selection: shift+'
- OSARA: Move to previous CC and add to selection: shift+;


src/midiEditorCommands.cpp
```diff
@@ -546,11 +648,149 @@ void postMidiSelectNotes(int command) {
...
+		{67, "Soft Pedal (on/off)"},
+		{68, "Legato Pedal (on/off)"},
+		{69, "Hold 2 Pedal (on/off)"},
+		{70, "Sound Variation"},	
+		{71, "Timbre/Resonance"},
+		{72, "Sound Release"},
+		{73, "Sound Attack"},
+		{74, "Brightness/Cutoff Freq"},
+		{75, "Sound Control 6"},
+		{76, "Sound Control 7"},
```
jcsteh:
```suggestion
		{76, "Sound Control 7"},
```

src/midiEditorCommands.cpp
```diff
@@ -357,6 +372,74 @@ vector<MidiNote> getSelectedNotes(MediaItem_Take* take, int offset=-1) {
...
+	MIDI_SetCC(take, cc, &select, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
+}
+
+bool isCCSelected(MediaItem_Take* take, const int cc) {
+	bool sel;
+	MIDI_GetCC(take, cc, &sel, NULL, NULL, NULL, NULL, NULL, NULL);
+	return sel;
+}
+
+vector<MidiControlChange> getSelectedCCs(MediaItem_Take* take, int offset=-1) {
```
jcsteh: 1. This doesn't seem to be used in this PR. Can we move it into another?
2. What's the use case for the offset argument? I just saw that getSelectedNotes has it as well, though I didn't think to ask this question in that review. 😳

src/midiEditorCommands.cpp
```diff
@@ -11,14 +11,24 @@
...
+typedef struct {
+	int channel;
+	int index;
+	int control;
+	int value;
+	double position;
+} MidiControlChange;
+
 const double DEFAULT_PREVIEW_LENGTH = 0.3;
+extern FakeFocus fakeFocus;
```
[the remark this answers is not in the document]
leonardder: Must have missed that. Will remove it.

src/midiEditorCommands.cpp
leonardder: > This doesn't seem to be used in this PR. Can we move it into another?

Sure, will probably need it in a subsequent pr but will introduce it there.

> What's the use case for the offset argument? I just saw that getSelectedNotes has it as well, though I didn't think to ask this question in that review. 😳

It allows you to start searching for selected notes from a certain note, leaving notes before out of scope. I'm currently not sure about the use case though, I just decided to expose it as an optional parameter instead of a constant.

src/midiEditorCommands.cpp
```diff
@@ -357,6 +372,74 @@ vector<MidiNote> getSelectedNotes(MediaItem_Take* take, int offset=-1) {
...
+			if (start < 0 || start >= count) {
+				// There's no adja	cent item in this direction,
+				// so move to the current one again.
+				start = currentCC;
+			}
+		}
+	}
+	bool found = false;
+	int chan, control, value;
+	for (; 0 <= start && start < count; start += direction) {
```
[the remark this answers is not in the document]
leonardder: Ugh, sorry, you're correct. Should be ok now.

PR #271, opened 2020-05-24
Add action OSARA: Select all notes with the same pitch within time se…

RDMurray:
Adds an action "OSARA: Select all notes with the same pitch starting in time selection".  

This action has been requested, but I was a bit hesitant to do it as it is adding reaper functionality that is not strictly to do with accessibility. Sorry @ScottChesworth and @GianlucaApollaro, 6 months is more than a bit hesitant.:) However I suspect that selecting multiple notes is much easier with the mouse, so maybe this action does belong in Osara.  What do you think @jcsteh ? There are other actions in Osara for selecting notes.

Another thing that I am unsure about is weather to make this consistent with the action "Edit: Select all notes in time selection" which is currently on the keymap, or "Edit: Select all notes starting in time selection".  I think the latter makes more sense, so that is how I have implemented it.  I welcome comments regarding which way everyone thinks would be most useful.

I could of course do it both ways if they are both considered useful. 

in case anyone is unsure of the difference, "Edit: Select all notes in time selection" will include notes that are positioned before the start of the time selection if they are long enough to end within the time selection.  I think this is surprising behaviour.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -568,6 +568,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Move to previous CC
 - OSARA: Move to next CC and add to selection
 - OSARA: Move to previous CC and add to selection
+- OSARA: Select all notes with the same pitch starting in time selection
 
 #### MIDI Event List Editor
 - OSARA: Focus event nearest edit cursor: control+f
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -855,6 +855,42 @@ void cmdMidiMoveToTrack(Command* command) {
 	outputMessage(s);
 }
 
+void cmdMidiSelectSamePitchStartingInTimeSelection(Command* command) {
+	double tsStart,tsEnd;
+	GetSet_LoopTimeRange(false, false, &tsStart, &tsEnd, false);
+	if(tsStart == tsEnd) {
+		outputMessage("No time selection");
+		return;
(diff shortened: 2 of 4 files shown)


RDMurray:
Sorry, just noticed a bug, I'll update the branch asap.

ScottChesworth:
I think the behaviour of "Edit: Select all notes starting in time selection" makes more sense when using the MIDI Editor with a screen reader. I'll float switching to that action in the key map group shortly.

jcsteh:
What's the real world scenario for this?

ScottChesworth:
My real world scenario is efficiency when tidying up MIDI drums. Quantizing just the footwork within a fill wrapped in a time selection for example. Prior to this action, there was no way to do that without isolating the section as its own item, which (given Reaper's issues with note offs around MIDI item splits) was a cumbersome workflow.

I know Robbie's action extends Reaper functionality and typically that's not the goal for OSARA, but having observed sighted folk editing MIDI in the piano roll, I think his argument earlier in the thread holds up. This is a workflow which is way more efficient with a mouse. The action simply serves to level the playing field.

If you're not into the idea of incorporating this into OSARA, could Robbie's code for the action be submitted to Cockos for their consideration? Like, are actions being written the same way they do? I know nothing about how this bit of Reaper works.

ScottChesworth:
Lovely stuff. Cheers.

PR #272, opened 2020-05-24
Add feedback for tempo related actions. fixes: #268

### LINKED ISSUE #268: Requesting action feedback for 3 actions related to adjusting project bPM on the fly
In an upcoming key map release with significant improvements, we'd like to map the following three actions which have been suggested by members of the Reaper Access WhatsApp channel. It would be super useful if they could provide action feedback, and would also negate the need for #265.

Action: Tempo: Decrease current project tempo 01 BPM
ID: 41130

Action: Tempo: Increase current project tempo 01 BPM
ID: 41129

Action: SWS/BR: Options - Toggle "Playback position follows project timebase when changing tempo"
ID: _BR_OPTIONS_PLAYBACK_TEMPO_CHANGE

FYI they're probably gonna be assigned to Alt+-, Alt+= and Alt+Backspace respectively.

Thanks in advance.

RDMurray:
- Tempo: Decrease current project tempo 01 BPM
- Tempo: Increase current project tempo 01 BPM
- SWS/BR: Options - Toggle "Playback position follows project timebase when changing tempo"

Fixes #268.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1043,6 +1043,20 @@ void postTakeChannelMode(int command) {
 	outputMessage(s);
 }
 
+void postChangeTempo(int command) {
+	double tempo = Master_GetTempo();
+	ostringstream s;
+	s << tempo << " BPM";
+	outputMessage(s);
+}
+
+void postTogglePlaybackPositionFollowsTimebase(int command) {
+	ostringstream s;
+	s << ( GetToggleCommandState(command) ? "Enabled" : "Disabled" );
+	s << " playback position follows project timebase when changing tempo" ;
+	outputMessage(s);
+}
+
 typedef void (*PostCommandExecute)(int);
 typedef struct PostCommand {
 	int cmd;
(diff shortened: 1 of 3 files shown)


src/reaper_osara.cpp
```diff
@@ -1010,6 +1010,20 @@ void postToggleCountIn(int command) {
...
+void postChangeTempo(int command) {
+	double tempo = Master_GetTempo();
+	ostringstream s;
+	s << tempo << " BPM";
+	outputMessage(s);
+}
+
+void postTogglePlaybackPositionFollowsTimebase(int command) {
+	ostringstream s;
+	s << "Playback position follows project timebase when changing tempo: " ;
```
jcsteh, line 1022: I worry about efficiency here: this is a very long message to hear before hearing the new state. It's true that OSARA normally says the name before the state; e.g. "repeat on", "metronome on". However, this one really is very long. If this is a command that might be used often, that could be a big efficiency killer.

One way of getting around this could be to say "enabled playback position follows project timebase when changing tempo" (or "disabled" if turning it off). That's a bit inconsistent with some other messages, but there's some prior art; e.g. we say "unmuted" instead of "mute off".

RDMurray: I agree, it's done.

PR #279, opened 2020-07-02
Change action id for "Item: Paste items/tracks" to match change in Re…

RDMurray:
The action id for "Item: Paste items/tracks" in Reaper has changed, causing Osara to report nothing when pasting.


ScottChesworth:
My hunch is that most folk are up to some flavour of 6 now, but I obviously don't know everyone's setup in detail, so I've polled RWP and WhatsApp asking to hear from anyone who's still using 5.9x. Gimme a couple of days and I'll report back.

jcsteh:
Just to be clear, if the only possible fix here was to drop 5.x support,
I'd just do it. However, there is another possible fix which doesn't
detriment 6.x users, hence the question.

RDMurray:
It's only one line, and Reaper does still have the old action, so I think it would be ok to keep support for the old action.

jcsteh:
Oh. I wonder why REAPEr added a new action then?

If that's the case, let's just go with adding back the old action as well.

RDMurray:
It behaves differently when there are hidden tracks. I don’t know the exact details. The new one is the default paste action in recent versions.

Issue #141, opened 2018-01-07
Auto Item Navigation gets stuck


Justinmac84:
Hi, 

Pretty much what the subject says. In a project with multiple automation items, you can't navigate between them reliably with control cursor left and right as this project hopefully demonstrates. 

[Auto Item Demo.rpp.zip](https://github.com/jcsteh/osara/files/4944946/Auto.Item.Demo.rpp.zip)

Thanks for looking at all of these, 

Justin

Aleskis:
The automation items are really powerful to manage pieces of envelopes and easily create some kind of LFOs. Would it be possible to investigate the blocking bug while browsing and submit a bug report to Cockos if there is a problem in REAPER? Over time, this feature has been greatly enriched. It would be great to use it easily.

ScottChesworth:
In the comments of #207, @jcsteh thought this might be due to a Reaper bug. Dunno if he's had time to dig into that any deeper, but I'm hesitant to take it to Cockos without confirmation that Reaper itself is the issue (in part because I don't use the feature here, so I'd be useless at anything beyond just outlining the issue). I'm not the only means of getting stuff seen by Cockos though. If you actually use the feature @Aleskis , try your luck in the bug reports forum at https://forum.cockos.com/forumdisplay.php?f=22. Remember that hardly anybody there will be using the OSARA key map, so it's best to use action names instead of keystrokes when writing reproduction steps.

Aleskis:
I comment this ticket, because #207 is a duplicate of this one. Before posting on Cockos's forum, we need more technical information on the issue and ensure it's not a OSARA bug.

ScottChesworth:
I won't be able to shed any light on technical information, but I think I have a way to check whether it's a Reaper bug. Unfortunately it requires some sighted assistance. Hopefully, I'll be able to figure that out tonight and report back.

ScottChesworth:
Took a run at diagnosing this as promised. My plan was to run a workflow that breaks using the OSARA key map, but without OSARA placed in UserPlugins and with eyes in attendance to verify selection. Figured that would identify whether the problem still occurs without OSARA in the equation. Unfortunately the plan fell flat because - without OSARA - I couldn't figure out how to move between automation items using keystrokes. There aren't actions for that. There didn't seem to be actions for it even once I'd reactivated OSARA, so for now I'm forced to conclude that I'm not smart enough to debug this. Back to you, @jcsteh. If I can be helpful in any way, just shout.

[...]

PR #283, opened 2020-07-18
Fix issue where Osara doesn't always find stretch markers

RDMurray:
In some cases, getStretchAtPos doesn't find the correct stretch marker, and instead finds the next stretch marker.  This causes "OSARA: Move last focused stretch marker to current edit cursor position" to fail.

This really feels like an ugly hack, but with the current Reaper API I can't see a better way of fixing it.  @jcsteh what do you think?

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -783,7 +783,8 @@ int getStretchAtPos(MediaItem_Take* take, double pos, double itemStart,
 	// Stretch marker positions are relative to the start of the item and the
 	// take's play rate.
 	double posRel = (pos - itemStart) * playRate;
-	double posRelAdj = posRel-0.00003; // approximately 1 sample at 48000. 
+	const double STRETCH_FUZZ_FACTOR = 0.00003; // approximately 1 sample at 48000.
+	double posRelAdj = posRel - STRETCH_FUZZ_FACTOR;
 	if (posRelAdj < 0)
 		return -1;
 	int index = GetTakeStretchMarker(take, -1, &posRelAdj, NULL);
@@ -792,7 +793,7 @@ int getStretchAtPos(MediaItem_Take* take, double pos, double itemStart,
 	double stretchRel;
 	// Get the real position; pos wasn't written.
 	GetTakeStretchMarker(take, index, &stretchRel, NULL);
-	if (abs(stretchRel - posRel)>0.00003)
+	if (abs(stretchRel - posRel) > STRETCH_FUZZ_FACTOR)
 		return -1; // Marker not near to  pos.
 	return index;
 }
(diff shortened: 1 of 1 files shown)


src/reaper_osara.cpp
```diff
@@ -783,16 +783,17 @@ int getStretchAtPos(MediaItem_Take* take, double pos, double itemStart,
 	// Stretch marker positions are relative to the start of the item and the
 	// take's play rate.
 	double posRel = (pos - itemStart) * playRate;
-	if (posRel < 0)
+	double posRelAdj = posRel-0.00003; // approximately 1 sample at 48000.
```
jcsteh, line 786: Given that we use this magic value twice, I think it should be a constant. A constant just inside this function is fine. Something like:
`const double STRETCH_FUZZ_FACTOR = 0.00003; // approximately 1 sample at 48000. `

jcsteh:
@RDMurray, if you're interested in inaccuracy errors like this, you may want to take a look at #141 as well.

src/reaper_osara.cpp
RDMurray: Ok, I'll make it a constant.  It could probably be a much smaller number, I'll experiment with that.

RDMurray:
I suspect it is floating point inaccuracy.  Something I just noticed about the test project I was send is that the item playrate is 0.99999999998646(from the rpp file).  It shows as 1.000000 in item properties. resetting it to 1 makes it work as expected.  

Unfortunately it does seam to be possible to have two stretch markers at exactly the same position, which is impossible to deal with without adding new actions for navigating stretch markers to osara.  Maybe adding new actions would be the best option, what do you think @jcsteh ?

jcsteh:
Ug. How do the REAPER actions cope with that? I assume they navigate to
both points, but OSARA doesn't know which one to report?

Rather than adding our own actions, maybe we could just keep track of the
previous position the command reached, and if it's the same, figure out
which point we're dealing with based on the direction we're moving?

I don't know if that should block landing of this fix, though. I'd be happy
to land this and then figure out better behaviour later, since things are
clearly better with your fix.

Lo-lo78:
Hi James,
stretch markers cannot be managed well, after having edited a stretch 
markers OSARA no longer recognizes it as stretch martkers.
Murray had something fixed but it has not yet been integrated into OSARA.
James, is there still some problem in the code or can it be integrated 
into OSARA?
   Greetings!

Il 21/07/2020 06:13, James Teh ha scritto:

jcsteh:
I reviewed the patch, but I requested a change which hasn't been made yet. See https://github.com/jcsteh/osara/pull/283#discussion_r456976303.

jcsteh (approved):
I went ahead and made the change myself.

ScottChesworth:
Thanks for pushing this through @jcsteh 

I'll post to RWP, WhatsApp and Discord now to make sure everyone who's been having problems with stretch markers knows that it's out.

PR #304, opened 2020-09-10
Fix note lengths in the midi editor. Fixes #249

### LINKED ISSUE #249: When reporting note lengths in the MIDI editor, OSARA doesn't currently factor in tempo markers (unless a tempo marker is positioned at the beginning of the project)
Reproduction steps:
1. Create a project with a BPM of 120 and a 4/4 time signature.
2. Insert a blank MIDI item, open the MIDI editor and insert an 8th note at bar 1, beat 3.
3. Hit UpArrow or DownArrow on that note to verify that OSARA reports the note length as 50%. All good so far.
4. Now insert a tempo marker at bar 1, beat 2, changing the BPM to 140.
5. Here comes the bug. Revisit the 8th note you inserted, and you'll find that OSARA is now reporting note length as 43%.
Expected behaviour would be that the note length is still reported as 50%, given that the note starts after the tempo marker.

I reckon this is an OSARA bug, because Reaper displays 50% in the Length field of the Event Properties dialogue.

Thanks in advance for any fixes. Shout out to Matej for the example.

RDMurray:
I gave up on supporting ignore project tempo, (see comments in #249 ) due to a bug in the Reaper API.


RDMurray:
Ok.  I'll make it a utility function.

The reason for PPQ in the Note struct is that I tried to support ignore project tempo.  I can change it back to project time until there is a solution to that if you like.

I was also hoping to support bars.beats.ticks in the future when it is set as the ruler format in the midi editor, although I don't think that is important.

jcsteh:
I think PPQ in the Note struct is going to be necessary eventually, but right now, it doesn't help us, so let's remove that change for now.
Regarding supporting mbt if the MIDI editor has its own ruler unit, I think that should be doable even now, since formatTime can take a TimeFormat constant specifying something other than the main ruler unit. Or am I misunderstanding you?

ScottChesworth:
Just chiming in to say that MBT being supported would send me off to my happy place. I believe in you, @RDMurray! :)

jcsteh:
Just to be clear, when we talk about mbt here, we're just talking about the MIDI editor having a separate ruler unit, right? That is, it's already possible to do mbt if you set the main ruler unit to mbt.

leonardder:
Is there something holding this back?

RDMurray:
@leonardder  I am planning to reimplement formatLength() as a utility function.  I just haven't got round to it yet, but I plan to do it next when I have some time.

leonardder:
@RDMurray feel free to shout out if you'd like me to take it over.

RDMurray:
@leonardder  sorry about the delay on this one. I definitely won't get round to it before the end of the year so feel free to take it over if you have the time.￼

src/midiEditorCommands.cpp
```diff
@@ -40,10 +40,13 @@ struct MidiNote {
 	int index;
 	double start;
 	double end;
+
 	double getLength() const {
 		return max (0, (end - start));
 	}
 };
+
+
```
jcsteh, line 49: nit: White space.
```suggestion
```

src/reaper_osara.cpp
```diff
@@ -277,6 +277,37 @@ void resetTimeCache(TimeFormat excludeFormat) {
 	}
 }
 
+string FormatNoteLength(double start, double end) {
+	int measureLength;
+	double startBeats;
+	double endBeats;
+	TimeMap2_timeToBeats(NULL, start, NULL, &measureLength, &startBeats, NULL);
```
jcsteh, line 284: nit: NULL -> nullptr
```suggestion
	TimeMap2_timeToBeats(nullptr, start, nullptr, &measureLength, &startBeats, nullptr);
```

jcsteh (approved):
Thanks.

PR #315, opened 2020-09-28
feedback for volume commands plus set item end to source media end

RDMurray:
Say track the first time  a track volume command is used to avoid confusion with item and take volume.
Implement feedback when changing item and take volume.

I would appreciate it if you could merge this before the new keymap, as the volume keys have changed and it will make getting used to the new keys easier.

@jcsteh  I changed lastCommand to int to allow it to be used in post commands.  I think the wrest of the Command struct is unused anyway. Is that correct or do you have something in mind for the future that requires lastCommand to be Command*?

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1119,6 +1119,22 @@ void postTogglePreservePitchWhenPlayRateChanged(int command) {
 	outputMessage(s);
 }
 
+void postSetItemEnd(int command) {
+	MediaItem* item = getItemInContext();
+	if(item == nullptr)
+		return;
+	ostringstream s;
+	int selCount = CountSelectedMediaItems(0);
+	if(selCount > 1){
+		s << selCount << " item ends set to source media end";
+	} else {
+		double endPos = GetMediaItemInfo_Value(item, "D_POSITION") + GetMediaItemInfo_Value(item, "D_LENGTH");
+		s << "Item end set to source media end: ";
+		s << formatTime(endPos, TF_RULER, false, false, true);
+	}
+	outputMessage(s);
+}
+
 typedef void (*PostCommandExecute)(int);
 typedef struct PostCommand {
 	int cmd;
@@ -1253,6 +1269,7 @@ PostCommand POST_COMMANDS[] = {
 	{41924, postChangeItemVolume}, // Item: Nudge items volume -1dB
 	{41927, postChangeTakeVolume}, // Take: Nudge active takes volume +1dB
 	{41926, postChangeTakeVolume}, // Take: Nudge active takes volume -1dB
+	{40612, postSetItemEnd}, // Item: Set item end to source media end
 	{0},
 };
 PostCommand MIDI_POST_COMMANDS[] = {
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -57,7 +57,7 @@ bool isSelectionContiguous = true;
 int lastCommand = 0;
 DWORD lastCommandTime = 0;
 int lastCommandRepeatCount;
-MediaItem* currentItem = NULL;
+MediaItem* currentItem = nullptr;
 
 
 /*** Utilities */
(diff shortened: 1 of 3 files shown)


RDMurray:
Added feedback for set item end to source media end. I hope you don't mind this being in the same pr, It is also required to avoid mistakes with the new keymap.

jcsteh:
> Added feedback for set item end to source media end. I hope you don't mind this being in the same pr, It is also required to avoid mistakes with the new keymap.

I'd prefer this be in a separate PR in future, since it's unrelated to the rest of it (except for the key map thing), but I'm not going to quibble on that for this PR.

jcsteh:
> I changed lastCommand to int to allow it to be used in post commands. I think the wrest of the Command struct is unused anyway. Is that correct or do you have something in mind for the future that requires lastCommand to be Command*?

One reason we might want to keep it as `Command*` is that custom actions don't have predefined ids; we get those at runtime. That would make matching against them difficult. That said, I can't think of a reason we'd actually do that, so I'm okay with this for now.

src/reaper_osara.cpp
```diff
@@ -655,25 +670,47 @@ void postGoToSpecificMarker(int command) {
...
+	MediaTrack* track = GetMasterTrack(0);
+	double volume = 0.0;
+	if ( !GetTrackUIVolPan(track, &volume, NULL) )
+		return;
+	postChangeVolumeH(volume, command, "Master ");
+}
+
+void postChangeItemVolume(int command) {
+	MediaItem* item = getItemInContext();
+	if(item == nullptr)
```
jcsteh, line 700:
```suggestion
	if (!item)
```

src/reaper_osara.cpp
```diff
@@ -655,25 +670,47 @@ void postGoToSpecificMarker(int command) {
...
+		return;
+	double volume = GetMediaItemInfo_Value(item, "D_VOL");
+	postChangeVolumeH(volume, command, "Item ");
+}
+
+void postChangeTakeVolume(int command) {
+	MediaItem* item = getItemInContext();
+	if(item == nullptr)
```
jcsteh, line 708:
```suggestion
	if (!item)
```

src/reaper_osara.cpp
```diff
@@ -1082,6 +1119,22 @@ void postTogglePreservePitchWhenPlayRateChanged(int command) {
 	outputMessage(s);
 }
 
+void postSetItemEnd(int command) {
+	MediaItem* item = getItemInContext();
+	if(item == nullptr)
```
jcsteh, line 1124:
```suggestion
	if (!item)
```

src/reaper_osara.cpp
```diff
@@ -2028,21 +2089,14 @@ void cmdMoveItems(Command* command) {
...
-	if (currentItem && ValidatePtr((void*)currentItem, "MediaItem*")
-		&& IsMediaItemSelected(currentItem)
-	) 
-	item = currentItem;
-	else if(CountSelectedMediaItems(0)>0)
-		item = GetSelectedMediaItem(0, 0);
-	else {
-		outputMessage("no items selected");
+	MediaItem* item = getItemInContext();
+	if (item == nullptr) {
```
jcsteh, line 2093:
```suggestion
	if (!item) {
```

jcsteh (changes requested):
Thanks!

src/reaper_osara.cpp
```diff
@@ -412,6 +414,19 @@ bool isTrackInClosedFolder(MediaTrack* track) {
...
+	MediaItem* item;
+	// try to provide information based on the last item spoken by osara if it is selected
+	if (currentItem && ValidatePtr((void*)currentItem, "MediaItem*")
+		&& IsMediaItemSelected(currentItem)
+	) 
+	return currentItem;
+	else if(CountSelectedMediaItems(0)>0)
+		return GetSelectedMediaItem(0, 0);
+	else 
+		return nullptr;
```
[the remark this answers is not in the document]
RDMurray: Yes, this is a problem with a lot of commands.  I can't think of a good solution.  It is the same for track volume for example.

src/reaper_osara.cpp
```diff
@@ -54,9 +54,11 @@ int oldHour;
 FakeFocus fakeFocus = FOCUS_NONE;
 bool isShortcutHelpEnabled = false;
 bool isSelectionContiguous = true;
-Command* lastCommand = NULL;
+int lastCommand = 0;
 DWORD lastCommandTime = 0;
 int lastCommandRepeatCount;
+MediaItem* currentItem = nullptr;
+
```
jcsteh, line 61: Extraneous blank line.
```suggestion
```

src/reaper_osara.cpp
```diff
@@ -655,25 +669,47 @@ void postGoToSpecificMarker(int command) {
...
+		return;
+	double volume = GetMediaItemInfo_Value(item, "D_VOL");
+	postChangeVolumeH(volume, command, "Item ");
+}
+
+void postChangeTakeVolume(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item)
+		return;
+	MediaItem_Take* take = GetActiveTake(item);
```
jcsteh, line 709: An empty item will have no take, so we have to check for that lest we crash.
```suggestion
	MediaItem_Take* take = GetActiveTake(item);
	if (!take) {
		return;
	}
```

jcsteh (approved):

PR #317, opened 2020-10-01
implement noncontiguous selection for envelope points

RDMurray:
This allows non-contiguous selection of envelope points in the same way as for tracks and items.

### DIFF
--- a/src/envelopeCommands.cpp
+++ b/src/envelopeCommands.cpp
@@ -154,7 +154,9 @@ int countSelectedEnvelopePoints(TrackEnvelope* envelope, bool max2=false) {
 	return numSel;
 }
 
-void moveToEnvelopePoint(int direction, bool clearSelection=true) {
+int currentEnvelopePoint;
+
+void moveToEnvelopePoint(int direction, bool clearSelection=true, bool select = true) {
 	TrackEnvelope* envelope;
 	double offset;
 	tie(envelope, offset) = getSelectedEnvelopeAndOffset();
@@ -175,11 +177,7 @@ void moveToEnvelopePoint(int direction, bool clearSelection=true) {
 	bool selected;
 	GetEnvelopePointEx(envelope, currentAutomationItem, point, &time, &value, NULL, NULL, &selected);
 	time += offset;
-	if ((direction == 1 && time < now)
-		// If this point is at the cursor, skip it only if it's selected.
-		// This allows you to easily get to a point at the cursor
-		// while still allowing you to move beyond it once you do.
-		|| (direction == 1 && selected && time == now)
+	if ((direction == 1 && time <= now)
 		// Moving backward should skip the point at the cursor.
 		|| (direction == -1 && time >= now)
 	) {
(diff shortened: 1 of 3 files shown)


src/envelopeCommands.cpp
```diff
@@ -154,7 +154,9 @@ int countSelectedEnvelopePoints(TrackEnvelope* envelope, bool max2=false) {
 	return numSel;
 }
 
-void moveToEnvelopePoint(int direction, bool clearSelection=true) {
+int currentEnvelopePoint;
```
jcsteh, line 157: This should probably be initialised to -1. See below.

src/envelopeCommands.cpp
```diff
@@ -175,11 +177,7 @@ void moveToEnvelopePoint(int direction, bool clearSelection=true) {
 	bool selected;
 	GetEnvelopePointEx(envelope, currentAutomationItem, point, &time, &value, NULL, NULL, &selected);
 	time += offset;
-	if ((direction == 1 && time < now)
-		// If this point is at the cursor, skip it only if it's selected.
-		// This allows you to easily get to a point at the cursor
-		// while still allowing you to move beyond it once you do.
-		|| (direction == 1 && selected && time == now)
+	if ((direction == 1 && time <= now)
```
jcsteh, line 180: I think this means that if there is a point right at the cursor, we'll always skip it when moving next. This is maybe not what the user expects? We could get around this by skipping only if the point is currentEnvelopePoint.

src/envelopeCommands.cpp
```diff
@@ -194,26 +192,43 @@ void moveToEnvelopePoint(int direction, bool clearSelection=true) {
...
 			s << " selected";
+	} else {
+		s << " unselected ";
 	}
 	s << " " << formatCursorPosition();
 	outputMessage(s);
 }
 
+bool toggleCurrentEnvelopePointSelection() {
+	TrackEnvelope* envelope = GetSelectedEnvelope(0);
```
jcsteh, line 224: This should check whether envelope is null. This function will probably need to return optional<bool> so we can tell the difference between "unselected" and "failed".

src/envelopeCommands.cpp
```diff
@@ -194,26 +192,43 @@ void moveToEnvelopePoint(int direction, bool clearSelection=true) {
...
+	bool isSelected;
+	GetEnvelopePointEx(envelope, currentAutomationItem, currentEnvelopePoint, NULL, NULL, NULL, NULL, &isSelected);
```
jcsteh, line 226: There's no guarantee that currentEnvelopePoint still exists. We should validate that and fail if it doesn't.

jcsteh (changes requested):

src/envelopeCommands.cpp
jcsteh: It looks like we don't distinguish failure for automation items either. I can live with just returning false here  too - it should be rare. However, we should return early if currentEnvelopePoint == -1, and we should set currentEnvelopePoint to -1 when switching envelopes (just like we do for currentAutomationItem).

RDMurray:
Okay, I made toggleCurrentEnvelopePointSelection() return optional<bool>.  Should I make the same change to toggleCurrentAutomationItemSelection() for the sake of consistency?

jcsteh (changes requested):
Thanks. Ideally, we'd change toggleCurrentAutomationItemSelection as well, but we can (and should) do that separately.

One more thing: we should clear currentEnvelopePoint when switching envelopes (just like we do for currentAutomationItem).

jcsteh:
Uh oh. This is failing both Windows and Mac automated builds due to the use of optional. I'm guessing the build systems we're using don't have new enough compilers/SDKs. I'll try to look into this when I get a chance.

jcsteh:
All sorted, after much tedium and exasperation. We really need to drop Xcode 9.

PR #348, opened 2020-10-25
Make it possible to move/resize the Parameters dialog for the benefit of sighted users/observers.

### LINKED ISSUE #347: Make it possible to move/resize the Parameters dialog
Hi James, I have this need, I try to communicate it to you then you will decide if it can be addressed:
when I have a sighted client I can't show him how I move the parameters of the effects, for example an equalizer or a compressor or a limiter in real time.
I'll explain what happens:
I insert an effect in the fx window, instead of closing it with "escape", I focus with NVDA or jaws on the track list.
So on the screen the fx window, then the fx list container with its buttons and the effect inside it remain visible on the screen but not focused.
The seer can then see the effect even though I am focused on anything else about reaper.
It happens that when I open the osara parameter window with the letter "P", it is placed in the center of the screen and in fact covers a part of the FX window and above all of the plugin focused inside it.
Even moving the FX window, which should be done with a viewer, the problem persists as our "form" created by osara remains in the center and above the fx window.
My request is the following if it is possible:
or put the normal controls on this window to make it smaller or move it or resize it anyway, or if you can't, make it smaller by "default" or add a check in the osara preferences that allows me to use it in minimum size or minimized.
I think this shouldn't create problems reading the list or the "slider" of values ​​neither to NVDA nor to jaws as they still read the elements through apii and not through the screen.
This would solve a great problem of embarrassment and collaboration with the client or the sighted colleague who otherwise has to wait until I have adjusted everything before being able to verify the result of what I am doing in real time.
I always thank you in advance for your kindness and I'm sorry if I ask you for things that would seem odd!

jcsteh:
Fixes #347.


jcsteh:
@ranetto, it'd be good if you could try [this test build](https://ci.appveyor.com/api/buildjobs/5cp5ccc3iflqdkey/artifacts/installer%2Fosara_pr348-245%2Cd13724f1.exe) to see if this works for you and your sighted observers. I'm totally blind myself, so I have no idea if this will work in practice. If it doesn't, I unfortunately won't be able to help you with this.

ranetto:
Hi James, perfect works you are a god !:
I am able to move my window and resize it with the dialog alt more space, and make it so that it does not bother the displayed effect.
The problem is that it should keep these dimensions and the position I gave it and instead every time I open it it returns to its original size.
If you can memorize everything works perfectly and a blind person can independently, resize more space from the alt menu, reduce it with the left arrow, and always move it from the same menu, with the right arrow move it to another part of the screen.
If you can do it you have realized my dream that will allow me to work to the fullest with all clients without embarrassment.
I would like to help you but I am an amateur programmer only in basic and in c- and only with dotnet and I don't currently know how to do it.
I hope to improve and be able to give you more and more help in the future.
In the meantime, thank you very much and I'm sorry if I keep asking.

PR #352, opened 2020-10-27
Report FX parameters changed by control surfaces, etc.

jcsteh:


ScottChesworth:
Feedback from Tim Burgess, current build in conjunction with a BCF2000 surface in Mackie mode:

1 - Pressing F12 for Osara keystroke help reported some buttons on the surface, but not all. For example previous/next marker is announced, but no Mute/solo, Select was announced.

2The buttons actions don't seem to be suppressed in Keystroke Help, so Play still starts the transport, etc.. Don't know if this is by design or a limitation on what can be done.

davidkreynolds:
Hmm, similar problems using XTouch, several keys slip through the net when keyboard help is on.

jcsteh:
REAPER doesn't provide a way to intercept (hook) control surface things. So, I can get notified about them, but I can't block them. That means input help isn't possible. The only cases where it works are those where a surface button triggers an action within REAPER, rather than using a control surface specific mechanism.

One option might be to get notified about what happened, report it and then trigger undo. I feel like that's kinda risky though and probably still wouldn't capture everything. It certainly wouldn't capture transport commands, for example. It's probably best just to say it isn't (and can't be) supported.

ScottChesworth:
Is it worth me asking Cockos for suggestions of workarounds, or better, asking whether they can extend the API for us to hook surface related stuff?

davidkreynolds:
It would certainly be to their advantage. I guess it might be a question of priority.

jcsteh:
You can ask, but my guess is it'll be difficult for them to justify the investment. Input help is a very specific feature and probably the single use case for this ability, so they'd basically be implementing this for one feature in one extension. Plus, it's something a user is only likely to use while they're learning a new surface and then only very rarely after that, if ever, so it's not even like we can say it's a critical feature. Anyway, I guess it's worth a try.

davidkreynolds:
It may be worth looking at the control surface integration project which is in the public domain. It is still being developed unlike the klinky plugin which hasn’t been touched since 2016.

leonardder:
Great to see this being extended!

ScottChesworth:
Is it possible for OSARA to distinguish between actions triggered from a keyboard and actions triggered from a surface? I'm noticing that with "Report changes made via control surfaces" checked in OSARA Config, action feedback for muting/unmuting, soloing/unsoling tracks becomes more verbose when engaging, and less context relevant when disengaging those toggles.

If this isn't possible, I'd suggest "Report changes made via control surfaces" being un-checked as the default setting, so that feedback is only altered for folks who do have surfaces.

jcsteh:
The surface code checks whether OSARA is already handling an action and doesn't speak in that case. So, this shouldn't be interfering with actions triggered from the keyboard. Can you be more specific about what's going wrong here; steps to reproduce, what happened, what you expected?

jcsteh:
Ah. Never mind; I see the problem.

It seems REAPER's surface API is very over-chatty. @leonardder already discovered this for mute/solo, but it seems even track selection suffers from this. So, when you mute a track, REAPER tells surfaces that it's being selected, even though it was already selected, causing OSARA to read it as a track change.

For now, I've disabled this option by default, though anyone who saved their configuration after this was added will need to disable it manually. I'll need to think on how to work around this.

ScottChesworth:
Got another one for ya. With "Report changes made via control surfaces" checked, when I hit Undo in the piano roll, focus gets grabbed by the TCP. Exploring with object nav, I can tell that the piano roll is still open.

jcsteh:
> I'm noticing that with "Report changes made via control surfaces" checked in OSARA Config, action feedback for muting/unmuting, soloing/unsoling tracks becomes more verbose when engaging, and less context relevant when disengaging those toggles.

Should be fixed now.


> With "Report changes made via control surfaces" checked, when I hit Undo in the piano roll, focus gets grabbed by the TCP.

Haven't tested this exact case, but it should also be fixed.

Volume, pan, mute, solo and arm changes are now also reported.

ScottChesworth:
Pre-287 solved the glitch with undo in the MIDI Editor here, thanks!

Speech verbosity is much improved. I'm still hearing some reports being said twice, but haven't figured out how to get it to happen consistently yet. Will poke more and report back.

jcsteh:
Double speaking is hopefully fixed now.

ScottChesworth:
Fantabulous. Thanks as always man!

ScottChesworth:
Question that came in via RWP:
Would it be possible to have the String spoken rather than the OSC message in Osara?
With a control surface, the return of NVDA with the Osara parameters are between 0.00 and 1.00.
If NVDA would tell us the String, we would have the plug-in values ​​rather than the OSC values.
Example: Reverb Type.
With the Osara parameters, NVDA tells us a value between 0.00 and 1.00 while the plug-in values ​​give us the names of the reverb types.
Penny for your thoughts, @jcsteh ?

jcsteh:
Are you saying you never get value names from the surface feedback? Using KK with Komplete Ultimate instruments, for example, I definitely get value names, not numbers. The numbers are a fallback and suggest REAPER was unable to get a value name.

PR #357, opened 2020-11-02
Fetch UiaDisconnectProvider and UiaDisconnectAllProviders at runtime.

### LINKED ISSUE #356: Because of the new UIA interface, the system below the earlier version of win10 cannot run reaper
The prompt for win7 users to open reaper is like this
REAPER v6.15/x64 (initializing):  reaper.exe  -Unable to find entrance
OK button
Unable to locate program input point uiadisconnectallproviders to DLL UIAutomationCore.DLL  Go ahead.

jcsteh:
Otherwise, OSARA fails to load on earlier versions of Windows where these functions don't exist.
Fixes #356.

### DIFF
--- a/src/uia.cpp
+++ b/src/uia.cpp
@@ -9,21 +9,40 @@
 #include <ole2.h>
 #include <tlhelp32.h>
 #include <atlcomcli.h>
+#include <memory>
 #include "osara.h"
 
 using namespace std;
 
 HWND uiaWnd = nullptr;
 const char* WINDOW_CLASS_NAME = "REAPEROSARANotificationWND";
-typedef HRESULT(WINAPI *UiaRaiseNotificationEvent_funcType)(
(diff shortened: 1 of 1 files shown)


jcsteh:
@cn-peter, does [this build](https://ci.appveyor.com/api/buildjobs/0ae3w9qdpla8s1qt/artifacts/installer%2Fosara_pr357-278%2Cd2350476.exe) fix the problem for you?

ranetto:
Hi James, for me he solved it I use windows 7 in the studio and the official one doesn't work but this works fine.
I can tell you that the official one doesn't work even if you disable events in the file
"extstate.ini"
adding the line:
"uiaNotificationEvents = 0"
This error occurs with both jaws and NVDA.
This version works with the two screen readers instead.
Thanks james.

jcsteh:
Thanks for confirming the fix. I'll merge it into the main snapshots now.

PR #368, opened 2020-11-15
Optionally move relative to the play cursor for time movement commands during playback.

### LINKED ISSUE #367: Move the edit cursor to the play cursor first for time movement commands
Most (maybe all?) time movement commands move relative to the edit cursor even during playback. So, for example, if you start playing at bar 1 and you wait so that you're now playing bar 5, moving to the next bar will take you to bar 2. I find this pretty annoying and unintuitive, and I believe Snowman ended up optionally hacking around this in his JAWS scripts. I'm considering doing this directly in OSARA.

The question is: is there any occasion where the default behaviour is actually more desirable? Do we need to make this optional, and if so, should it be default on or off?

I'm thinking this would also apply to moving by item.

@scottchesworth, I'd be grateful for your thoughts.

jcsteh:
This is configured via the OSARA Configuration dialog and is disabled by default.
It affects scrubbing, moving by bar/beat, moving by grid, moving to next/previous project marker and moving by item.
This also reorganises the Configuration dialog slightly.
Fixes #367.


jcsteh:
@ScottChesworth, @leonardder, @dgl1984, I'd appreciate your feedback on [this build](https://ci.appveyor.com/api/buildjobs/rgbtsj4aua7lfwj9/artifacts/installer%2Fosara_pr368-302%2C70ce9276.exe). Note the new option you have to enable in the OSARA Config dialog.

One thing I've noticed is that moving to previous item doesn't really work during playback with this enabled. It confused me at first, but it actually makes sense because by the time you move back, you've already passed the start of the item, so yes, the item before the cursor is indeed the same item you just hit. I guess we'll hit the same problem with previous marker. That does tend to make me think we shouldn't support this for items and markers... but now I'm wondering whether the whole idea falls on its head somewhat.

Or should we just limit the scope to very specific commands: scrubbing, moving by bar/beat? If so, is the current naming of the option intuitive enough? If not, suggestions?

ranetto:
Hi all guys, this difficulty of the two sliders detached and not aligned by default is one of the things I hated the most in reaper since I started using it, I solved this by creating a multi action for the plai pause that always aligns the edit slider to playback slider and here I get behavior similar to what I had in sonar:
I always use this multi action when I record, for example when I record a voice of a singer or a guitar, so if I have to stop to resume from that point or make an entry on the fly I am already where I need it and I do not always have to go back to the point where I was.
I use reaper's default behavior only when I have to record a midi part and maybe I have to repeat it.
Having the two separate cursors for a blind person is much more difficult as not being able to see at a glance where the recorded item ends is always looking for that point.
To move by grid values ​​I have created a multi action on the dialogue to push which with the left and right arrows makes me move by the amount set in the grid.
Here this is just my experience I hope I was able to explain what I mean.

ScottChesworth:
Hey @ranetto, your approach sounds like it might be useful for other people too. Would you mind exporting and sharing the actions you've made so we can host them in the Reaper Dropbox while this new feature is being refined?

ScottChesworth:
It's brilliant to be able to scrub from the position of the play cursor again (this feels so much more intuitive that I'd like to see it become the default behaviour), and I'm also enjoying the new behaviour when moving by bar/beat more than I thought I would.
I can however foresee plenty of workflows where it would be useful to be able to toggle the behaviour of those two aspects individually. So, could I persuade you to add another check box? Thinking the labels would be something like: 
"Scrub relative to the play cursor when scrubbing from playback", checked by default
"Move relative to the play cursor when navigating by measure or beat during playback", probably unchecked by default until we find out how popular it is in the wild

I appreciate the logic of why navigating backward through items/markers doesn't work as things currently stand, but it just feels broken lol. I can't think of a situation where that behaviour would be desireable here, so unless anyone else can, I reckon just adopt the behaviour for scrub and moving by measure/beat for now.

ranetto:
certainly Scott, would you prefer me to send you the single actions or my entire ini file?
I do it as soon as I get back to the studio!

jcsteh:
Technically, the only difference between Sonar and REAPER is that Sonar had an option to do the equivalent of REAPER "pause" when you pressed stop. That is, pressing stop would snap the edit cursor to the play cursor. During actual playback, the edit cursor stayed back where playback started.

jcsteh:
> It's brilliant to be able to scrub from the position of the play cursor again

Does this suffer from the same problems your custom actions had? I didn't use them, so I don't understand what the problem was.

> I can however foresee plenty of workflows where it would be useful to be able to toggle the behaviour of those two aspects individually.

I'm going to push back on this to be my usual pedantic self. (That's why you love working with me, right?) Why is moving by bar/beat special? I can maybe see why scrubbing is special, although even that's arguable; both of them involve moving the cursor. Sure, they have different use cases, but why would you want to be able to immediately fine edit from where you're playing, but only jump forward/backward in larger increments from the edit cursor?

> So, could I persuade you to add another check box?
> "Move relative to the play cursor when navigating by measure or beat during playback", probably unchecked by default until we find out how popular it is in the wild

What about grid division? What about any other time movement actions we want to add here? I'm really reluctant to make this so specific to bar/beat, because I foresee 7 options for this functionality somewhere in our future, each for a different time movement action.

> I appreciate the logic of why navigating backward through items/markers doesn't work as things currently stand, but it just feels broken lol.

One thought I had is that I could possibly make the previous item/marker actions look for an item/marker 1 second (or 500 ms?) before the play cursor. The net impact would be that tapping the action twice would jump to the previous one as you would expect, kinda like CD players of old.

ScottChesworth:
> Does this suffer from the same problems your custom actions had?
>
Nope, with your implementation, scrubbing has been working fine and dandy all day here going straight from playback and also from pause (the latter being what got broke by the custom actions).
> I'm going to push back on this to be my usual pedantic self. (That's why you love working with me, right?)
> 
It's in among the top 20 reasons, yes. Numero 19 every time I make the list, but hey, it's in there. :)_
> Why is moving by bar/beat special? I can maybe see why scrubbing is special, although even that's arguable; both of them involve moving the cursor. Sure, they have different use cases, but why would you want to be able to immediately fine edit from where you're playing, but only jump forward/backward in larger increments from the edit cursor?
> 
Yeah ok, now I've spent a bit more time getting comfortable with the new behaviour in practice, I don't think there's a strong case for these to be separate options anymore.
> What about grid division?
> 
I reckon roll this out without movement by grid factored in to start with, see whether anyone requests that to be added. To me, it makes sense for grid-based movement to stay as predictable as possible. IE, if I deliberately set myself up to be part way through a division, I wanna know the edit cursor is still gonna be positioned in that spot throughout future bars/beats until I decide to move it.
> One thought I had is that I could possibly make the previous item/marker actions look for an item/marker 1 second (or 500 ms?) before the play cursor. The net impact would be that tapping the action twice would jump to the previous one as you would expect, kinda like CD players of old.
> 
Nifty solution, but I'm not sure. It doubles the amount of keystrokes when you need to zip back through items and just hear the first few seconds to find something specific. I reckon just don't include item/marker navigation until users start asking for those specifically.

jcsteh:
> It's in among the top 20 reasons, yes. Numero 19 every time I make the list, but hey, it's in there. :)_

Now I want to know what reason 20 is. :)

> Yeah ok, now I've spent a bit more time getting comfortable with the new behaviour in practice, I don't think there's a strong case for these to be separate options anymore.

See? This is why I push back on these things. :)

> I reckon roll this out without movement by grid factored in to start with

Okay. For now, it just does scrub and bar/beat.

ranetto:
Hi Scott, I'm sending you the commands taken from my keymap that we talked about: 
I am very happy: 
you will also find the explanation in this meil of the same: 
let me know! 


CTRL WIN ALT SPACE: stefano plaistopp.ReaperKeyMap 

play and stop as in sonar 


CTRL WIN ALT left: Move cursor left to grid division.ReaperKeyMap 

CTRL WIN ALT right: Move cursor right to grid division.ReaperKeyMap 


next measure precise.ReaperKeyMap 

previeus misure precise.ReaperKeyMap 

They move to the previous or next measure as it happened in sonar, that is, without the percentages but always at beat 1 percent 0 


select first track.ReaperKeyMap 

select last track.ReaperKeyMap 


always available for any question! 

[stefano mapp.zip](https://github.com/jcsteh/osara/files/5543534/stefano.mapp.zip)

PR #376, opened 2020-11-17
Delay UIA initialisation in hopes of fixing an issue where some users don't get OSARA messages unless they restart NVDA.

jcsteh:
This is a very speculative fix. I really have no idea whatsoever whether it will work.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3234,9 +3234,16 @@ bool handleMainCommandFallback(int command, int flag) {
 	return false;
 }
 
+IReaperControlSurface* surface = nullptr;
+
 // Initialisation that must be done after REAPER_PLUGIN_ENTRYPOINT;
 // e.g. because it depends on stuff registered by other plug-ins.
 void CALLBACK delayedInit(HWND hwnd, UINT msg, UINT_PTR event, DWORD time) {
+#ifdef _WIN32
+	initializeUia();
+#endif
+	surface = createSurface();
+	plugin_register("csurf_inst", (void*)surface);
 	NF_GetSWSTrackNotes = (decltype(NF_GetSWSTrackNotes))plugin_getapi(
 		"NF_GetSWSTrackNotes");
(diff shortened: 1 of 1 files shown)


leonardder:
Is there a reason why surface initialization is also moved?

jcsteh:
Nice to see someone is watching my code. Haha.

I've noticed that REAPER tends to tell the surface that the transport has stopped just after REAPER starts. This raises three concerns:

1. It results in an annoying "stop" message on startup sometimes.
2. If that message gets sent before UIA is initialised (which happens when you move that into delayedInit), shouldUseUiaNotifications caches that UIA wasn't initialised and so we never use UIA. I could have fixed that by moving the uiaWnd check out of the cache block. However...
3. I also worry that sending a UIA notification that early in REAPER's startup might result in a situation where UIA Core decides to ignore events from REAPER; e.g. because it thinks the process is taking too long to respond since it's still doing startup stuff, keeping in mind that NVDA now has UIA connection recovery enabled, etc.

leonardder:
Sounds reasonable to me.

> Nice to see someone is watching my code. Haha.

Well, I am suffering from this myself and noticed this coming along, so I thought I'd have a look.

jcsteh:
Ah. Well, if this fails, perhaps I could shoot you a build with some debugging crap to see if I can glean some more info? I can't reproduce it here.

leonardder:
I can only reproduce it on one of my systems. I"ll have a look.

leonardder:
Ugh, now I can't trigger it there either.

I think your rationale is ok, and therefore I'd merge this and see how 
it performs. It does no harm at all I think.

jcsteh:
I'd really prefer not to merge this if it doesn't help; it just makes for
more variables. That said, I've asked some people on RWP who were
experiencing this more frequently to test the build to see if it helps. If
it does, I'll merge it. If not, I'll abandon it.

jcsteh:
This seems to fix the issue for one person at least.

Issue #340, opened 2020-10-21
Announce the value of the sensitivity and threshold of the transients


Lo-lo78:
Hi James and everyone!
I am exploring the transients and I feel the need to have some feedback from OSARA to know the percentage of sensitivity or the threshold of the transients.
When I do:

Transient detection sensitivity: Decrease
Transient detection sensitivity: Increase

Transient detection threshold: Decrease
Transient detection threshold: Increase

I don't know the actual value and that's why I can't understand the right setting; I have to go instinct!

When I go into the dialog box:

Transient detection sensitivity / threshold: Adjust ...

You can only adjust the first progress slider, the sensitivity one, when I go to adjust the second progress slider, the threshold one, I can't even use the NVDA object navigator.

Is it possible to have OSARA announce the values ​​when we do the shortcuts?
Thanks a lot!

ranetto:
I think this would be very useful, also to be used in dynamic cropping of items.
I can only give a suggestion, when you have focused the second slider of the adjustment window, try to adjust its values with the controll and the arrows.

ranetto:
Hi James and everyone, sorry for my unfamiliarity with "Gtub", but I didn't understand if this thing cannot be solved due to a lack of bees, or if it is being created: it would be very important: thank you James indeed!

jcsteh:
@scottchesworth, I think I've found a way to force the tab and arrow keys in the transient detection adjust dialog to be handled by the dialog instead of the main section. The question is whether that's a good idea or not.

The reason that REAPER doesn't do this by default is because people generally want to be able to play with stuff while adjusting the transients. Tab is especially problematic because tab is move to transient and we're adjusting transients! I'd argue that for screen reader users, the inability to tab to transient here is outweighed by the need to be able to tab through the damned dialog. However, I'd like your opinion on this before I go charging ahead with it, especially as it's a bit tricky to get right.

There are a few paths forward here:

1. Implement the change I mentioned above, which would mean you can tab through all the fields in that dialog and use the arrow keys, etc. One nice part about that is you can also use the text boxes to type numbers.
2. Implement feedback for the increase/decrease commands.
3. Both of the above.

jcsteh:
By the way, @ranetto, please keep in mind that I work on OSARA in my very limited spare time and am not compensated for it in any way these days. If something can't be done or is being done, I'll generally comment on the issue explaining that. If I don't comment, it probably means I haven't had time or inclination to look into it. Thanks for your understanding.

[...]

PR #377, opened 2020-11-17
Allow tab, the cursor keys and space to interact with the Transient Detection Settings dialog.

jcsteh:
Re #340.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -2893,6 +2893,52 @@ void cmdAbout(Command* command) {
 	reviewMessage("About OSARA", s.str().c_str());
 }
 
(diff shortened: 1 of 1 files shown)


Lo-lo78:
Hi James, the dialog works fine; a little problem happens. If I septum with a shortcut and I open the dialog box and repeat the same shortcut that closes the window, OSARA does not take me to the TCP. With the up or down arrow it moves to another focus. 
Conversely, if I just hit the ESC key, the TCP is focused, it works fine. 
Try it!

jcsteh:
Thanks. That's because I am daft. Fix upcoming.

ranetto:
Great James, it works great!
That's a great thing, I'll be using it tonight!

PR #379, opened 2020-11-18
Make osara respect the undo checkboxes in the general section of preferences 

RDMurray:
There are check boxes in preferences that can be disabled to prevent item and track selection from creating undo points.  This makes Osara respect those settings.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -483,11 +483,12 @@ void reviewMessage(const char* title, const char* message) {
 	ShowWindow(dialog, SW_SHOWNORMAL);
 }
 
-unsigned long getConfigUndoMask() {
-	char bufOut[10];
-	if(!get_config_var_string("undomask", reinterpret_cast<char*>(&bufOut), 10))
+unsigned int getConfigUndoMask() {
+	int size{0};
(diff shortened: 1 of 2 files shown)


src/reaper_osara.cpp
```diff
@@ -483,6 +483,13 @@ void reviewMessage(const char* title, const char* message) {
 	ShowWindow(dialog, SW_SHOWNORMAL);
 }
 
+unsigned long getConfigUndoMask() {
+	char bufOut[10];
+	if(!get_config_var_string("undomask", reinterpret_cast<char*>(&bufOut), 10))
+		return 0;
+	return stoul(bufOut);
```
jcsteh, line 490: Did you try using get_config_var instead? I think that should return this as an unsigned long directly (so you can just cast), rather than having to convert from a string here. Or it might return it as a double, but casting a double to an unsigned long is still cleaner.

RDMurray: I have changed it to use get_config_var with unsigned int.  The only reason I didn't do that initially was that get_config_var is undocumented.

If we need to access more config vars in the future, it might be an idea to use utility/config.h from sws.

jcsteh (approved):
Thanks.

PR #386, opened 2020-11-21
Add feedback to normalize actions

leonardder:
Added feedback to the following actions:

- Item properties: Normalize items
- Item properties: Normalize multiple items to common gain

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -670,10 +670,14 @@ void postCursorMovementScrub(int command) {
 void postItemNormalize(int command) {
 	int selectedItemsCount = CountSelectedMediaItems(0);
 	ostringstream s;
-	s << selectedItemsCount << (selectedItemsCount == 1 ? " item" : " items") << " normalized";
-	if (command == 40254) {
-		// Item properties: Normalize multiple items to common gain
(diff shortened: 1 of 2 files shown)


src/reaper_osara.cpp
```diff
@@ -667,6 +667,17 @@ void postCursorMovementScrub(int command) {
 		fakeFocus = FOCUS_RULER; // Set this even if we aren't reporting.
 }
 
+void postItemNormalize(int command) {
+	int selectedItemsCount = CountSelectedMediaItems(0);
```
jcsteh, line 671: I know we say "0 items removed", "0 items added", etc., but those feel like they add clarity since they're potentially changing the number of items. In contrast, saying "0 items normalized" feels weird; I worry people will think something happened when it didn't. Can we just return or say "no selected items" if the count is 0?

jcsteh (changes requested):
Thanks.

src/reaper_osara.cpp
leonardder: That makes sense. I went for "no selected items".

jcsteh (approved):

jcsteh (approved):

PR #390, opened 2020-11-23
Add View propertiess for current media item / take / automation item (depending on focus)

### LINKED ISSUE #128: Checkbox state not reported by screen readers when importing/exporting prefs
Filing this here, but it's probably destined for Cockos rather than one to fix via OSARA.

From the changelog:
v5.51 - September 30 2017
  + Configuration import/export: allow toggling checkbox state with space key

This was a helpful ux change, but the state of each item isn't reported by screen readers yet.

To reproduce, go to Preferences / General, then hit either Import or Export.

Jamie, when you get a sec, can you pass this along to Justin?

### LINKED ISSUE #138: Suggested shortcut for invoking automation item properties
I agree with Gianluka that the most elegant solution for this would be to have shift f2 invoke the right properties dialogue at the right time, however, I wondered if control shift f2 would be a good ke in the mean-time. It makes a kind of mnemonic sense.

leonardder:
Note, I'm not sure whether the case blocks are complete like this. Do I also need to add FOCUS_ENVELOPE to the command for automation item propertis? I'm also not very familiar to automation items actually, I mixed them up with media items the other day.

Closes #138

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -594,6 +594,7 @@ OSARA also includes some other miscellaneous actions.
 #### Main
 - OSARA: Move to next item (leaving other items selected): control+shift+rightArrow
 - OSARA: Move to previous item (leaving other items selected): control+shift+leftArrow
+- OSARA: View propertiess for current media item / take / automation item (depending on focus)
 - OSARA: View I/O for master track: shift+i
 - OSARA: Report ripple editing mode: control+shift+p
 - OSARA: Report muted tracks: Alt+shift+f5
--- a/readme.md
+++ b/readme.md
@@ -594,7 +594,7 @@ OSARA also includes some other miscellaneous actions.
 #### Main
 - OSARA: Move to next item (leaving other items selected): control+shift+rightArrow
 - OSARA: Move to previous item (leaving other items selected): control+shift+leftArrow
-- OSARA: View propertiess for current media item / take / automation item (depending on focus)
+- OSARA: View properties for current media item/take/automation item (depending on focus)
 - OSARA: View I/O for master track: shift+i
 - OSARA: Report ripple editing mode: control+shift+p
 - OSARA: Report muted tracks: Alt+shift+f5
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -2425,17 +2425,10 @@ void cmdMoveToPrevItemKeepSel(Command* command) {
 }
 
 void cmdPropertiesFocus(Command* command) {
-	switch (fakeFocus) {
-		case FOCUS_ITEM:
-		case FOCUS_TAKEMARKER:
-			Main_OnCommand(40009, 0); // Item properties: Show media item/take properties
-			break;
-		case FOCUS_AUTOMATIONITEM:
-		case FOCUS_ENVELOPE:
-			Main_OnCommand(42090, 0); // Envelope: Automation item properties...
-			break;
-		default:
-			break;
+	if (shouldMoveToAutoItem) {
+		Main_OnCommand(42090, 0); // Envelope: Automation item properties...
+	} else {
+		Main_OnCommand(40009, 0); // Item properties: Show media item/take properties
 	}
 }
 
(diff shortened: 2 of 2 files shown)


readme.md
```diff
@@ -594,6 +594,7 @@ OSARA also includes some other miscellaneous actions.
 #### Main
 - OSARA: Move to next item (leaving other items selected): control+shift+rightArrow
 - OSARA: Move to previous item (leaving other items selected): control+shift+leftArrow
+- OSARA: View propertiess for current media item / take / automation item (depending on focus)
```
jcsteh, line 597: Spelling error, no spaces between slashes for consistency.
```suggestion
- OSARA: View properties for current media item/take/automation item (depending on focus)
```

src/reaper_osara.cpp
```diff
@@ -2424,6 +2424,20 @@ void cmdMoveToPrevItemKeepSel(Command* command) {
 	}
 }
 
+void cmdPropertiesFocus(Command* command) {
+	switch (fakeFocus) {
+		case FOCUS_ITEM:
+		case FOCUS_TAKEMARKER:
+			Main_OnCommand(40009, 0); // Item properties: Show media item/take properties
+			break;
+		case FOCUS_AUTOMATIONITEM:
```
jcsteh, line 2433: As you suspected, this should include FOCUS_ENVELOPE. The reason is that you might move to an automation item, then move to an envelope point within that item, which will set FOCUS_ENVELOPE... but the automation item is still selected.
```suggestion
		case FOCUS_AUTOMATIONITEM:
		case FOCUS_ENVELOPE:
```

src/reaper_osara.cpp
```diff
@@ -2424,6 +2424,20 @@ void cmdMoveToPrevItemKeepSel(Command* command) {
...
+			Main_OnCommand(42090, 0); // Envelope: Automation item properties...
+			break;
+		default:
+			break;
```
jcsteh, line 2437: While this does make sense for a new command, because we're looking to replace the existing item properties command, I worry that users will be confused when they press shift+f2 when focus is on the ruler (just as an example) and nothing happens. On the other hand, it's probably weird if they move to an auto item, move on the ruler, press properties and end up with media item properties.

My suggestion is that if we don't know how to handle the focus (this default block), we should make a decision based on the shouldMoveToAutoItem variable. If true, use auto item properties. If false, use media item properties. shouldMoveToAutoItem gets set to false whenever you move to a track and set to true whenever you move to an envelope or envelope point.

That makes me wonder whether we should even bother looking at fakeFocus at all at this stage or just make the decision purely based on that variable. Obviously, that might have to change in future if we start adding more properties things to this command.

src/reaper_osara.cpp
```diff
@@ -3082,6 +3096,7 @@ Command COMMANDS[] = {
 	// Our own commands.
 	{MAIN_SECTION, {DEFACCEL, "OSARA: Move to next item (leaving other items selected)"}, "OSARA_NEXTITEMKEEPSEL", cmdMoveToNextItemKeepSel},
 	{MAIN_SECTION, {DEFACCEL, "OSARA: Move to previous item (leaving other items selected)"}, "OSARA_PREVITEMKEEPSEL", cmdMoveToPrevItemKeepSel},
+	{MAIN_SECTION, {DEFACCEL, "OSARA: View propertiess for current media item / take / automation item (depending on focus)"}, "OSARA_PROPERTIES", cmdPropertiesFocus},
```
jcsteh, line 3099: Ditto
```suggestion
	{MAIN_SECTION, {DEFACCEL, "OSARA: View properties for current media item/take/automation item (depending on focus)"}, "OSARA_PROPERTIES", cmdPropertiesFocus},
```

jcsteh (changes requested):
Thanks.

src/reaper_osara.cpp
leonardder: This makes sense to me. Changed as suggested.

jcsteh (approved):
Thanks.

PR #391, opened 2020-11-24
Use a binary search when looking for chords and notes in the MIDI editor to improve performance.

### LINKED ISSUE #105: Lags in moving be3tween notes in large MIDI items
In large MIDI files moving between notes is taking much of time. Here is an example big MIDI file: http://www.kunstderfuge.com/-/midi.asp?file=stravinsky_firebird_suite_(c)nakajima.mid
If you load this file as one multichannel item, open MIDI editor and start to move between notes it will be probably verry laggish. Of course usually MIDI items will have far less notes, but if someone would like to record some concert or long song using MIDI and edit it it can be somewhat problematic.

jcsteh:
I've spent several hours implementing this. To be honest, this is still pretty slow and I don't think this is all OSARA. My guess is that REAPER just isn't great at handling MIDI items that are this big. Still, it does seem to at least make the lag a bit more consistent. However, it needs some heavy testing, and if we aren't seeing significant benefits, it's probably not worth it, as it complicates the code quite a bit.

Closes #105.


Lo-lo78:
Yes, the scrolling of notes / chords is much faster! Fantastic! Before 
if I went forward with the right arrow it was fast but if with large 
items I went backwards with the left arrow it was slower, now it's perfect!

jcsteh:
If it's improving things for someone, I think it's worth merging. Thanks for updating to master, @leonardder.

src/midiEditorCommands.cpp
```diff
@@ -304,49 +407,46 @@ const string getMidiNoteName(MediaItem_Take *take, int pitch, int channel) {
...
-	int lastNote = firstNote;
-	for (int note = lastNote + movement; 0 <= note && note < count; note += movement) {
-		double start;
-		MIDI_GetNote(take, note, NULL, NULL, &start, NULL, NULL, NULL, NULL);
-		start = MIDI_GetProjTimeFromPPQPos(take, start);
-		if (start != firstStart)
+	double firstStart = *firstNote;
+	MidiNoteIterator lastNote = firstNote;
+	MidiNoteIterator note = firstNote;
+	for (note += direction; begin <= note && note < end; note += direction) {
```
leonardder, line 442: I'm a bit puzzled by this. Why does this note += direction twice? Doesn't this mean note is updated twice in the loop, once at the start of the loop and one at the end?

leonardder (commented):
I'm diving into this algorithm a little more and think I don't get something. May be it's a bug, I don't know. Should have done that before merge, haha.

jcsteh:
No. The first clause of a for loop is the initialiser, only run once at the
start of the whole loop. The final clause runs at the end of each
iteration. So, this is saying: start with the note one beyond the first
note, then keep advancing one beyond that for each iteration.

jcsteh:
I guess I could have made this more readable like this:
`for (MidiNoteIterator note = firstNote + direction; begin <= note && note < end; note += direction) {`
However, that would require implementing operator+ for the iterator. That's not hard, but it just didn't seem worth adding yet another function there. :)

PR #392, opened 2020-11-24
Add feedback for move notes left/right one grid unit/pixel actions

leonardder:
Note that moving notes horizontally doesn't change the cursor position, therefore the generalisation behavior is somewhat different from the change pitch / length actions.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1000,7 +1000,7 @@ void postMidiChangeLength(int command) {
 	MediaItem_Take* take = MIDIEditor_GetTake(editor);	
 	// Get selected notes.
 	vector<MidiNote> selectedNotes = getSelectedNotes(take);
-		if (selectedNotes.size() == 0) {
+	if (selectedNotes.size() == 0) {
 		return;
 	}
 	bool generalize = false;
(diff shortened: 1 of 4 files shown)


src/midiEditorCommands.cpp
```diff
@@ -1121,6 +1121,59 @@ void postMidiChangePitch(int command) {
 	outputMessage(s);
 }
 
+void postMidiMoveStart(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
```
jcsteh, line 1126: nit: Spacing
```suggestion
	MediaItem_Take* take = MIDIEditor_GetTake(editor);
```

src/midiEditorCommands.cpp
```diff
@@ -1121,6 +1121,59 @@ void postMidiChangePitch(int command) {
...
+					break;
+				case 40184:
+					s << "grid unit right";
+					break;				
+				default:
+					s << "start moved";
+					break;
+			}
+		} else{ 
+			for (auto note = selectedNotes.cbegin(); note != selectedNotes.cend(); ++note) {
```
jcsteh, line 1163: Because we're now using C++ 17, you can probably just do this (but I haven't tested):
`for (auto note: as_const(selectedNotes)) {`
(You need to `#include <utility>` for as_const.)
However, this is just something to consider, not a required change for approval.

src/midiEditorCommands.cpp
```diff
@@ -1121,6 +1121,59 @@ void postMidiChangePitch(int command) {
...
+				if (note == selectedNotes.cbegin()) {
+					s << formatTime(note->start, TF_MEASURE) << " ";
```
jcsteh, line 1165: Can we do this using `firstStart` (defined earlier) before the loop, rather than doing this within the loop?

jcsteh (changes requested):
Thanks.

src/midiEditorCommands.cpp
leonardder: I had tried something like that the other day, specifically:
`for (auto const& note: selectedNotes) {`
However, I haven't managed in converting the `if (note != selectedNotes.cend() - 1) {`: Tried
`if (note != *(selectedNotes.cend() - 1)) {` but then, it can't compare const MidiNote against const MidiNote. I assume I have to implement an operator for that?

src/midiEditorCommands.cpp
leonardder: Yes, of course. I hadn't thought about that when I changed the logic to no longer find chords.

jcsteh (approved):
Thanks. Sounds like the for loop change is just going to make things more complicated at this point, so let's leave it alone.

PR #403, opened 2020-12-02
Add an FX preset dialog to easily browse presets without activating them.

### LINKED ISSUE #173: Opening the presets combo box in the FX window for faster scrolling
When scrolling through the presets combo box in the FX window using Up or Down arrows, the default behaviour is that the combo box remains closed, and as a consequence, plugins will load each preset as the user scrolls. With most EQs and compressors and the like this isn't a problem because plugin settings update almost instantaneously, but with many synth libraries and more resource intensive plugins, it decreases efficiency when searching through presets.

The UX for sighted users is different, and a lot more productive. They single click on the combo box to open it, then scroll through presets without the plugin updating presets as they pass while the combo box is open. A double click on the desired preset closes the combo box, and only then does the plugin load those settings.

Poking at this with object nav, I can replicate the behaviour that my sightling reported, but it involves more button bashing than is practical.

Part of the problem seems to be that Alt+DownArrow and Alt+UpArrow don't get passed through to the FX window, so it's not currently possible to open the presets combo box without object nav.

Is there anything that can be done to improve the UX here, either by OSARA or Cockos?

Thanks in advance for any thoughts.

jcsteh:
REAPER's FX preset combo box doesn't allow keyboard users to move through presets without activating them.
This new dialog, activated by pressing alt+downArrow when focused on REAPER's FX preset combo box, displays the preset list and allows you to navigate it with the keyboard without activating the preset.
As an added bonus, you can filter presets to show only those containing a certain piece of text.
Pressing the OK button activates the selected preset.
This only works on Windows for now.
Fixes #173.


src/osara.h
```diff
@@ -261,5 +261,7 @@ IReaperControlSurface* createSurface();
 extern bool selectedEnvelopeIsTake;
 // exports.cpp
 void registerExports(reaper_plugin_info_t* rec);
+// fxChain.cpp
+bool maybeOpenFxPresetDialog();
```
leonardder, line 265: Is this one really needed given src/fxChain.h is there?

jcsteh:
Blah. Mac build fails, even though it's disabled on Mac.

I need to go to bed. If anyone wants to test this, [here's the successful Windows build](https://ci.appveyor.com/api/buildjobs/fht5837ep2kef8do/artifacts/installer%2Fosara_pr403-357%2C487ad25b.exe). I'll fix the Mac crap some other day.

src/osara.h
jcsteh: Ooo! Great catch. Nope, that was before I decided to stop being lazy and add a new header file... but then I forgot to remove the manifestation of my laziness. :)

ScottChesworth:
I'm facing Australia and fan-girling so hard right now.

One little gotcha here is that hitting Enter moves to the desired preset, but doesn't seem to actually load it with any of the plugs I've tested yet (some Waves stuff, SSD5.5, EZdrummer etc). So, I still have to move off of the desired preset and back onto it to get it to load. To be fair, that's already a vast, vast productivity improvement particularly with larger RPL's loaded, but if there's anything you can do to tighten up loading, that would be awesome.

jcsteh:
> One little gotcha here is that hitting Enter moves to the desired preset, but doesn't seem to actually load it

Fixed.

jcsteh:
@ScottChesworth , from my perspective, this is good to go unless you find any further issues. Please let me know when you think we're good to hit the button.

ScottChesworth:
> Please let me know when you think we're good to hit the button.
> 
I'd like to do a quick, aggressive round of testing with certain larger, problematic RPL files here if that's ok. I know of some folks who are depending on these in proper production environments and should be able to test quicksharp. This is gonna be such a game-changer, I'd really like to see a smooth reception for it.
I'll circulate this commit to relevant people, try really hard to break it here and report back tomorrow.

jcsteh:
Please do test the hell out of it. No rush. If all is well, it's very easy
for me to hit the button whenever.

ScottChesworth:
I've been trying all afternoon and haven't managed to break it yet. Nothing other than positive reports from the people I sent it out too either.
Let 'em have it, I reckon. 👍

PR #411, opened 2020-12-05
Play note preview in event list when note gets focus

### LINKED ISSUE #410: Play selected note in midi event list
I'd like to explore the possibility to play the selected note when changing selection in the midi event list. @jcsteh have you ever considered this? I'm afraid it will need some pretty hacky hooking.

leonardder:
Fixes #410

@jcsteh I'm afraid this went a bit of a mix between C and C++. I considered using a std::string in place of the c style string, from C++17 and onwards it should be possible to provide string.data() as a pointer to functions that expect C style strings. However, there were some issues in reusing the string for the several calls I implemented.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -588,6 +588,9 @@ Noncontiguous selection is also possible.
 You do this in the same way described above for tracks and items.
 That is, press shift+space (OSARA: Enable noncontiguous selection/toggle selection of current chord/note) to switch to noncontiguous selection, move to other chords/notes with shift plus the arrow keys and press shift+space to select/unselect the current chord/note.
 
+#### Note preview in de event list (Windows Only)
+When the MIDI Editor is set to Event List mode, REAPER presents a list with all the events in the current MIDI item. When a note gets focus in the list, OSARA will play a preview of the respective note.
+
 ### Navigating FX Presets Without Activating Them (Windows Only)
 REAPER's FX preset combo box doesn't allow keyboard users to move through presets without activating them.
 Sometimes, you need to be able to examine the presets without activating each one.
--- a/readme.md
+++ b/readme.md
@@ -588,8 +588,9 @@ Noncontiguous selection is also possible.
 You do this in the same way described above for tracks and items.
 That is, press shift+space (OSARA: Enable noncontiguous selection/toggle selection of current chord/note) to switch to noncontiguous selection, move to other chords/notes with shift plus the arrow keys and press shift+space to select/unselect the current chord/note.
 
-#### Note preview in de event list (Windows Only)
-When the MIDI Editor is set to Event List mode, REAPER presents a list with all the events in the current MIDI item. When a note gets focus in the list, OSARA will play a preview of the respective note.
+### Note Preview in the MIDI Event List (Windows Only)
+When the MIDI Editor is set to Event List mode, REAPER presents a list with all the events in the current MIDI item.
+When a note gets focus in the list, OSARA will play a preview of the focused note.
 
 ### Navigating FX Presets Without Activating Them (Windows Only)
 REAPER's FX preset combo box doesn't allow keyboard users to move through presets without activating them.
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1000,7 +1000,7 @@ void maybePreviewCurrentNoteInEventList(HWND hwnd) {
 	// Get the text from the length column (2).
 	// If this column is empty, we aren't dealing with a note.
 	ListView_GetItemText(hwnd, focused, 2, text, sizeof(text));
-	if (strcmp(text, "\0") == 0) {
+	if (!text[0]) {
 		return;
 	}
 	MidiNote note;
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -996,29 +996,27 @@ void cmdMidiFilterWindow(Command *command) {
 
 void maybePreviewCurrentNoteInEventList(HWND hwnd) {
 	auto focused = ListView_GetNextItem(hwnd, -1, LVNI_FOCUSED);
-	char text[50];
-	// Get the text from the type column (4).
-	ListView_GetItemText(hwnd, focused, 4, text, sizeof(text));
-	if (strcmp(text, "Note") != 0) {
+	char text[50] = "\0";
+	// Get the text from the length column (2).
+	// If this column is empty, we aren't dealing with a note.
+	ListView_GetItemText(hwnd, focused, 2, text, sizeof(text));
+	if (strcmp(text, "\0") == 0) {
 		return;
 	}
-	fill(begin(text), end(text), 0);
 	MidiNote note;
+	// Convert length to project time. text is always in measures.beats.
+	auto length = parse_timestr_len(text, 0, 2);
+	text[0] = '\0';
 	// Get the text from the position column (1).
 	ListView_GetItemText(hwnd, focused, 1, text, sizeof(text));
 	// Convert this to project time. text is always in measures.beats.
 	note.start = parse_timestr_pos(text, 2);
-	fill(begin(text), end(text), 0);
-	// Get the text from the length column (2).
-	ListView_GetItemText(hwnd, focused, 2, text, sizeof(text));
-	// Convert this to project time. text is always in measures.beats.
-	auto length = parse_timestr_len(text, 0, 2);
 	note.end = note.start + length;
-	fill(begin(text), end(text), 0);
+	text[0] = '\0';
 	// Get the text from the channel column (3).
 	ListView_GetItemText(hwnd, focused, 3, text, sizeof(text));
 	note.channel = atoi(text);
-	fill(begin(text), end(text), 0);
+	text[0] = '\0';
 	// Get the text from the parameter (note name) column (5).
 	ListView_GetItemText(hwnd, focused, 5, text, sizeof(text));
 	string noteNameWithOctave { text };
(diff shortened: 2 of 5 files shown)


readme.md
```diff
@@ -588,6 +588,9 @@ Noncontiguous selection is also possible.
 You do this in the same way described above for tracks and items.
 That is, press shift+space (OSARA: Enable noncontiguous selection/toggle selection of current chord/note) to switch to noncontiguous selection, move to other chords/notes with shift plus the arrow keys and press shift+space to select/unselect the current chord/note.
 
+#### Note preview in de event list (Windows Only)
```
jcsteh, line 591: Make level 3 heading, linguistic/capitalisation fix.
```suggestion
### Note Preview in the MIDI Event List (Windows Only)
```

readme.md
```diff
@@ -588,6 +588,9 @@ Noncontiguous selection is also possible.
...
+When the MIDI Editor is set to Event List mode, REAPER presents a list with all the events in the current MIDI item. When a note gets focus in the list, OSARA will play a preview of the respective note.
```
jcsteh, line 592: Split sentences onto separate lines. Minor linguistic tweak.
```suggestion
When the MIDI Editor is set to Event List mode, REAPER presents a list with all the events in the current MIDI item.
When a note gets focus in the list, OSARA will play a preview of the focused note.
```

src/midiEditorCommands.cpp
```diff
@@ -994,6 +994,55 @@ void cmdMidiFilterWindow(Command *command) {
 	}
 }
 
+void maybePreviewCurrentNoteInEventList(HWND hwnd) {
+	auto focused = ListView_GetNextItem(hwnd, -1, LVNI_FOCUSED);
+	char text[50];
+	// Get the text from the type column (4).
+	ListView_GetItemText(hwnd, focused, 4, text, sizeof(text));
+	if (strcmp(text, "Note") != 0) {
```
jcsteh, line 1002: While OSARA itself isn't translatable, I try to avoid any code which depends on REAPER's own localised strings. Is there some other way we can detect this? For example, I'm guessing notes are the only event type that have a length value?

src/midiEditorCommands.cpp
```diff
@@ -994,6 +994,55 @@ void cmdMidiFilterWindow(Command *command) {
 	}
 }
 
+void maybePreviewCurrentNoteInEventList(HWND hwnd) {
+	auto focused = ListView_GetNextItem(hwnd, -1, LVNI_FOCUSED);
+	char text[50];
```
jcsteh, line 999: If you're worried about this macro failing and not touching the string (which it seems you might be given the nulling later), you might want to ensure the first char of text is null here, since you're not explicitly zeroing it currently:
`char text[50] = "\0";`

src/midiEditorCommands.cpp
```diff
@@ -994,6 +994,55 @@ void cmdMidiFilterWindow(Command *command) {
...
+	// Get the text from the type column (4).
+	ListView_GetItemText(hwnd, focused, 4, text, sizeof(text));
+	if (strcmp(text, "Note") != 0) {
+		return;
+	}
+	fill(begin(text), end(text), 0);
```
jcsteh, line 1005: It should be sufficient (and IMO clearer) to do this:
`text[0] = '\0';`

src/midiEditorCommands.cpp
```diff
@@ -994,6 +994,55 @@ void cmdMidiFilterWindow(Command *command) {
...
+	// Get the text from the channel column (3).
+	ListView_GetItemText(hwnd, focused, 3, text, sizeof(text));
+	note.channel = atoi(text);
+	fill(begin(text), end(text), 0);
+	// Get the text from the parameter (note name) column (5).
+	ListView_GetItemText(hwnd, focused, 5, text, sizeof(text));
+	string noteNameWithOctave { text };
+	static const string noteNames[] = {"C", "C#", "D", "D#", "E", "F",
+		"F#", "G", "G#", "A", "A#", "B"};
```
jcsteh, line 1026: Note names can be customised, so this is going to fail in that case. See OSARA's getMidiNoteName function and its use of GetTrackMIDINoteName. I don't really know if there's a way to deal with that here?

jcsteh (changes requested):
Very clever. Nice.

readme.md
leonardder: I made this heading level 4 initially since it also has to do with the midi editor.

src/midiEditorCommands.cpp
leonardder: I will look in to this, good question.

src/midiEditorCommands.cpp
leonardder: Doesn't that only zero the first byte of the string? I"m especially worried about cases where calling ListView_GetItemText writes a string to the string array that's shorter than the string before, which means garbage in it. Or is this fear unfounded?

src/midiEditorCommands.cpp
leonardder: Ugh, problem here is that we're transforming from note name to pitch.
I've also considered finding a way to get the right note event and just use that to play the note, but I"m afraid we can only iter through the list and count all the notes, which is probably pretty slow.

Lo-lo78:
Hi Leonard, in my opinion you could think of making the notes say in 
mute without being reproduced, because in play it is not reproduced, 
however, a feedback of its existence is important, the reproduction when 
it is selected is not necessary. it doesn't have to be read as if it has 
velocity at 0 or volume at 0, you just don't play it; always if 
possible. The important thing is that OSARA reads the existence of the 
note when we pass over it, even if it is muted. Thank you very much!

src/midiEditorCommands.cpp
jcsteh: It does only zero the first byte and it does mean there could be garbage for a shorter string. However, if it's writing a shorter string, it should also write out a null terminator (\0), which is required for C strings. So, the only case you have to worry about is where it does nothing, in which case zeroing the first byte will be considered to be an empty C string, since any functions accepting C strings won't go past a null terminator.

src/midiEditorCommands.cpp
jcsteh: Hmm. I wonder if Cockos would be willing to give us an API function to convert event list view indexes to note/CC indexes?

In the meantime, I think we can still go ahead with this, since it's still useful in cases (probably the majority) where there aren't custom note names.

readme.md
jcsteh: I understand. However, the previous section is very specifically about moving through events in the piano roll. It should perhaps say "Piano Roll" instead of "MIDI Editor", but the problem is that this also applies to some other views, everything but the Event List.

src/midiEditorCommands.cpp
```diff
@@ -994,6 +994,53 @@ void cmdMidiFilterWindow(Command *command) {
 	}
 }
 
+void maybePreviewCurrentNoteInEventList(HWND hwnd) {
+	auto focused = ListView_GetNextItem(hwnd, -1, LVNI_FOCUSED);
+	char text[50] = "\0";
+	// Get the text from the length column (2).
+	// If this column is empty, we aren't dealing with a note.
+	ListView_GetItemText(hwnd, focused, 2, text, sizeof(text));
+	if (strcmp(text, "\0") == 0) {
```
jcsteh, line 1003:
```suggestion
	if (!text[0]) {
```

src/midiEditorCommands.cpp
```diff
@@ -994,6 +994,53 @@ void cmdMidiFilterWindow(Command *command) {
...
+	// Loop through noteNames in reverse order so the sharp notes are handled first.
+	for (int i = static_cast<int>(size(noteNames)) -1; i >= 0; --i) {
+		auto noteNameLength = noteNames[i].length();
+		if (noteNameWithOctave.compare(0, noteNameLength, noteNames[i]) != 0) {
+			continue;
+		}
+		int octave = stoi(noteNameWithOctave.substr(noteNameLength, string::npos));
+		note.pitch = ((octave + 1) * 12) + i;
+		break;
+	}
```
jcsteh, line 1034: What happens in the case where we don't match a note name (e.g. custom note names)? It looks like we preview the notes regardless. That seems potentially problematic.

src/midiEditorCommands.cpp
leonardder: > Hmm. I wonder if Cockos would be willing to give us an API function to convert event list view indexes to note/CC indexes?

Either that, or a function to convert custom note names back to pitch.

src/midiEditorCommands.cpp
leonardder: That's correct, as I didn't deal with that yet. I was searching for an opportunity to test these custom note names.

leonardder:
Yay, we're lucky. If a note is named, the note name will be appended to the note in the list, so when a note (e.g. c#4) is called `squeeze`, it is shown as `C#4 squeeze`.
Therefore, my code should cover all cases now. I'm pretty exited.

src/reaper_osara.cpp
```diff
@@ -3351,6 +3356,11 @@ void CALLBACK handleWinEvent(HWINEVENTHOOK hook, DWORD event, HWND hwnd, LONG ob
 			// screen scraping, which causes spurious messages to be reported.
 			annotateAccRole(hwnd, ROLE_SYSTEM_PANE);
 		}
+
+		bool focusIsMidiEditorEventsListView = isMidiEditorEventsListView(hwnd);
+		if (focusIsMidiEditorEventsListView) {
```
jcsteh, line 3361: nit: The variable doesn't really help here and is only used once.
```suggestion
		if (isMidiEditorEventsListView(hwnd)) {
```

jcsteh (approved):
Thanks.

PR #444, opened 2020-12-23
Bind OSARA's note preview to Options: Preview notes when inserting or editing

### LINKED ISSUE #440: Add an option to disable note previews in the event list
In Event List mode, is it possible to turn off note previews? If not, could Osara have an option to either toggle this, or have it set in Osara configuration?

leonardder:
Fixes #440

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -165,6 +165,9 @@ bool compareNotesByLength(const MidiNote& note1, const MidiNote& note2) {
 }
 
 void previewNotes(MediaItem_Take* take, const vector<MidiNote>& notes) {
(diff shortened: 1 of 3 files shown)


jcsteh (changes requested):
Thanks!

1. Since most of the time we just call previewNotes, would it be cleaner to just have previewNotes return early at the top if the toggle is enabled? The one exception is the event list preview, which does a bunch of other stuff, so that function should also return early at the top. Thoughts?
2. Please add a comment to the lines where the action id is used specifying the name of the action.

leonardder:
Done as requested.

jcsteh:
Now that previewNotes checks the toggle, it isn't needed anywhere else except the event list preview.

leonardder:
Ugh, I'm sorry. This was my intention, but something went wrong with git.

jcsteh:
I figured it might've been something like that. :)

jcsteh:
Messed up a merge and had to force push, so this is now master f4f18d52.

Issue #2, opened 2015-02-05
Reporting minutes and seconds correctly when navigating the project


marioloreti:
Hi, I've noticed that when I set the time ruler from the view menù to minutes:seconds Osara keeps announcing bars and beats.

I hope you can fix this. Thanks, keep up the good work.

Issue #1, opened 2015-02-05
Please allow Osara to interact with the master track


marioloreti:
It will be cool if Osara could interact with the master track as Rea Access does. Specifically Rea Access allows to insert fx and it can provide access to the master track related parameters by pressing shift+tab.

jcsteh:
Sadly, REAPER won't actually let you focus the master track in the normal way tracks get focus.

From what I recall, ReaAccess has a separate action for viewing fx parameters on the master track; i.e. it isn't the same command you use for fx parameters on normal tracks. Is that correct? I'll probably have to do the same in OSARA.

Regarding shift+tab, I assume that too is different to accessing parameters on normal tracks? Or is there some ReaAccess specific way of "focusing" the master track?

marioloreti:
Hi,

with Rea Access I focus the master track with shift+tab and I am  on the 
volume parameter. with shift+tab again I can explore all the other 
parameters.

About FX, yes, I can add these with a separated action, assigned to shift+f.
Il 05/02/2015 13:44, jcsteh ha scritto:
>
> Sadly, REAPER won't actually let you focus the master track in the 
> normal way tracks get focus.
>
> From what I recall, ReaAccess has a separate action for viewing fx 
> parameters on the master track; i.e. it isn't the same command you use 
> for fx parameters on normal tracks. Is that correct? I'll probably 
> have to do the same in OSARA.
>
> Regarding shift+tab, I assume that too is different to accessing 
> parameters on normal tracks? Or is there some ReaAccess specific way 
> of "focusing" the master track?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/jcsteh/osara/issues/1#issuecomment-73040247>.
>

-- 
Mario Loreti

Iscriviti a Dropbox, il servizio per condividere i tuoi file gratis con uno spazio pari a 3500mb. Clicca:
http://db.tt/W6C56gy

marioloreti:
Hi,
I wanted to let you know that it is already possible to add an fx to the 
master track using the action called view fx chain for master track. 
What rea-access did is the possibility to adjust the parameters for 
those fx, the same way it can be done for the current track. The key 
rea-access assigned to that action is CtrlShiftP and the action is 
called "parameter plugin for master track". The tab thing that we have 
on rea-access is a virtual mode that allows us to control volume, pan, 
and it can be useful when checking where things are, for example, when 
we normalize one item we can check at what volume it gets. In 
[...]

PR #445, opened 2020-12-24
Correct reporting of regions.

### LINKED ISSUE #442: Incorrect announcement of Regions
The semicolon and apostrophe keys move from marker to marker and also move from region to region.  When regions overlap, these keys provide incorrect feedback of the region name as well as the start and end points of the region.  Here is an example:

1.  Set edit markers at the beginnings of measures 10 and 20 using the left and right brackets.
2.  Hit Shift+R to create a region from the selected range.
3.  Hit shift+alt+R to label this "Region #2".
4.  Set edit markers at the beginnings of measures 1 and 15 using the left and right brackets.
5. Create a region by hitting shift+R.
6. Hit shift+alt+R to label the region "Region #1
7. Hit the semicolon and apostrophe keys to navigate through the markers associated with the beginnings and ends of the regions.  Here is what you will find in order::
measure 1: "start of region #1"
measure 10: "start of region #2"
measure 15: "end of region #2"
measure 20: "end of project"
Note: measure 15 is incorrectly labeled as the end of region #2
9.  You can verify the start and end of each region by hitting control+shift+alt+R to get into the Markers and Region Manager.  The start and end of each region is properly labeled in this dialog.
10. Now use the semicolon and apostrophe keys to move to measure 15.  This is reported incorrectly as the end of region #2.
11. Hit delete.
12. Instead of deleting region #2 you will find that region #1 is deleted.
Summary: The navigation commands are providing incorrect info for the region markers which can lead to errors in deleting the desired region.
10. 9.

jcsteh:
1. Report "region start" and "region end" instead of just "region", since the marker navigation commands move to end positions as well.
2. Correctly detect the end position of a region, even where there are overlapping regions.

Fixes #442.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -694,12 +694,39 @@ void postCycleTrackFolderCollapsed(int command) {
 	outputMessage(getFolderCompacting(track));
 }
 
+int findRegionEndingAt(double wantedEndPos) {
+	double start = wantedEndPos;
+	for (; ;) {
+		if (start == 0) {
+			return -1;
+		}
+		// GetLastMarkerAndCurRegion doesn't return a region at its end position,
+		// so subtract a bit.
+		double tempPos = max(start - 0.001, 0);
+		int region;
+		GetLastMarkerAndCurRegion(nullptr, tempPos, nullptr, &region);
+		if (region < 0) {
+			return -1;
+		}
+		double end;
+		EnumProjectMarkers(region, nullptr, &start, &end, nullptr, nullptr);
+		if (end == wantedEndPos) {
+			return region;
+		}
+		// If there are overlapping regions, GetLastMarkerAndCurRegion will return
+		// the region which starts nearest to the given position. There might be a
+		// region which starts earlier but ends earlier. The next iteration will
+		// try the region just prior to this region's start position.
+	}
+	return -1;
+}
+
 void postGoToMarker(int command) {
 	ostringstream s;
 	int marker, region;
 	double markerPos;
 	double cursorPos = GetCursorPosition();
-	GetLastMarkerAndCurRegion(0, cursorPos, &marker, &region);
+	GetLastMarkerAndCurRegion(nullptr, cursorPos, &marker, &region);
 	const char* name;
 	int number;
 	if (marker >= 0) {
@@ -712,15 +739,30 @@ void postGoToMarker(int command) {
 				s << "marker " << number << " ";
 		}
 	}
-	if (region >= 0) {
-		EnumProjectMarkers(region, NULL, NULL, NULL, &name, &number);
-		fakeFocus = FOCUS_REGION;
-		if (name[0])
-			s << name << " region ";
-		else
-			s << "region " << number << " ";
-	}
 	double start, end;
+	if (region >= 0) {
+		EnumProjectMarkers(region, nullptr, &start, &end, &name, &number);
+		if (start == cursorPos) {
+			fakeFocus = FOCUS_REGION;
+			if (name[0]) {
+				s << name << " region ";
+			} else {
+				s << "region " << number << " ";
+			}
+			s << "start ";
+		}
+	}
+	region = findRegionEndingAt(cursorPos);
+	if (region >= 0) {
+		EnumProjectMarkers(region, nullptr, nullptr, nullptr, &name, &number);
+		fakeFocus = FOCUS_REGION;
+		if (name[0]) {
+			s << name << " region ";
+		} else {
+			s << "region " << number << " ";
+		}
+		s << "end ";
+	}
 	GetSet_LoopTimeRange(false, false, &start, &end, false);
 	if (start != end) {
 		if (cursorPos == start)


jcsteh:
@ptorpey, please give the test build in this pull request a try when it completes in a few minutes. It won't fix deletion, but it should fix reporting. Deletion should be fixed separately, potentially in REAPER itself.

ptorpey:
Microsoft Defender is complaining about downloading this because of a potential virus.  Should I just turn off MS Defender?  I assume this is safe, but just checking.

 

Thanks.

 

--pete

ptorpey:
Just to be on the safe side, I did a full backup of my system before turning off my anti-virus to download and install this build.  Yes, probably not necessary but…

 

Anyway, start and ends of regions are now correctly identified for both adjoining regions and regions that are not adjacent.  Very nice.

 

One small issue for thought depending on how easy it is to fix:

 

When I have two regions adjacent to each other, let’s say section #1 followed by section #2, here is the feedback I get as I hit apostrophe several times to move through the regions:

 

section #1 region start bar 5 
section #2 region start section #1 region end bar 10 
section #2 region end bar 15 



Note that the beginning of section #2 is announced before the end of section #1.  Logically it might make more sense to announce the end of section #1 before the beginning of section #2 when one is moving in that direction.

 

Interestingly, when I move the other way by hitting semicolon I get:

 

section #2 region end bar 15 
section #2 region start section #1 region end bar 10 
section #1 region start bar 5 



So the beginning of section #2 is properly announcd before the end of section #1 going in the reverse direction.

 

One more note.  Using JAWS you can see that I get a spoken line like:

 

section #2 region start section #1 region end bar 10

 

I assume that this really wanted to say:

 

section #2 region start bar 10 section #1 region end bar 10

 

but that the complete message about section #2 starting at bar 10 got truncated by the speech synthesizer because the second message about coming to the end of section #1 crashed into it.  That may not be fixable since it could be a timing issue.

 

Anyway, this is much better feedback than we had before since now we can reliably tell where regions start and end.

 

Thanks a bunch.  Let me know if you want me to do any more testing.

 

--Pete

jcsteh:
> Microsoft Defender is complaining about downloading this because of a potential virus.

Yeah, Defender seems to be overzealous concerning OSARA installers unfortunately.

> Logically it might make more sense to announce the end of section #1 before the beginning of section #2 when one is moving in that direction.

I follow your logic, but on the flip side:

1. If you have adjacent regions and you're moving between them, it's likely you care more about the start of a region than the end. If the end is reported first, you have to wait until that is read before you know what region you've moved into. That's less efficient for fast movement.
2. While possible, your suggestion is not trivial to implement.

> Using JAWS you can see that I get a spoken line like: section #2 region start section #1 region end bar 10 I assume that this really wanted to say: section #2 region start bar 10 section #1 region end bar 10 but that the complete message about section #2 starting at bar 10 got truncated by the speech synthesizer

There is no truncation; this is intentional. Those commands can only move to a single position each press, so reporting "bar 10" twice is redundant.

ptorpey:
1.	If you have adjacent regions and you're moving between them, it's likely you care more about the start of a region than the end. If the end is reported first, you have to wait until that is read before you know what region you've moved into. That's less efficient for fast movement.

[PT] By that logic one would be more interested in hearing that one is moving into a region that is earlier in the project before moving out of a later region and hearing that you are at the beginning of that later region when moving backwards through a project.  But when moving backwards the region that is later in the project is still announced prior to the next region earlier in the project that one is moving into.  The way this works now, the latest region in the project is announced first no matter which way one is navigating with adjacent items.

2.	While possible, your suggestion is not trivial to implement.

[PT] To me this is the stronger argument for keeping things the way they are now.  At least this is giving much more useful feedback than before.  As you say, a lot comes down to user preference and  sometimes one just has to make a decision to do things one way or another.  Folks will eventually get used to the chosen solution.

Using JAWS you can see that I get a spoken line like: section #2 <https://github.com/jcsteh/osara/issues/2>  region start section #1 <https://github.com/jcsteh/osara/issues/1>  region end bar 10 I assume that this really wanted to say: section #2 <https://github.com/jcsteh/osara/issues/2>  region start bar 10 section #1 <https://github.com/jcsteh/osara/issues/1>  region end bar 10 but that the complete message about section #2 <https://github.com/jcsteh/osara/issues/2>  starting at bar 10 got truncated by the speech synthesizer

There is no truncation; this is intentional. Those commands can only move to a single position each press, so reporting "bar 10" twice is redundant.

[PT] Sounds good.  Thanks and happy holidays!

 

--pete

jcsteh:
I don't quite follow your thoughts regarding moving backwards. When moving backwards, I'd argue you're still usually more interested in moving to the start of a previous region, rather than the end. For example, if you have a song with adjacent regions verse 1 and chorus 1 and you move back by region, I'd think you more likely care about the start of chorus 1 and the start of verse 1 when moving backwards and vice versa when moving forwards. Sure, there are occasions where knowing about the end is useful (e.g. where there is space between regions), but my thinking is adjacent is more common. Granted, I have no evidence for that assertion, so it could be entirely wrong.

As you say, a decision had to be made either way. We can always adjust this later if repeated user feedback reveals productivity problems. FWIW, I'm open to adjusting, but let's give it time some to bake.

jcsteh:
Edited my previous comment to clarify some confusion. :)

ptorpey:
Good point.  I was mostly thinking analytically about logically moving from start to end of one region and then start to end of the next region when moving forward in a project.

Hearing the start of the first region followed by the start of the second region followed by the end of the first region and then the end of the second region just sounded a bit off to me.

 

But, as you say, in practice the method you implemented might work better.

 

In any case, the feedback now is much better and can always be changed if folks think it should be another way based on actual experience using regions.

 

--pete

PR #447, opened 2020-12-26
When moving between CC events, properly announce non control events

### LINKED ISSUE #436: Unable to use previous and next CC actions to move to events that are not CCs
The OSARA: Move to previous CC and OSARA: Move to next CC actions seem to treat all events that are not notes as CCs. So, if an item contains pitch bend, aftertouch, or program change messages, incorrect numbers and values are reported.

leonardder:
Fixes #436

Note that it looks like that the increase/decrease actions don't do anything for poly aftertouch and program. I think that's a REAPER bug that might need a report to Cockos.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -810,7 +810,7 @@ void moveToCC(int direction, bool clearSelection=true, bool select=true) {
 	s << formatCursorPosition(TF_MEASURE) << " ";
 	if (cc.message1 == 0xA0) {
 		s << "Poly Aftertouch ";
-		// Note: separate the note and value with two spaces to avoid treatment as thausands separator.
+		// Note: separate the note and value with two spaces to avoid treatment as thousands separator.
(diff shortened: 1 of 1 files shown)


src/midiEditorCommands.cpp
```diff
@@ -805,8 +808,23 @@ void moveToCC(int direction, bool clearSelection=true, bool select=true) {
 	fakeFocus = FOCUS_CC;
 	ostringstream s;
 	s << formatCursorPosition(TF_MEASURE) << " ";
-	s << getMidiControlName(take, cc.control, cc.channel) << ", ";
-	s << cc.value;
+	if (cc.message1 == 0xA0) {
+		s << "Poly Aftertouch ";
+		// Note: separate the note and value with two spaces to avoid treatment as thausands separator.
```
jcsteh, line 813: Typo
```suggestion
		// Note: separate the note and value with two spaces to avoid treatment as thousands separator.
```

jcsteh:
Messed up a merge and had to force push, so this is now master b94a6c47.

PR #477, opened 2021-02-17
Initial support for translation of messages using tinygettext and fmt.

jcsteh:
Re #13.

PR #485, opened 2021-02-25
Try to add multilingual support to osara's installer

dingpengyu:
This is the extended pull request for #477 
Try to add multilingual support to osara's installer.


jcsteh:
Note that there are strings in the installer that will need to be translated; e.g. the title, the options for standard/portable/key map, etc. I honestly can't remember how this is supposed to be done; it's been a very long time since I had to deal with NSIS localisation.

PR #502, opened 2021-03-30
Support LUFS measurement in the Peak Watcher.

### LINKED ISSUE #166: Request for Option in Peak Watcher to report LUFS, or RMS values
Requesting additional functionality within Peak Watcher such that the mode could be set to either RMS over a set or configurable window, and/or LUFS momentary /short term / integrated since last reset, or the existing peak value. You would pick the mode within the dialog and than the existing keys could be used for reporting back the results  

There's good support for LUFS and RMS from SWS offline analysis, however no way of grabbing the values in real time.
Ideally, LUFS would be an option, but no idea of the difficulty of implementation  or availability of the appropriate API.

jcsteh:
This is achieved using the "JS: EBUR128 Loudness Measurement V2.11 (TBProAudio)" effect, which must be installed for this to work.
If a LUFS type is selected in the Peak Watcher, this effect will be added to the track if it wasn't already.

Remaining work:

- [x] Reset integrated LUFS.
- [x] Remove the effect when Peak Watcher is disabled if we added it.
- [x] Don't report channels for LUFS, since it seems there aren't separate measurements for left and right.
- [x] Fix the positioning in the GUI. Ug.
- [x] Include the level types in the translation template.
- [x] Update the readme.

Fixes #166.


jcsteh:
@ScottChesworth, I'd be keen to get your thoughts on this. There are two problems with this:

1. It can't reset the integrated LUFS. That has to be done via an inaccessible icon in the GUI. One way might be to remove and re-add the effect when Peak Watcher is reset, but... yuck.
2. When Peak Watcher is disabled for that track, it doesn't remove the effect. I can try to implement this, though.

jcsteh:
Remaining work:

- [ ] Reset integrated LUFS.
- [ ] Remove the effect when Peak Watcher is disabled if we added it.
- [ ] Don't report channels for LUFS, since it seems there aren't separate measurements for left and right.
- [ ] Fix the positioning in the GUI. Ug.
- [ ] Include the level types in the translation template.

ptorpey:
Nice job.  This seems to work as expected using Reaper 6.26.

 

Q: Scott complained about the persistent chatter of speech.  I noticed that when using the Peak Watcher the meter values are also displayed in braille.  Unfortunately, this info appears on the right side of the display after the bar/beat and track number info and scrolls off the display.  I wonder if it is possible to just display the meter values when using the Peak Meter and folks with a braille display can silence speech to just read the display as audio is played?

 

Also, to save braille display real estate, instead of writing out “channel 1”, etc. perhaps the display can read “ch 1” or something similar.

 

Anyway, nice addition to include the various LUFS values.

 

--Pete

jcsteh:
>     1. It can't reset the integrated LUFS. That has to be done via an inaccessible icon in the GUI. One way might be to remove and re-add the effect when Peak Watcher is reset, but... yuck.

That's what I did.

>     2. When Peak Watcher is disabled for that track, it doesn't remove the effect. I can try to implement this, though.

It does now.

jcsteh:
@ptorpey, I have no control over what gets brailled with JAWS. That's managed entirely by the Snowman's scripts.

ptorpey:
I have no control over what gets brailled with JAWS. That's managed entirely by the Snowman's scripts.

[PT] Thanks.  I’ll check with Jim and/or look at the scripts myself.

 

--pete

Lo-lo78:
Great job! Since with Reaper 6.26 we have accessibility for RMS, on Peak 
Watcher, we could put RMS in the combo box as well. This way we have 
completed all forms of audio stream analysis. For example:

Peak Watcher Dialogo

Level type:

peak dB

RMS

integrated LUFS

momentary LUFS

short term LUFS

So we are truly at the top!

ScottChesworth:
I I'm not sure about adding RMS to the "Level type" combo box @Lo-lo78. Seeing as RMS is only provided for the master track, I wonder whether the easiest UX when configuring Peak Watcher would be to rename the option that's currently available to "Master Peak", then add a new "Master RMS" option to the "First Track" and "Second Track" combo boxes, instead of adding RMS to the "Level type" combo box? My thinking is that it's not obvious RMS is only supported for the master track unless a user is nerdy enough to follow change logs or the discussion we've been having, and I think only Master being available in both combo boxes once RMS is selected could be kinda confusing for a new user who's experimenting with different loudness types. An alternative strategy to increase clarity could be to name the RMS option "RMS (master track only" if you'd prefer to add it to the "Level type" combo box, but this way, it wouldn't be possible to choose to monitor peaks for the first followed track and Master RMS for the second.

@jcsteh, would you like me to reach out to the developer of ebu128 to see if he can implement a cleaner way for you to reset its meters?

ScottChesworth:
BTW @jcsteh, my feet are getting kinda cold here, what you've done this week keeps blowing my socks off. :)

ScottChesworth:
Me again (full of questions today)!
@jcsteh, I've got two Peak Watcher language tweaks I'd like to put forward and have also found a small bug that I can reproduce. Do you want those submitted here seeing as you're overhauling Peak Watcher, or opened as separate issues?

jcsteh:
Unless they're specifically related to LUFS (or unless they are regressions which don't happen in main line OSARA snapshots; i.e. I broke something here), please file other issues separately. I may still choose to implement them here, but if I don't, it's much easier than trying to untangle them from this PR. Thanks.

Unless it's causing user visible problems, I think I can live with removing and re-adding the effect when resetting. I have to do this when doing other kinds of reset (e.g. switching tracks or modes), so it's actually easier this way, since I use the same code path for both kinds of reset.

jcsteh:
Let's deal with master RMS as a separate issue, since there are open questions here that still need to be resolved. However, one question has been raised here which also impacts LUFS. Do you explicitly want to be able to monitor the two different tracks with different level types (LUFS on master, dB peak for the second track)? I guess I can understand why you'd do this, but it seems really confusing from a UX perspective. We'd end up needing to introduce two level type combo boxes, which feels kinda horrible. Conflating the level type with the track also feels pretty nasty and would result in a mess of options to scan through.

ScottChesworth:
> Do you explicitly want to be able to monitor the two different tracks with different level types (LUFS on master, dB peak for the second track)?
> 
What you've done so far is already a game-changer here, but if you've got the time/energy to keep going, yes, being able to mix level types would be an addition that I'd definitely make use of.
> 
> it seems really confusing from a UX perspective. We'd end up needing to introduce two level type combo boxes, which feels kinda horrible. Conflating the level type with the track also feels pretty nasty and would result in a mess of options to scan through.
> 
What if the first four items in Peak Watchers tab order went like this:
"First track" combo box (landing position when Peak Watcher dialog opens)
"Level type for first track" combo box
"Second track" combo box
"Level type for second track" combo box
I know that's one extra combo box than what we have now, but I think the increased functionality would be worth it.

ScottChesworth:
Language tweak in light of LUFS support: can "Hold peaks" be renamed to "Hold highest level"?

ScottChesworth:
Reporting bug with current LUFS implementation:
1. Enable "Report FX when moving to tracks/takes" in OSARA Configuration dialog.
2. Create a track, open Peak Watcher and choose any LUFS type
3. Still in Peak Watcher, choose to follow current track.
4. Back in the TCP, create a new track.
5. Move between those two tracks, and notice how when you switch track selection, the presence of EBU128 isn't reported.

jcsteh:
>     5. Move between those two tracks, and notice how when you switch track selection, the presence of EBU128 isn't reported.

That's because the effect is inserted after you move to the track, not before. I can't really "fix" that. If you open the FX chain or OSARA FX Parameters, I'm fairly sure you'll see it's there.

jcsteh:
One problem with mixing level types is that if you set both tracks to the same (or one is follow and the other is the current track), you won't know which type is being reported. I can append the type to the report (level -5 momentary LUFS), but that's very verbose.

jcsteh:
Ug. One other issue with mixing types: the levels are quite different, so I'm not sure having a single notification threshold makes much sense. You might want to know about a -1 peak dB, but a -13 momentary LUFS. Adding separate notification levels for each type is going to make this even more complicated.

jcsteh:
I've already implemented the two level types, but given the above concerns, I'm not sure how much sense it makes to keep this. Which is unfortunate given I've spent the last couple of hours on it. :(

ptorpey:
Q: I wasn’t sure what “follow current track” meant.  At first I thought this was to follow the track that had focus when the Peak Watcher was configured, regardless of which track one moved to later.  But from this thread it seems like the “follow current track” setting actually changes which track the meter is monitoring as one arrows from track to track.  If so, that is really nice!

 

--Pete

jcsteh:
> it seems like the “follow current track” setting actually changes which track the meter is monitoring as one arrows from track to track.

That's correct. If you explicitly want to monitor the current track, you can choose the specific track in Peak WAtcher.

jcsteh:
Another thing to think about is what OSARA should do if the EBUR128 effect isn't installed. Right now, it will just silently not retrieve any useful levels. I guess it should probably report an error? What makes this particularly painful is that we have no way of knowing whether the effect is installed until we try to add it to a track, so preventing the user from selecting those options or displaying an error as soon as the user presses OK and then throwing them back into the Peak Watcher dialog isn't an option.

I guess the big question is: given how tricky this will be to implement, can we live without this error reporting for now or does it need to be implemented before we merge this?

ScottChesworth:
> 
> 
> I've already implemented the two level types, but given the above concerns, I'm not sure how much sense it makes to keep this. Which is unfortunate given I've spent the last couple of hours on it. :(
> 
Pants! Hadn't considered that. Sorry I missed it man. Seeing as you've done the work already, I reckon we should all at least use this in situ for a while to see how it fairs before we pick it apart any further.

> I guess the big question is: given how tricky this will be to implement, can we live without this error reporting for now or does it need to be implemented before we merge this?
> 
Some sort of error report would be good to have, yeah. A couple of ideas:
1. Would it be any easier to deliver a report to the screen reader if EBUR128 isn't found once the user is back in the TCP, an off screen prompt instead of displaying a dialog box?
2. Appending something like "(requires js ABUR128 from tb-software.com)" to each LUFS related option in the Level type combo boxes would make users aware of the requirement. That's more repetitive speech output than is ideal of course, but hey, at least it'd make the dependency obvious.

If it helps, I don't think there's any huge rush to implement this part of the equation. Everyone I've ever known to be interested in LUFS reporting is either a power user or someone who's got a decent level of technical knowledge, and a handful of 'em are here on the thread already.

ptorpey:
How about a check at Osara startup to see if that plug-in is installed and setting an appropriate flag?

 

--pete

Lo-lo78:
Hi, nice idea to separate the first track from the second track! Yes, 
it's true, the accessibility on the RMS is only on the master; if the 
Cockos extended to the tracks it's really ok! Too bad we can't integrate 
the ebur128_analysis source code into OSARA, legally it can be done but 
I don't think it's possible in a simple way. The JS is not easy to find 
on the web; we could put it in a repository, so when OSARA reports an 
error, we could put a button that downloads the JS: ebur128_analysis, 
downloaded, there should be a warning saying to restart Reaper for the 
skan; for the JS the skan is automatic even if in the Reaper preferences 
the skan at startup is not activated for the VST. It would be perfect if 
JS: ebur128_analysis were inside OSARA, transformed into an OSARA 
action, used as a module for the functioning of the 'action OSARA: View 
Peak Watcher Congratulations to all! 👍

ptorpey:
Scott said:

Language tweak in light of LUFS support: can "Hold peaks" be renamed to "Hold highest level"?

[PT] Along those same lines, when tabbing to the radio buttons for how peaks should be reported, arrow down to the last item:

“Hold highest level: for (ms): radio button‑ checked“

Now tab once to the edit field.  On this edit field JAWS announces:

“Hold highest level: When level reaches: Edit”

I’m guessing that this is because the edit field doesn’t have a label and JAWS is picking up the label of the Threshold edit field.  Interestingly, NVDA acts correctly here and just announces the edit field.

I wonder if that could be fixed for JAWS by giving the edit field a label?  Can be a bit confusing for folks using JAWS, but certainly not a deal breaker.

 

--Pete

jcsteh:
> How about a check at Osara startup to see if that plug-in is installed and setting an appropriate flag?

As I noted, you can't check whether an effect is installed until you try to add it to a track. There's no other API. I guess you could check the file system, but it could be in any subdirectory, so at that point, you're doing your own effect scan.

> Too bad we can't integrate the ebur128_analysis source code into OSARA, legally it can be done but I don't think it's possible in a simple way.

I've actually considered bundling it with OSARA. The installer could easily create an Effects/OSARA directory in the REAPER resource directory and toss it in there. I'm not sure how well REAPER will cope with two effects having the same name, though, which could happen if someone had already installed it. That needs to be investigated. I guess I could rename the effect to have OSARA in its name, but even though that's legal, it feels unethical.

> The JS is not easy to find on the web;

The readme includes a link to the TBProAudio download page, where it's easy enough to find. My bigger concern is users knowing how to install it.

> It would be perfect if JS: ebur128_analysis were inside OSARA, transformed into an OSARA action, used as a module for the functioning of the 'action OSARA: View Peak Watcher Congratulations to all! 👍

You can't implement an effect inside an OSARA extension.

jcsteh:
It looks like having two copies of the same effect in different directories is fine. It causes two copies of the effect to show up in the browser, but they are suffixed with the directory name and either will work just fine anyway.

Scott, given this, what do you think about just bundling the effect with OSARA and having it installed by the installer?

Lo-lo78:
I'm not Scott, that's ok for me! In the time Reaper puts the RMS in the 
tracks, we could also use: JS: Audio Statistics. This JS is default on 
Reaper. I also give a lot of importance to the RMS because it is anyway, 
however much now old as an analysis, it is an indicator of the volume of 
our project

Well done, go on like this! 👏

ptorpey:
I'm not Scott, that's ok for me! In the time Reaper puts the RMS in the 
tracks, we could also use: JS: Audio Statistics. 

[PT] Another possibility is to add another entry to the combo box in the Peak Watcher that says something like “RMS (Master only)” unless Cockos can be convinced to add the RMS capability to all tracks.  Might be simpler than worrying about error messages and the like for plug-ins that we can’t be sure the user has.

 

When the Peak Watcher could only watch one track I used to watch the Master track and then solo other tracks to see which one caused a meter issue.

 

--Pete

ScottChesworth:
> 
> 
> Scott, given this, what do you think about just bundling the effect with OSARA and having it installed by the installer?
> 
That seems like it's gonna be the cleanest way to go then, yeah. Nice!

Lo-lo78:
Hi James, I have been doing a mix academy for a few months now, we are 
told how to produce the file from the rendering; often the volume is 
asked in LUFS. I am getting along really well with the build of 
osara_pr502-502,898d7335; I noticed with great surprise that OSARA also 
in the rendering says the volume, I had never noticed it. Very useful! 
Here you can find the latest version: https: 
//www.tbproaudio.de/downloadGreetings!

Lo-lo78:
[EBUR128LMJSFX08032021.zip](https://github.com/jcsteh/osara/files/6289453/EBUR128LMJSFX08032021.zip)
I am attaching the latest version ...

jcsteh:
I don't want to consider merging this until we get some answers on the UX problems concerning being able to specify multiple types; namely this:

> One problem with mixing level types is that if you set both tracks to the same (or one is follow and the other is the current track), you won't know which type is being reported. I can append the type to the report (level -5 momentary LUFS), but that's very verbose.

and this:

> One other issue with mixing types: the levels are quite different, so I'm not sure having a single notification threshold makes much sense. You might want to know about a -1 peak dB, but a -13 momentary LUFS. Adding separate notification levels for each type is going to make this even more complicated.

Adding support for two level types and then removing it later is going to be really messy, so we need to get this correct now. My gut feeling is that we should just go back to having a single level type. It's slightly less functional, but has far less potential UX confusion. But I'll leave that decision to the experts.

ptorpey:
I am certainly not an expert, but my feeling is that besides the added complexity of the UI, the difference in the levels generally seen by LUFS versus peak makes following multiple types of meters impractical.

 

In fact I wonder if it is even worthwhile following multiple tracks at once considering the possibly confusing and abundant amount of speech feedback.  One approach I have used in my limited work with this new feature is to have the meter set to “follow current track” and put focus on the master track as I review audio.  If the threshold then goes over a specified level then I can easily review the section of audio by arrowing to each other track until I find the track that was responsible for causing the problem.

 

BTW, on a slightly different topic, I would like to find a way of hooking into the output of the Peak Watcher in a JAWS script so that these meter readings could be displayed silently in braille as I did for the “meter display” mode in the braille support I added for Samplitude.  Looking over the Peak Watcher.cpp code it seems like this might be possible using the GetSpecificTrackFXParam function but, although I’ve done lots of C programming years ago, I’m not as familiar with C++ and the process flow in Windows programs.  If this might be possible and you have any hints or pointers for me to follow up I would be willing to give it a shot.

 

Anyway, it sure is great to have these other meter types available and applicable to the output of any track.  

 

--Pete

jcsteh:
I'm making an executive decision here so we can move this forward. We're going with a single level type.

PR #506, opened 2021-04-07
Use keyWindow instead of mainWindow when posting Mac a11y announcements.

### LINKED ISSUE #149: OSARA on MacOS with VoiceOver won't speak when REAPER is full screen
As I installed OSARA on Windows it works perfectly including the very helpful and necessary speech feedback. On MacOS instaid, the installation process also works fine, files are copying and keymap seems to work also. However, VoiceOver won't speak as I am known by Windows. Is there anything to do? I found nothing about this issue. Maybe I have forgotten to set anything other in VoiceOver? So if I launch me Reaper, I will be informed about version and license information. Pressing the shortcut for OSARA configuration window, it opens up, there is speech output checked.

jcsteh:
I'm hoping this might fix OSARA messages on mac when REAPER is full screen, but I really have no clue. It's just a hunch and a completely untested one at that.

@pitermach, @TimNoonanVoice, @kyleman, would you mind testing a build from here to see if it helps?

Fixes #149 (maybe).

### DIFF
--- a/src/osxa11y.mm
+++ b/src/osxa11y.mm
@@ -35,6 +35,6 @@ namespace NSA11yWrapper {
   NSDictionary *announcementInfo = @{NSAccessibilityAnnouncementKey : message,
 				     NSAccessibilityPriorityKey : @(NSAccessibilityPriorityHigh)};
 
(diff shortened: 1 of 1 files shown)


pitermach:
Good news! I just did some testing and everything  works great in and out of fullscreen even with no other windows open. Tested both in the Intel and ARM versions of Reaper.

ScottChesworth:
Wew, that hunch packed a punch!

PR #508, opened 2021-04-22
Allow Peak Watcher level type to be specified independently for first and second track.

jcsteh:
During the work on LUFS support in #502, I added the ability for the level types to be set independently for the first and second tracks. However, this resulted in some UX concerns:

> One problem with mixing level types is that if you set both tracks to the same (or one is follow and the other is the current track), you won't know which type is being reported. I can append the type to the report (level -5 momentary LUFS), but that's very verbose.

> One other issue with mixing types: the levels are quite different, so I'm not sure having a single notification threshold makes much sense. You might want to know about a -1 peak dB, but a -13 momentary LUFS. Adding separate notification levels for each type is going to make this even more complicated.

@ptorpey  added:

> I am certainly not an expert, but my feeling is that besides the added complexity of the UI, the difference in the levels generally seen by LUFS versus peak makes following multiple types of meters impractical.

As we weren't able to resolve these concerns, I opted to drop this so we could get the LUFS support merged. I'm pushing this work here for potential further discussion/iteration.


GarthHum:
Really glad to see this merged. 
My own preference would be to retain the ability to set each monitor to separate units, one on peak, and one on short term for example. I get this would involve a more complex UI.
Use case for me is to be able to just set the peak watcher to look at the master track for both monitor channels and have one set to peak and the other to short term or integrated. Ideally being able to set notification levels for each independently.  
Suggest renaming to Monitor 1 and Monitor 2 to avoid some confusion I see with the current dialog, from some new users. 
Suggest The UI be handled something like below. 
A popup for what track monitor one  references, then popup for what kind of feedback monitor one provides, then value for auto notification, and leave blank for none. And then repeat for Monitor two 
The re-set options could then continue to exist as they currently do or I suppose could be set inline independently for each monitor, but don’t feel this is as needed as the other options.
As for ambiguity with what value is being reported, personally not too concerned about this, and would probably prefer to keep the verbiage down, however adding just the words peak integrated, momentary, etc, might not be to intrusive. 
The fact that peak values are so different from the other LUFS values would be a pretty clear indication of what was being reported.

jcsteh:
To be clear, LUFS support has been merged, but this PR (independent levels) has not.

I'm not sure "Monitor 1" and "Monitor 2" are any clearer. It isn't necessarily obvious that this refers to tracks and this could be confused with input monitoring.

As for specifying the levels separately, remember that there are also checkboxes for each channel. While these aren't relevant for LUFS, they are for other types. So, we'd also have to duplicate these checkboxes for the second track, which could get pretty confusing/cluttered.

If we don't report the type after each value when watching, the problem is that you won't know which type is being reported if, for example, you're watching integrated LUFS and momentary LUFS on the same track. And while LUFS and peak values might be obviously different to the experienced user, I don't think we should assume this for everyone. But then we're back to the verbosity problem... and this would make things much more verbose even for users who are just monitoring peaks for a single track, which is effectively worsening an existing experience.

ptorpey:
I just tried the Peak Watcher in the latest build.

 

I still see that one can select different meters for the two tracks being monitored.  Based on your last note I was expecting to see only one combo box for choosing what type of meter would be used.  There are still two combo boxes for making a different selection for each track.  Did I misunderstand?

 

BTW, I think that was the right decision because of the reasons you already stated.

 

--Pete

 

From: James Teh ***@***.***> 
Sent: Thursday, April 22, 2021 5:07 PM
To: jcsteh/osara ***@***.***>
Cc: ptorpey ***@***.***>; Mention ***@***.***>
Subject: [jcsteh/osara] Allow Peak Watcher level type to be specified independently for first and second track. (#508)

 

During the work on LUFS support in #502 <https://github.com/jcsteh/osara/pull/502> , I added the ability for the level types to be set independently for the first and second tracks. However, this resulted in some UX concerns:

One problem with mixing level types is that if you set both tracks to the same (or one is follow and the other is the current track), you won't know which type is being reported. I can append the type to the report (level -5 momentary LUFS), but that's very verbose.

One other issue with mixing types: the levels are quite different, so I'm not sure having a single notification threshold makes much sense. You might want to know about a -1 peak dB, but a -13 momentary LUFS. Adding separate notification levels for each type is going to make this even more complicated.

@ptorpey <https://github.com/ptorpey>  added:

I am certainly not an expert, but my feeling is that besides the added complexity of the UI, the difference in the levels generally seen by LUFS versus peak makes following multiple types of meters impractical.

As we weren't able to resolve these concerns, I opted to drop this so we could get the LUFS support merged. I'm pushing this work here for potential further discussion/iteration.

  _____  


  https://github.com/jcsteh/osara/pull/508


Commit Summary


*	Allow level type to be specified independently for first and second track.


File Changes


*	M readme.md <https://github.com/jcsteh/osara/pull/508/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15>  (10) 
*	M src/peakWatcher.cpp <https://github.com/jcsteh/osara/pull/508/files#diff-40f068621e7e7c3b2a029d94846c3a6ee5979aa7ba09529141958d007c43da73>  (59) 
*	M src/reaper_osara.rc <https://github.com/jcsteh/osara/pull/508/files#diff-6400c4c0f6390d18df0bb6a55944de040e984c596547bbd0f230a664fd9f78d6>  (36) 
*	M src/resource.h <https://github.com/jcsteh/osara/pull/508/files#diff-c0eb6faf4086c6ecf4360169c6471910a5c47625578d3fc225ae5ce219baa36d>  (3) 


Patch Links:


*	https://github.com/jcsteh/osara/pull/508.patch
*	https://github.com/jcsteh/osara/pull/508.diff

ptorpey:
Ah, didn’t see this note before posting my last comment.  I was surprised to see choices for two different types of meters at a time.  I guess you’ll be pulling that out of the UI later to avoid confusion?

 

--Pete

jcsteh:
I'm... very confused. I'm reasonably certain the stuff I merged only contains a single level selection. Are you trying a build from the snapshots page?

jcsteh:
To clarify, the builds available on this pull request do of course contain the multiple level selection. However, builds mainline builds on the snapshots page should only contain a single level selection.

ptorpey:
Sorry for any confusion. I downloaded the link from the last email. It was not the official release link on the GitHub page.



Sent from my iPhone

On Apr 22, 2021, at 10:47 PM, James Teh ***@***.***> wrote:

﻿
I'm... very confused. I'm reasonably certain the stuff I merged only contains a single level selection. Are you trying a build from the snapshots page?

ScottChesworth:
> 
> The fact that peak values are so different from the other LUFS values would be a pretty clear indication of what was being reported.
> 
That's one of those things that's obvious once you know it, but clear as mud if you don't. Given how soon newbies tend to start investigating Peak Watcher, I don't think we can assume they'll grasp that intuitively.

GarthHum:
Yep, get that Scot. Personally I’d just like to find a way of getting the functionality of having peak watcher report both peak and integrated values without having to go in and mess with it's settings. Totally a fan of a little more complexity with the benefit of some very useful functionality on this one.  
 
With respect to the verbosity, could we reasonably expect someone to know how they have setup peak watcher, and  so it’s not so important when reporting values they pole through the keystrokes to give the level type, however when OSARA reports automatically, more descriptive responses could be provided. 
Again, not really seeing any good clear way through but would really like the functionality.
Monitor might not be the best term however feel like there should be something better then first and second track

jcsteh:
To be clear, I'm only talking about auto reporting here. For on-demand querying, we don't need to change anything... although interestingly, the extra verbosity is far less problematic with on-demand.

Why do you find "first track", "second track" confusing given the context of the Peak Watcher dialog?

ptorpey:
The other question we might ask is do people actually want to monitor two tracks at a time or two different meter types at a time?

 

One possibility that might be worth considering is to give the user the option of watching either the peak meter, or one of the LUFS meters, or both on a single track.  This could be done with check boxes to toggle the activation of each type of meter on/off.

 

A work flow that I have used is to have the Peak Watcher set to watch the “current track”.  Then I put focus on the Master track as audio is playing.  When I hear a peak value that I don’t like I can arrow to another track to see which track is responsible for causing the problem without having to get into the Peak Watcher again.

 

--Pete

jcsteh:
I'm fairly certain watching two tracks was an early request. Even if we went with a single track and checkboxes, that still doesn't solve the verbosity problem with auto reporting.

ptorpey:
Yes, I believe that you are correct, being able to monitor two tracks at a time was an early request.  I guess I’m wondering if, after all of this discussion and the trade offs involved that it is still worth considering or if there might be other options.  As I indicated in my last note, perhaps being able to monitor the “current track” on the master and then simply arrow down to another track might scratch that itch and make the UI more simple.

 

There hasn’t been a whole lot of feedback on this and I’m just wondering how folks feel after this has been rumbling around for a bit.  Maybe now that you’ve posted this to the RWP list you can get more useful feedback.

 

--Pete

GarthHum:
Personally I don’t find first track, second track confusing. It has sometimes been a little confusing for new users though. I think it is just a language thing as a project will actually contain a first and usually second track. Then they see that the on demand keystrokes are, for example,  Report Peak Watcher value for channel 1 of first track and even though they have just setup the first track to be set to follow current track or maybe the first track is set to track 2, they wonder if it is going to tell them about track 1 in the project. 
It’s a small thing but thought maybe could be made a little clearer somehow.

With respect to auto verses on demand reporting, I haven’t played enough to understand why additional verbosity on demand reporting would be less problematic than on auto reporting. I guess for my work flows, I rarely have the threshold for auto reporting set low enough that I have OSARA reporting much automatically. It is mostly only heard when I ask for it. In any case, you’re not talking about adding additional verbosity to the on demand reporting so it’s a mute point. 

To a previous point, you mentioned that we would have to put in the check boxes for the optional reporting of channel 1 and channel 2 for both first and second track and that this would also add to the complexity of the dialog. Agreed, but again I think it is worth the trade off. This said, with respect to these check boxes, I don’t think it is as much of a priority that they are able to be set per track, and could be left as global preferences as they are now, as are the preferences around value holding 

For me the priority is to be able to set which tracks are reported, which type of feedback is reported for each of those tracks, and the threshold for auto reporting. It is definitely a more complex dialog however would result in an extremely versatile and functional monitoring tool, I believe.

ptorpey:
With respect to auto verses on demand reporting, I haven’t played enough to understand why additional verbosity on demand reporting would be less problematic than on auto reporting. I guess for my work flows, I rarely have the threshold for auto reporting set low enough that I have OSARA reporting much automatically. It is mostly only heard when I ask for it. In any case, you’re not talking about adding additional verbosity to the on demand reporting so it’s a mute point. 

[PT] One does have to be careful with the levels one sets when auto reporting otherwise, as you say, there can be a tremendous amount of speech feedback.  If one were watching a LUFS level at -20 for example and it only went off once in a while, a peak meter on another track (or the same track) would be spilling out tons of speech.  I’ve inadvertently set the levels too low at times and, besides the fact that speech can’t nearly keep up with where one is in the audio, sometimes one can’t even shut speech up using the control key.

 

--Pete

 



To a previous point, you mentioned that we would have to put in the check boxes for the optional reporting of channel 1 and channel 2 for both first and second track and that this would also add to the complexity of the dialog. Agreed, but again I think it is worth the trade off. This said, with respect to these check boxes, I don’t think it is as much of a priority that they are able to be set per track, and could be left as global preferences as they are now, as are the preferences around value holding 

For me the priority is to be able to set which tracks are reported, which type of feedback is reported for each of those tracks, and the threshold for auto reporting. It is definitely a more complex dialog however would result in an extremely versatile and functional monitoring tool, I believe.

jcsteh:
I thought about this some more and I think I have come up with a UX that would solve most of the concerns here. Similar to bands in ReaEQ's UI, we could have tabs for the first and second watcher. All of the settings (including level type, notification threshold and hold) would be independently settable on both tabs.

The one thing this doesn't solve is verbosity if two different level types are set for the same track. The best I can come up with there is to report the type after the level ("-13 integrated LUFS"), but only if there are two different types for the same track.

The catch is that this is going to be super painful and time consuming to implement, so it's unlikely I'll get to it any time soon, if ever. Still, it'd be good to get thoughts on the UX in case I (or someone else maybe?) gets time to poke at this.

jcsteh:
To clarify, each tab would only allow you to specify one track, but there would be two tabs, hence two possible tracks (but with different settings for each).

GarthHum:
Sounds good to me.

ScottChesworth:
Liking the tabs idea a lot, named "First Track and Second Track" that would be nice and intuitive.

Justinmac84:
Having read through all of this, I'm with Garth. I would prefer function over form and complexity and versatility over unambiguity. I certainly don't think that we should be hesitant to implement something on the basis that some people may not understand it at first. Reaper already has some aspects that are hard to understand at first, but at least they're there. 

I would certainly prefer something over nothing so, if it's unlikely people are going to be able ever to get to a tab style UX, which does sound nice, I would prefer a more cluttered, single-tab UI if it's easier to implement. I really do feel that having something is better than having nothing at all. 

So regarding knowing which level is being reported, I only really see this being an issue if people have two different meters applied to a single track and, even then, not necessarily if they listen out for the track 1 track 2, though I appreciate that, if they're looking at the same track, this might be confusing. I also take the point that people may get confused with monitoring if the names of followed tracks were changed to monitors, although the reporting happens in the context of the Peakwatcher so such confusion may seldom be more than hypathetical. Anyway, if that is such a concern as to be an impediment, what about watcher 1 and watcher 2? But I definitely agree with Garth that, if you're the one that set up the Peakwatcher to achieve a specific purpose, surely you're going to know what levels are being reported? what about meter 1 and meter 2 come to that? 

As an on demand Peakwatcher user, I don't worry about auto-reporting all that much so can't speak to that really more than I already have.

All I know is that sighted people can monitor their audio in multiple ways at a glance and there are many use-cases when being able to do the same in a way that isn't time-consuming would be invaluable - monitoring true peak or Lufs on the master while monitoring peak or true peak on an individual track; monitoring true peak and Lufs simultaneously on the master; monitoring peak and true peak on a particular track. I find this last helpful to fix panning issues. If I have recorded something and there is an offset in the stereo field, I find it useful to do a peak check to work out how much that offset is or if it is an offset caused by a phase problem in one of the channels. That way I very quickly know by how much to adjust the panning, but I also have to render sounds so that they don't exceed certain true peak levels. Not having frequently to go in and out of the Peakwatcher dialogue to change things would be exceedingly helpful  

wish I'd seen all this back in April. Don't know what I was doing back then, but anyway.

ScottChesworth:
> All I know is that sighted people can monitor their audio in multiple ways at a glance...
>
Keep in mind that a glance can inherently gather more information than a stream of speech, because indicators such as colour, shading, positioning etc that we don't have equivalents for are present in the visual UX. OSARA can't compete with that, it's unavoidable that we're going to take a bit longer to ingest speech feedback where accuracy is the priority. When accuracy isn't the highest priority, the Accessible Peak Meter plug-in might fit that use case, but that presents its own UX challenges.

> there are many use-cases when being able to do the same in a way that isn't time-consuming would be invaluable.
>
I've never observed a sighted person change metering type in less than three clicks. Right now, for most changes we have three keystrokes. Alt+W to load Peak Watcher, first letter navigation to choose new report, Enter. Doesn't seem like we're taking a productivity hit to me.

To be clear, I'm not saying "don't change anything," and I'm not saying "Peak Watcher is as good as it can get" either, I'm just attempting to make the comparison fair. It's always easy to say "sighted people can do this faster" or "sighted people have more information available to them", but it's also always important to factor in that we're not working with concurrent streams when we relay information to users via screen reader software.

jcsteh:
I'd point out that beyond the work I've already done here (you can try out
a build above if you're interested), implementing this in a single tab UI
isn't trivial either. It's not like a cluttered single tab UI is easy and a
nice multi tabbed UI is hard. If anything, figuring out how to double all
the notification controls in a single tabbed UI in a way that will make any
sense to anyone is a pretty big challenge. Either way, there's a lot of
work to be done here and no one appears to have the time or inclination to
do it.

FelipeZanabria:
Perhaps something that might be useful is adding verbalization using tones, just like the vst Accessible peak meter plug-in does. I don't know if the best option is to code a js effect, or implement it in osara's own code. The tones can give a faster response.
The link on the page is at:
http://depic.eecs.qmul.ac.uk/apm/

ptorpey:
Yes, the Accessible Peake meter with its quick sounds is a very nice way of knowing when audio goes over a certain level without a lot of extraneous feedback.  Very nice concept.  I think the Accessible Peak meter can also monitor different types of peak values such as Peak, True Peak, RMS, LUFS, etc., all accessible through Osara’s parameters list.

 

--Pete

leonardder:
I think this is superseded by #623, right?

jcsteh:
Yes. I didn't close it because of uncertainty about whether the rewrite would fly, but I think we're past that point now.

PR #510, opened 2021-05-07
When moving to an envelope point, report its associated shape.

jcsteh:
@vick08, is this what you're after? (There should be an automated build here for you to test shortly.)


vick08:
I actually spotted the new branch a few hours ago and was able to compile and test the build.
All seems to be working as inntended so far!
Thanks a lot!

PR #517, opened 2021-05-20
Support for reporting track groups.

### LINKED ISSUE #316: Feedback when the track is in a group
Hello,
when I do alt + g and assign controls to the group on multiple tracks, if I scroll in the track view it is not reported by OSARA.
Do you think it is possible in tcp to signal when a track is in a group?
It would be very useful, track number, track name and then the rest of the information, in this case if it is in group 1, 2, 3 and so on.
Thank you!

jcsteh:
1. When navigating to a track, report if it is grouped.
2. Add OSARA: Report groups for current track action to report full group information.

Fixes #316.


Lo-lo78:
Excuse me, I did it all and tried the action; everything works 
perfectly! Congratulations!
👍 Greetings!

Lo-lo78:
James, way above my expectations! Really good!

In my opinion you can release the new OSARA release. Everything works!

Greetings!

PR #522, opened 2021-05-21
Midi velocity feedback fixes

ScottChesworth:
Action feedback for decreasing velocity by 1 and increasing by 10 was the wrong way round.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1142,10 +1142,10 @@ void postMidiChangeVelocity(int command) {
 				s << "velocity +1";
 				break;
 			case 40463:
-				s << "velocity -1";
+				s << "velocity +10";
 				break;
 			case 40464:
-				s << "velocity +10";
+				s << "velocity -1";
 				break;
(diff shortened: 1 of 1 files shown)


jcsteh:
Thanks! Nice to see you getting your toes wet. :)

So you correctly created a different branch for this pull request, which is awesome. However, this branch also includes the commit from your other branch. The two branches ideally should be independent. To do that, you'd need to checkout master, then create your new branch from there.

I'm going to rebase the branch for you and remove the earlier commit because learning these parts of git can be pretty overwhelming and I don't want to stall merging this when it's trivial for me to clean up. Please keep the above in mind for future branches, though.

leonardder:
Ough, I thought I fixed this in the past already. Turns out I didn't.

jcsteh:
Heh, I thought you did too. You fixed the action mappings in the map, which were also incorrect. But there's also the messages which get spoken when we generalise and those weren't fixed.

ScottChesworth:
> checkout master, then create your new branch from there.
> 
Got it. Will do.

PR #524, opened 2021-05-22
In OSARA parameters dialogs, pass function keys, anything with the control key or anything with both alt and shift through to the main window.

### LINKED ISSUE #523: Is it possible to insert mute and solo in the FX Parameters window?
Hello everyone!
When I try to do an EQ, a compression on the snare and so on, in the FX Parameters window I miss putting the solo on the track I'm working on and then removing the solo to hear how the snare sounds with the whole project.
Unfortunately, it is not possible to mute or solo the track in the FX Parameters window.
I find that when making a mix this can speed up the work a lot, without having to esc or close and then hear and remove the solo to see if the e-q or the compressor has been adjusted well.
If I use a plugin like a channel strip, which contains the entire channel of a mixer, then I find myself, preamp, gate, e-q, compressor and so on; in this case it is very important to be able to have a solo and mute toggle. A channel strip has a lot of parameters and getting in and out of the FX Parameters window can be very confusing and slow.
When making a mix, the speed to compare one sound with another is essential; the brain tends to find a balance, the more time passes between one sound and another and the more difficult it is to understand the differences.
For this, tests (A) and (B) are performed practically instantly.
Greetings!

jcsteh:
This necessitates closing the Parameters dialog if focus is moved out of it, since some commands might do that.

Fixes #523.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -430,44 +430,75 @@ class ParamsDialog {
 				DestroyWindow(dialogHwnd);
 				delete dialog;
 				return TRUE;
+			case WM_ACTIVATE:
+				if (LOWORD(wParam) == WA_INACTIVE) {
+					// If something steals focus, close the dialog. Otherwise, we won't
+					// unregister the key hook,  surface feedback won't report FX parameter
+					// changes and there will be a dialog left open the user can't get to
+					// easily.
+					PostMessage(dialogHwnd, WM_CLOSE, 0, 0);
+					return TRUE;
+				}
 		}
 		return FALSE;
 	}
 
 	accelerator_register_t accelReg;
 	static int translateAccel(MSG* msg, accelerator_register_t* accelReg) {
-		// We handle key presses for the slider ourselves.
 		ParamsDialog* dialog = (ParamsDialog*)accelReg->user;
-		if (msg->message != WM_KEYDOWN || msg->hwnd != dialog->slider) {
-			return 0;
+		if (msg->message != WM_KEYDOWN && msg->message != WM_SYSKEYDOWN) {
+			return 0; // Not interested.
 		}
-		double newVal = dialog->val;
-		switch (msg->wParam) {
-			case VK_UP:
-			case VK_RIGHT:
-				newVal += dialog->param->step;
-				break;
-			case VK_DOWN:
-			case VK_LEFT:
-				newVal -= dialog->param->step;
-				break;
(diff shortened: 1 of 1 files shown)


jcsteh:
@Lo-lo78, @ScottChesworth, please give this a try. Also, it'd be good to get this tested on Mac, since it's possible things may behave differently there.

jcsteh:
Arrrgh! Did I mention I hate having to support mac?

Lo-lo78:
It's ok James, perfect!

👍

ScottChesworth:
Tagging extra Mac folk for testing. @GarthHum, @tbdalgaard. Please report back.

GarthHum:
Seems to work as expected on Mac, and doesn’t break copying and pasting parameter values from one instance to another. 
Very nice 
Appreciate the Mac support

jcsteh:
> doesn’t break copying and pasting parameter values from one instance to another.

Interestingly, this does break on Windows because we pass anything with the control key through. REAPER itself seems inconsistent in this respect. In text boxes in the Item Properties dialog, control+c, control+arrows and control+home do what you'd expect in a text box. In text boxes in the FX chain dialog, they don't; they trigger actions.

Should we just pass through anything involving function keys and anything involving alt+shift and remove the control key case?

ScottChesworth:
So is it Reaper or Windows itself that's scuppering us with Control? If it's the former, maybe that's something Cockos would address. There's a clear use case to fix it in the FX window, pasting when activating plug-ins.

jcsteh:
In the case of the Parameters dialog, it's OSARA: I made an explicit choice
to allow function keys, control+anything and alt+shift+anything in an
attempt to allow as many commands as possible. I figured that was
acceptable given the FX chain dialog, but I didn't think of the use case of
copying the raw parameter value. I can get rid of the control+anything
case, but that does obviously reduce the actions you can run. Maybe we
should allow control+shift+anything and control+alt+anything but not
control+anything?

In the case of the FX chain dialog, that's REAPER.

ptorpey:
Is there a way of knowing when one is in an edit field or not?  I would guess that one might like to have control+C,X,V active in edit fields but not need that when focus is on other elements.

jcsteh:
That's a good idea: override control+anything but only outside of text boxes. Also, I realised control+shift+anything shouldn't be overridden in text boxes either, as you might want to select a word (control+shift+rightArrow), etc.

Issue #476, opened 2021-02-16
mac key map conflicting with system command


KeithRempel:
In the midi editor there is an action assigned to command shift q. Quantize selected notes to grid or something close to that. this is in conflict with an OS default command to log out of current user account. My fix for this has been to switch the action over to control shift q. Probably worth switching across the board.

Another issue with the key map is that the slash key in the main actions dialogue has not been mapped to the action that moves edit cursor to play cursor. I think this is standard in windows. If needed I can make these tweaks to a fresh osara key map and pass an updated version along?

ScottChesworth:
Hey Keith, please make those changes to a freshly installed map and link to it here. I'll do the GitHub dance to get it incorporated and update documentation. Thanks!

KeithRempel:
https://www.dropbox.com/s/ok4jxnhne18856u/Mac%20modified%20by%20Keith.ReaperKeyMap?dl=0
Clean osara key map with the following modifications
1. Slash key mapped to View: Move edit cursor to play cursor
2. Added shortcut Control+Shift+Q to midi editor action “quantize note’s position to grid” I couldn’t figure out how to get rid of the command+shift+Q shortcut mapped here. Hoping to do so for the sake of tidiness.
3. mapped shift+TopRowDash (underscore) to the zoom out horizontally action in midi editor.

ScottChesworth:
Move to where both shortcuts are displayed, use Arrows to select Command+Shift+Q, then move to the Delete button, wallop.
Whoopsy, that zoom assignment is missing on Windows too. I'll fix that when I submit your new map to the reppo.

KeithRempel:
https://www.dropbox.com/s/a84jegy95wt1ouz/Mac%20Modified%20again%20by%20keith.ReaperKeyMap?dl=0 <https://www.dropbox.com/s/a84jegy95wt1ouz/Mac%20Modified%20again%20by%20keith.ReaperKeyMap?dl=0>

Same thing without the extra shortcut.

ScottChesworth:
Thanks. Submitting posthaste. :)

ScottChesworth:
Actually Keith, before I submit these, can I get you to do two more tweaks on Mac?
In the Main (Alt recording) category of the actions list, I'd like to get Space assigned to "Transport: play/stop" and R assigned to "Transport: record".
This is to prevent an edge case where if you enable the "Use alternate key map when recording" preference then start recording, there's no way to stop because nothing's mapped in that category.

[...]

PR #526, opened 2021-05-23
Key map corrections as requested in #476

ScottChesworth:
Mac tweaks made by @@KeithRempel


ScottChesworth:
Ugh. What did I do wrong this time?

jcsteh:
Oops. That wasn't you. Looks like I broke Mac builds.

ScottChesworth:
I saw it was something to do with the params dialog after commenting, just didn't wanna rub it in.

jcsteh:
I see this also includes Windows key map changes. Could you please provide a summary of what has changed here across both key maps so I can put that in the commit message? I realies some of this will be copy/pasting from the other issue, but I'm kinda unclear on what has changed here especially given this also touches Windows. Thanks.

ScottChesworth:
- Mac:
- Slash key mapped to "View: Move edit cursor to play cursor" in Main category to improve parity with Windows map.
- Moved "Quantize note's position to grid" in MIDI Editor category to Control+Shift+Q. Removed previous assignment because it clashed with a default OS shortcut to log out of current user account.
Both platforms:
- Assigned Shift+Dash to "View: Zoom out horizontally" in MIDI Editor category.
- Assigned Space to "Transport: play/stop" and R to "Transport: record" in Main (alt recording) category. This resolves an edge case where if users enabled the "Use alternate key map when recording" preference then started recording, there would've been no way to stop because nothing was mapped in that category.

PR #530, opened 2021-05-27
Mac installer: Fix detection of portable copies to handle app renames in REAPER 6.29.

### LINKED ISSUE #529: OSARA installer needs updater update on Mac OS to reflect changes to title of the Reaper application file
After Reaper 6.29 was updated it is no longer possible to install Reaper in a portable installation.
The reason is that the installer looks for Reaper64.app or Reaper32.app.
It should look for reaper.app instead.

@RDMurray you made some nice tweaks to this installer earlier. Would you mind looking into this? I guess we could add reaper.app to the list of filenames the script should look for, just to keep support for older versions (for now).

jcsteh:
Untested. @tbdalgaard, would you mind testing this when a build shows up here in a few minutes?

Fixes #529.

### DIFF
--- a/installer/mac/content/Install OSARA.command
+++ b/installer/mac/content/Install OSARA.command
@@ -7,6 +7,8 @@ var finder = Application("Finder");
 
 function isPortableReaper ( dir ) {
 	return (
+		finder.exists(Path(dir + "/REAPER.app")) ||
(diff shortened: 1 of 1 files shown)


tbdalgaard:
Cool. It works! Thanks so much for fixing this.

PR #533, opened 2021-05-27
Add a button in the Parameters dialog to show an envelope for the current parameter.

jcsteh:
I find creating FX envelopes really tedious. I can do it using control+alt+l (show envelope for last touched FX parameter), using the Param button in the FX chain dialog or using the envelopes dialog. Unless I'm missing something, I still find all of these somewhat inefficient/unintuitive. Last touched is probably the fastest, but that still means opening the OSARA FX Parameters dialog, wiggling the parameter, closing the dialog and finally pressing control+alt+l. This pull request adds a "Show envelope" button to the Parameters dialog, no wiggling required.

Currently, this only works for track and take FX. It won't work on track/item parameters, even though it shows up there. I might be able to get it working for some of those parameters, but I'm not sure yet.

Questions:

1. Would anyone other than me find this useful? Now I think about it, wiggling the parameter and pressing control+alt+l (you don't even have to close the dialog any more) isn't that hard. I don't know why it bothered me so much last time I had to do it. :)
2. Is it useful to get this working for other track/item parameters that support it?
3. The button will exist even if the envelope is already shown. Is that a big deal?


cn-peter:
### It's useful for people who need FX parameter automation.

Lo-lo78:
Yes, James, it's a great proposition!
Greetings!

ScottChesworth:
> Now I think about it, wiggling the parameter and pressing control+alt+l (you don't even have to close the dialog any more) isn't that hard.
>
This is your friendly rabbit hole alarm notification. :)
Up to you what you spend your time on man, but I'm not seeing any immediate UX benefit to this yet.
Folks who've said it's useful, please explain how/why with use case examples.

leonardder:
I don't think this change will benefit as long as it doesn't work for track parameter envelopes.

jcsteh:
This is no longer needed now that we have keyboard access to the Envelopes dialog.

PR #535, opened 2021-05-30
Changes which require REAPER 6.30

jcsteh:
This is for stuff that requires API changes in REAPER 6.30, which isn't yet released. I'll merge this once it is released.


ScottChesworth:
Ooh, solo defeat in reports is coming back then. Nice!

jcsteh:
I figured if I was already asking for one extra thing to be added to that
API function, I may as well ask for both. :)

ScottChesworth:
This is all working superbulously!
One question: how soon after Reaper 6.30 drops are you thinking of merging this? I took it for a quick test run in 6.29 to gauge the likely UX for anyone who misses the memo that 6.30 is required, and inserting a track caused both NVDA and Reaper to crash lol.
I'm not expecting you to solve that, just flagging it because we probably need to tread carefully. I don't think there's been an OSARA snapshot where the very latest Reaper is mandatory before has there. Of course, I'll make noise in all the community channels, but there's bound to be some places I won't reach, especially as I only speak English.
If you're keen to merge soon, is it worth forcing something like "requires Reaper 6.30" early in the file name for a few snapshots, d'ya think?

jcsteh:
Ooo. I wasn't expecting a crash. I was expecting that anything in the Peak
Watcher except peak dB wouldn't work, that armed would always be reported
first and that solo defeat wouldn't be reported. While that's not great, I
figured it would be acceptable enough; anyone who wanted those features
would probably realise they needed to upgrade, given how recent LUFS
support was added. Do you think that's acceptable?

I think I can probably fix the crash. I have a hunch I know what's
happening there.

jcsteh:
> If you're keen to merge soon, is it worth forcing something like "requires Reaper 6.30" early in the file name for a few snapshots, d'ya think?

That would be non-trivial; there's a bunch of stuff I'd have to change in ugly ways to make that work and I'm reluctant to do that for what would really be a temporary change.

jcsteh:
The changes I just pushed (build coming up soon) should fix the crash, though I haven't tested in 6.29.

jcsteh:
Should I add Peak Watcher support for LRA? It's easy enough to do. However, it's a bit weird because the way you interpret those numbers is a bit different: I'm not sure how much it makes sense to alert if loudness range goes above a certain level. Certainly, the default of 0 doesn't make sense for LRA. That said, people can easily disable notification or change the level.

ScottChesworth:
> The changes I just pushed (build coming up soon) should fix the crash, though I haven't tested in 6.29.
> 
Fixed. Thanks! Everything performs as described above, the only unexpected thing I noticed is that hitting Control+Win+F6 always reports "not defeating solo". If it's easy to have it just not report, that would probably be more intuitive. Don't feel obligated to put any sweat into that though, it makes more sense for me to put effort into encouraging people to upgrade. There's a lot for them to gain. Between the improved Peak Watcher functionality and also all the action/render-based loudness options Cockos are adding, this update is a massive leap forward.

ScottChesworth:
> Should I add Peak Watcher support for LRA?
> 
I've not encountered/used LRA measurement much here yet, but from a quick read about it, it seems like it might be a good fit for podcasters and spoken word broadcast where people don't tend to agonize so hard over dynamic range. I reckon the scale going from zero upward might be easier to grasp than other LUFS measurements for less techie types.
All that to say, yes please. :)
> it's a bit weird because the way you interpret those numbers is a bit different: I'm not sure how much it makes sense to alert if loudness range goes above a certain level. Certainly, the default of 0 doesn't make sense for LRA. That said, people can easily disable notification or change the level.
> 
Hmmm, tricky. Is there any mileage in disabling notifications by default when LRA is selected? It's a bit lumpy UX-wise I know, but at least then people can get the info they're seeking on demand without getting flooded. Alternatively, I can write up a cheat sheet with recommended settings and plug it in the support channels whenever someone comes a cropper, maybe that'll be enough.

jcsteh:
> the only unexpected thing I noticed is that hitting Control+Win+F6 always reports "not defeating solo". If it's easy to have it just not report, that would probably be more intuitive.

Done.

> Is there any mileage in disabling notifications by default when LRA is selected?

Technically, it could be done, but it has nasty UX side effects that would be equally unintuitive. For example, say the user selects LRA, then realises they didn't want that and selects LUFS. Do we restore the notification settings to defaults? Worse, what if they change their notification settings, choose LRA, realise they didn't want LRA and then choose LUFS. We reset their notification settings to defaults... but now we just clobbered their settings.

ScottChesworth:
> > Is there any mileage in disabling notifications by default when LRA is selected?
> 
> Technically, it could be done, but it has nasty UX side effects that would be equally unintuitive.
> 
Fair points. Doesn't seem like we can anticipate enough to make changes by default. I'll tinker with LRA and get a cheat sheet written for the Dropbox/wiki, maybe record a demo if I can find time.

Lo-lo78:
I just tried it, I think I should have Reaper 6.30? On 6.29 many things 
are not working in the peak watcher. Greetings!

Lo-lo78:
I read the mega news!

FANTASTIC!!!


v6.29+dev0608 - June 8 2021
  * Includes feature branch: render normalization
  * Includes feature branch: MIDI editor note reordering
  * Includes feature branch: media item lanes
  * Includes feature branch: EEL2 x86_64/SSE
  * Includes feature branch: VST3 bridging
  + Master track: support LUFS-M metering
  + ReaScript: GetSetProjectInfoString(project, "RENDER_METADATA", "", false) will return a list of all defined project metadata identifiers [t=254407]
  + ReaScript: support running a caller-specified action before returning render statistics (see API help for details)
  + Render: add actions to calculate loudness of master mix via dry run render
  # API: make GetItemFromPoint consistent with internal hit testing [p=2451995]
  # JSFX loudness meter: expose current LUFS-M/I/S and LRA values as parameters for script accessibility/etc
  # JSFX loudness meter: expose true peak and RMS readings as parameters
  # JSFX: don't display partially-calculated LUFS
  # JSFX: fix initial layout of loudness meter when not processing FX on stop
  # Media items: increase vertical offset of overlapping items, don't draw overlapping items opaque by default
  # ReaScript: slightly change the output of GetSetProjectInfo_String("RENDER_STATS") for consistency and futureproofing
  # VST3: require ModuleEntry on linux

v6.29+dev0607 - June 7 2021
  + Media items: fix hit testing to prefer opaque unselected items over transparent selected items
  + ReaScript: add GetSetProjectInfo_String("RENDER_STATS"), to retrieve statistics for the most recently rendered files
  # Media item lanes: remove preference to auto-crossfade media items in separate lanes
  # Options: restore option to display overlapping items in lanes, as "offset overlapping media items vertically"
  # Render: don't display "show in explorer" or "launch file" buttons after dry run render

v6.29+dev0605 - June 5 2021
  + Render: add info button to render statistics display
  + Render: display LRA (loudness range) after render
  # Actions: action to calculate loudness of selected tracks will calculate master mix loudness if the master track is the only track selected
  # JSFX: fix auto-reset of loudness meter when not processing FX when stopped
  # JSFX: fix hidden sliders altering FX parameter count [p=2451177]
  # JSFX: optimize loudness meter CPU use when not displaying true peak
  # MIDI editor: improve auto-selection of custom note view when gluing all items on track [p=2450964]
  # Render: add separate button for dry run render
  # Render: fix 1-pixel peaks drawing error

v6.29+dev0603 - June 3 2021
  + Actions: add action to calculate loudness of selected tracks within time selection via dry run render
  + ReaScript: support getting/setting render normalization mode and target
  + Render: add preference to calculate true peak on render when not normalizing to it
  + VST3: add IPlugInterfaceSupport, for plugins to request compatibility features
  + VST3: add support for IMidiLearn interface
  + VST3: avoid threading conflicts when processing expression and keyswitch information
  + VST3: internal threading changes to MIDI controller and program change handling
  # Batch converter: fix button positioning when window is resized
  # Convert: fix normalization for source files that do not have an embedded source start time [p=2450739]
  # MIDI editor: fix keyboard drawing glitches around top of view [p=2450683]
  # MIDI note ordering: fix ordering issue after undo [p=2449996]
  # MIDI note ordering: give more visual feedback about selected note [p=2450000]
  # Rendering: fix denormal issues with silence [p=2450456]

v6.29+dev0602 - June 2 2021
  + Actions: add action to calculate loudness of selected tracks via dry run render
  + Batch converter: support normalizing when converting
  + Media explorer: fix formatting of absolute paths from tree browser on linux/macOS
  + Project files: allow filenames to contain quotes on supported platforms
  # JSFX: add option to loudness meter to reset data on playback start
  # Media items: fix tape-mode recording option potentially being ignored [p=2449852]
  # Render: support normalizing to digital peak or true peak

v6.29+dev0531 - May 31 2021
  + JSFX: calculate "true peak" in loudness meter
  + JSFX: display true peak clip count in loudness meter

v6.29+dev0530 - May 30 2021
  + JSFX: add help for sliderchange() function
  + JSFX: add slider_show() function, to hide or display sliders
  + macOS: improve appearance of fade menu items when using dark mode [t=254007]
  # JSFX: add alert settings to loudness meter
  # JSFX: add loudness range calculation to loudness meter
  # JSFX: fix reversed LUFS-M current and maximum value display
  # JSFX: support customizable colors in loudness meter

v6.29+dev0529 - May 29 2021
  + ReaScript: add GetSetMediaTrackInfo() B_SOLO_DEFEAT and B_AUTO_RECARM
  + Render: ignore preference to auto-close render dialog if normalizing or dummy rendering
  + Render: support optional LUFS handling of LFE and surround channels when normalizing
  # JSFX: add LUFS-M histogram support
  # JSFX: add short-term LUFS to loudness meter
  # JSFX: fix loudness peak meter channel order
  # JSFX: improve loudness meter layout at small sizes, hide RMS by default
  # JSFX: loudness meter is LUFS-multichannel-aware
  # Render: fix dummy render of multichannel files

v6.29+dev0528 - May 28 2021
  + Actions: add action to calculate loudness of selected items via dry-run render
  + JSFX: improve LUFS loudness calculation
  + JSFX: improve LUFS meter display when embedded
  + ReaTune: enable TCP/MCP embed view for tuner
  # JSFX: add separate momentary RMS measurement to loudness meter
  # JSFX: improve LUFS meter scaling
  # Render: fix normalizing selected media items or multiple time regions per track
  # Render: fix saving normalization settings to render presets
  # Render: improve LUFS loudness calculation
  # Render: potentially improve precision of normalization slightly

v6.29+dev0527 - May 27 2021
  + JSFX: add peak/RMS/LUFS loudness meter
  # Render: fix LUFS calculation
  # Render: fix embedding bwf position when normalizing
  # Render: fix peaks drawing with multichannel renders
  # Render: fix up peaks drawing

v6.29+dev0526 - May 26 2021
  + Render: display LUFS, support normalizing to target LUFS
  + Render: display peak, RMS, overs even if not normalizing
  + Render: display statistics for all rendered files
  + Render: support dry run rendering (everything is processed as normal, but no output files are written)
  # MIDI editor: custom note ordering is stored per-track rather than per-item
  # MIDI editor: fix paint linear notes with reordered notes
  # Render: use fixed 400ms window for RMS normalization

v6.29+dev0524 - May 24 2021
  + Actions: cycle folder collapsed state will not modify state of selected non-folder tracks
  + Render: allow rendering to primary/secondary of the same type (e.g. wav 16-bit and 24-bit, $bitdepth recommended in wildcard)
  + Render: apply dither/noise shaping to secondary render format if enabled and format supports
  + Render: disable audio normalizing for video renders
  + Render: display peak, rms, and overs when normalizing
  + Render: improve render speed when normalizing
  + Render: show first actual file used when rendering to multiple files that are autodisambiguated
  + Super8: allow controlling loop length via automatable parameter
  + Super8: fix off-by-one-sample error in first-pass recording
  + Super8: per-channel length-division setting
  # Render: display render meters post-normalization for greater-than-mono tracks
  # Render: if normalization would require adding more than 60dB, don't normalize
  # super8: fix sync mode reset bug added with hidden parameter

v6.29+dev0521 - May 21 2021
  + Render: draw peaks on first pass of 2nd pass render
  + Render: optionally mark overs (samples over +0dbFS, potential digital clipping) in render peaks display
  # Automation items: entirely un-pool automation items when editing with preference enabled [p=2446337]
  # Render: change to internal storage for render normalization (project normalization settings saved in +dev0520 may load differently)
  # Render: display render meters post-normalization
  # VST3: do not require win32 InitDLL nor linux ModuleEntry
  # Windows: revert non-functional XP rx2 support bleh

v6.30+dev0520 - May 20 2021
  + Automation items: add preference to pool/unpool automation items when split due to media item edit or razor edit
  + JSFX: export_buffer_to_project() can take a planar_pitch parameter to allow exporting planar multichannel content
  + MIDI: fix duplicate CC values being sent at the start of linear/bezier curves
  + Razor edit: fix context menu inconsistency when razor edit mouse modifiers are remapped [p=2445668]
  + Render: add options to normalize render to target peak or RMS
  + Super8: allow controlling synchronization via automatable parameter
  + Super8: when adding loops to project, export linked channels as stereo file
  + Tracks: prevent tiny (<0.1%) volume change on initial track fader click [t=191454]
  + VST3: require bundle/module/dll entry-points to match Cubase/VST3 spec
  + Windows: improve installer behavior when choosing Portable checkbox after editing installation path [t=253741]
  + Windows: support rx2 files on 32-bit Windows XP [t=253306]
  # VST3 bridging: fix Windows non-embed sizing
  # VST3: fix resize notifications

v6.28+dev0430 - April 30 2021
  # MIDI editor: autoscroll when reordering note rows

v6.28+dev0429 - April 29 2021
  + MIDI editor: add custom note order/visiblity mode (ctrl+drag note key in named notes view)
  + MIDI editor: import/export of note names includes custom-view ordering information if set

v6.28+dev0423 - April 23 2021
  + Windows: ignore alt-key mouseup when using alt+mousewheel [t=251477]

v6.27+dev0417 - April 17 2021
  # Item group containers: remove feature for now

v6.27+dev0416a - April 16 2021
  # Item group containers: fix creating more than one group container in a project
  # Item group containers: fix selecting containers

v6.27+dev0416 - April 16 2021
  + Media items: add support for grouping media items into container items

v6.27+dev0412 - April 12 2021
  + Media item lanes: action to implode items across tracks to items on a single track puts items in separate lanes, if the first track is a fixed-lane track
  + Tempo markers: support tempo basis other than 1/4 note (affects displayed BPM number only, not metronome or tempo envelope display) [t=250880]
  # JSFX: SSE bitwise op behavior fixes [p=2426806]
  # Media item lanes: fix auto-coloring when adding takes in fixed lane mode [p=2428272]

v6.26+dev0401 - April 1 2021
  + Media item lanes: add preference for minimum track height to display lane-add area
  # JSFX: fix more SSE-mode bugs [p=2426302]

v6.26+dev0331 - March 31 2021
  # JSFX: fix flushing of inf/nan to zero [t=251603]
  # Media item lanes: fix dragging items into lane-add area [p=2425679]

v6.26+dev0330 - March 30 2021
  + JSFX: optimized x86_64 (now supporting SSE), aarch64, armv7l code generation
  # Media item lanes: don't autoarrange media items when switching a track to fixed lane view
  # Media item lanes: reset any existing item lane data (for media items that were previously on a fixed lane track) when adding lanes on new recording

v6.25+dev0323 - March 23 2021
  + macOS: improve JSFX performance when running under Rosetta2
  # Media item lanes: fix various issues when loading a project with overlapping items saved in an older version of REAPER

v6.25+dev0322 - March 22 2021
  + Media items: change default item mix behavior to "items always mix", existing projects and user default project settings are unaffected

v6.25+dev0319 - March 19 2021
  + Glue: adjust glued items for item lane position
  + VST3 bridging support: not feature-complete but usable. especially useful for M1 macs
  # Media items: fix calculating item label height with free item positioning

v6.25+dev0318 - March 18 2021
  # Media item lanes: don't create an empty second lane by default
  # Media item lanes: improve pencil-draw behavior
  # Media item lanes: maintain relative lane relationships when editing multiple items at once
  # Media item lanes: when moving multiple items between tracks, increase the destination track lane count if necessary

v6.25+dev0317 - March 17 2021
  + Color theme: add theme color for track lane add area
  + Media item lanes: auto-arrange overlapping media items when switching a track to fixed lane view
  + Media item lanes: automatically set fixed track lanes if needed when recording new items and preference set to create new items for overlapping recording
  + Media item lanes: for fixed lane tracks, apply item mix behavior (enclosed items replace enclosing items, or items always replace earlier items) only for items in the same lane
  + Media item lanes: increase track item lanes by dragging items into a band at the bottom of the track
  + Media item lanes: remove option to automatically show overlapping items in lanes
  + Media item lanes: track items lanes decrease automatically
  + Options: add option to auto-crossfade media items in different lanes
  + Recording: respect option to always add takes to new recording when looping
  # Media item lanes: add preference to enable fixed item lanes for new tracks

v6.25+dev0315 - March 15 2021
  + Media items: add preference to collapse item lanes below a minimum track height
  # Media items: improve behavior when dragging items between tracks with different numbers of lanes
  # Media items: preference for minimum peaks height applies per-item, not per-channel

v6.25+dev0313 - March 13 2021
  + Media items: add preference for minimum media item height to display peaks

v6.25+dev0310 - March 10 2021
  + Media item lanes: when displaying overlapping items in lanes, increase to a fixed lane count if needed when adding recorded media
  + Menus/Actions: refer to "takes" instead of "take lanes" everywhere, to avoid ambiguity with media item lanes
  # Media item lanes: automatically reposition items when changing number of lanes
  # Media items lanes: fix height threshold for displaying labels over items with also displaying overlapping items in lanes
  # Media items: increase height threshold for drawing media item peaks

v6.25+dev0309 - March 9 2021
  + Media item lanes: don't draw peaks at tiny media item heights
  # Media item lanes: don't collapse lanes until they are tiny
  # Media item lanes: improve behavior when changing number of lanes

v6.25+dev0308 - March 8 2021
  # Media item lanes: ensure that all items are displayed when reducing number of item lanes
  # Media item lanes: more consistent lane collapsing

v6.25+dev0307 - March 7 2021
  + Tracks: Add actions to increase/decrease number of track item lanes
  # Item lanes: actions to move items up/down one lane work for tracks with a fixed number of item lanes

v6.24+dev0304a - March 4 2021
  # Item lanes: prevent items overlapping vertically when displayed in lanes

v6.24+dev0304 - March 4 2021
  # Free item positioning: automatically reposition items when switching to FIPM
  # Item lanes: improve automatic item arrangement compactness when displaying overlapping items in lanes
  # Item lanes: improve behavior when drawing new media items
  # Item lanes: improve behavior when moving items between tracks with different numbers of item lanes
  # Razor edit: maintain media item lane relationships when changing tracks

v6.24+dev0303 - March 3 2021
  + Item editing: improve interaction between free item positioning edits and mouse modifiers
  + Item lanes: when recording in free item positioning mode, only adjust existing item y-positions if necessary
  + Item lanes: when recording into a track with a fixed number of lanes, auto-increase the number of lanes if necessary
  + Media items: auto-arrange items in lanes using the same logic as auto-arranging free item positioning
  + Media items: draw peaks affected by take mute envelope as faint, rather than zeroed out
  + Track manager: display track channel count
  + Track manager: display track item lanes
  + Tracks: support setting a fixed number of item lanes per track
  # Media item channel pin mapper: removed

v6.23+test0224 - February 24 2021
  # Render: fix mono rendering

v6.23+dev0214 - February 14 2021
  # Media channel pins: add undo state when changing via alt+click of item properties button

v6.23+dev0212 - February 12 2021
  + Media items: alt+click on media item information button opens channel mapper
  # Media items: display only one channel in media channel mapper for mono or channel mode mono media

v6.23+dev0211 - February 11 2021
  + Actions: add action to open channel mapper for selected items
  # Media items: inprove handling of various combinations of channel mode and channel mapping

v6.23+dev0210 - February 10 2021
  # Media items: media channel mapper takes affect after media channel mode setting, does not affect display
  # Media items: support mapping mono items to channels > 2

v6.23+dev0209a - February 9 2021
  # Media items: only modify peaks drawing if custom channel mapping does not have higher-numbered channel mappings
  # Media items: support mapping media item channels to track channels higher than the media item channel count

v6.23+dev0209 - February 9 2021
  + Media items: support custom channel mapping in media item properties dialog

v6.19+dev1228 - December 28 2020

v6.19+dev1222 - December 22 2020
  # macOS: revert space keyboard passthrough workaround code

v6.19+dev1219 - December 19 2020
  + macOS: allow space keyboard passthrough from Waves StudioRack-hosted plug-ins

v6.18+dev1209 - December 9 2020
  + FX: add option to run anticipative FX synchronously during playback start

v6.12+dev0625 - June 25 2020
  + System: fix potential crashes due to high socket use on macOS and Linux [t=238309]
  + Web interface: reduce latency of web server
  # MIDI: revert option to prevent MIDI notes from retriggering when splitting media items [p=2309790]

v6.12+dev0622 - June 22 2020
  + MIDI: add preference to suppress note retriggering when splitting media items

v6.12+dev0621 - June 21 2020
  + Actions: add action to suppress midi note retrigger at start/end of specific media items
  + MIDI: automatically suppress midi note retrigger when splitting MIDI media items

v6.12+dev0617 - June 17 2020
  # Area selection: removed

v6.11+dev0604a - June 4 2020
  # Actions: fix Remove items/tracks/envelope points (depending on focus) -- from previous build

v6.11+dev0527a - May 27 2020
  # Actions: action to paste at mouse cursor respects snap, does not move edit cursor if preference disabled
  # Area selection: fix copying part of an automation item
  # Area selection: fix crash when rapidly switching between right and left mouse drag
  # Area selection: only clear area selection on paste after cut
  # Area selection: suppress edits until the mouse has moved a few pixels
  # Media item edits: fix stuck vertical moves in +dev0527

v6.11+dev0527 - May 27 2020
  + Actions: add action to paste tracks/items at mouse position
  # Area selection: always add edge points when moving envelope areas [p=2294871]
  # Area selection: fix copy/paste of automation items
  # Area selection: fix vertical edge editing with item-edits-affect-envelopes enabled
  # Area selection: unless focus is on track panels, cut/copy/delete actions always target area selection if there is one

v6.11+dev0526a - May 26 2020
  # Area selection: fix copy/paste of envelope areas [p=2294757]
  # Area selection: prevent potential hang when resizing area selection vertically [p=2294782]
  # Area selection: snap respects preference to snap to media item start/end [p=2294791]

v6.11+dev0526 - May 26 2020
  # Area selection: action to split/select area selections affects automation items as well as media items
  # Area selection: add area selection click mouse modifiers to split or copy items within the area
  # Area selection: better handling of editing area selections over existing automation items [p=2294221]
  # Area selection: handle duplication of multiple non-time-contiguous areas possibly more predictably
  # Area selection: minor improvements to marquee preview display

v6.11+dev0525a - May 25 2020
  + Area selection: add action to duplicate area selection
  # Area Selection: prevent creating selections that start at negative times
  # Area selection: improve paste behavior when the area selection contains empty space
  # Area selection: improve some behaviors when moving an area selection on a single envelope
  # Area selection: only support top/bottom edge edit at the top and bottom of contiguous areas
  # Area selection: snap to visible grid if configured

v6.11+dev0525 - May 25 2020
  + Area selection: add area selection and area selection edge mouse modifiers
  + Area selection: drag areas to move or copy, drag edges to adjust selection or stretch contents
  + Area selection: support creating area selections in arrange view right-drag mouse modifiers
  + Area selection: support cut/copy/paste
  + Area selection: support selecting areas of media items and envelopes

v6.09+dev0503 - May 3 2020
  + MIDI: add action to send explicit note-off messages for every channel/pitch combination to every MIDI output and plugin

v6.09+dev0428 - April 28 2020
  + Take markers: avoid potential crash when media item source file is not available

v6.08+dev0421 - April 21 2020
  # ARA: minor optimization when using ARA as per-take FX

v6.05+dev0308 - March 8 2020
  # Take markers: fix drawing when media item is partially scrolled off the top of the arrange view

v6.02+dev1221 - December 21 2019
  + MacOS: optionally display last user action in REAPER title bar
  # Toolbars: fix toggle state for SWS cycle actions [p=2220575]

v6.02+dev1219 - December 19 2019
  + Actions: show custom action toggle state as enabled/disabled if all component actions that report a toggle state are enabled/disabled [t=229048]
  + Toolbars: refresh toolbar after running any action via toolbar [t=229048]

v6.01+dev1213 - December 13 2019
  + MIDI editor: add option to prevent mouse edits of single CC events from moving past other CC events

v6.29rc4 - May 11 2021
  + Razor edit: fix positioning of initial transition envelope point [p=2450795]

v6.28+dev0505 - May 5 2021
  # VST3 bridge: fix Windows UI embed hang for RealPhones [p=2438320]

v6.28+dev0502 - May 2 2021
  # EEL/sse: fix compiletime issue with namespaced functions on Windows and probably elsewhere

v6.27+dev0420 - April 20 2021
  + Misc: update zlib to 1.12.11 and libpng to 1.6.37

v6.29 - May 18 2021
  + Actions: rename "Envelope: Clear envelope" to "Envelope: Clear or remove envelope"
  + API: add hwnd_info hook for improved plug-in global hotkey integration
  + API: fix typo in documentation of IReaperUIEmbedInterface [p=2438620]
  + API: GetSetProjectInfo_String() supports TRACK_GROUP_NAME:x
  + API: run hookpostcommand2 hooks for reascripts and custom actions
  + Automation items: action to create new automation item uses razor edit areas if they exist
  + Envelopes: add "remove envelope" action to envelope control panel context menu
  + Envelopes: fix possible redraw of media lane envelopes when opening project [p=2444590]
  + FX: fix monitoring window persistence when using new undo point option [t=252896]
  + FX: improve embedded FX update frequency when mouse-editing
  + FX: improve docked mixer visibility restoration when adding FX via docked FX browser [t=253330]
  + FX: fix visual display of TCP/MCP parameter knob when parameter min/max is reversed
  + JSFX: allow export_buffer_to_project() to use 64 channels
  + JSFX: effects may now request samplerate for opened media using file_riff(fh,nch='rqsr',srate=desired_samplerate);
  + macOS: rename ARM build to REAPER-ARM.app, Intel 64-bit to REAPER.app, update volume labels of disk images
  + Media explorer: add option to preview through track named "Media Explorer Preview"
  + Media explorer: fix re-sorting randomized file list after running action to set file as favorite [p=2437642]
  + Media explorer: fix searching after renaming file
  + Media explorer: refresh display after resetting column order to default
  + Media explorer: support adding additional user-specifiable metadata display columns
  + Media items: when mouse-moving selected area of grouped items, group the remaining right-hand side of items separately from the left-hand side, for consistency with cut/paste behavior
  + Notation: fix key snapping on first beat of local key signature change [t=221385]
  + Notation: improve edit cursor position rounding when using snap to samplerate
  + Notation: assign current voice when inserting notes via "insert note at nearest A/B/C/etc" actions [t=216159]
  + OSC: support touch status for up to 2048 tracks [t=253320]
  + Project bay: show bypass/offline status for FX that are unavailable
  + Project metadata: disallow spaces in IXML element identifiers
  + Razor edit: add razor edit envelope editing behaviors
  + Razor edit: add mouse modifiers to expand or compress envelope range, relative to average value or toward top/bottom
  + Razor edit: ensure that pasted media items remain grouped and pasted automation items remain pooled across tracks
  + Razor edit: fix media item display after pasting razor edit area and zooming [t=252759]
  + Razor edit: skip hidden envelope lanes when marquee-creating razor edit areas with "move envelope points with razor edits" disabled
  + Razor edit: when editing or cutting, group the remaining right-hand side of items separately from the left-hand side [p=2439372]
  + Razor edit: when mouse-copying items, create new groups if necessary
  + ReaPlugs: display more significant digits of ReaDelay/ReaPitch pan position [t=253521]
  + ReaScript: show file/line number for Lua calls to ReaScriptError()
  + ReaScript: improve Lua ReaScriptError() behavior when called from extension-added functions, allow ! prefix to stop script execution
  + ReaScript: support "NUMCHANNELS", "NUMSPEAKERS", "RESETCHANNELS" for ReaSurroundPan via Track/TakeFX_SetNamedConfigParm
  + ReaScript: ReaSurroundPan FX_Get/SetNamedConfigParm("NUMSPEAKERS") will interpret a value like "6S" as 6-channel surround, "8I" as 8-channel ITU, "12A" as 12-channel Auro, etc
  + ReaSurroundPan: add new 2D/3D surround panner plug-in
  + ReaSurroundPan: supports common speaker spatial arrangements and user-customizable arrangements up to 64 inputs/outputs
  + ReaSurroundPan: add actions to display ReaSurroundPan embedded in TCP or MCP
  + ReaSurroundPan: support axis linking (forcing edits to a particular shape on the z-axis)
  + ReaSurroundPan: support editing anchor channel via controller, but does not record automation
  + ReaSurroundPan: support per-channel delay
  + Ruler: add option to display region name when region start is offscreen
  + Ruler: add option to hide region number if region is named
  + Spectral edits: fix issue with playback starting after item loop [t=252747]
  + Spectral edits: improve performance with large numbers of spectral edits
  + Super8: allow drag-drop of samples to channels
  + Super8: improve layouts when UI is resized very wide or very tall
  + Super8: improve MIDI-control behavior when same note/CC is assigned to a channel's record and play controls
  + Theme: add razor_env, razor_env_l, razor_env_r cursors
  + Video: improve default ordering of 2x2 matrix preset [t=252286]
  + VST3: actually support third-party embeddable UIs via IReaperUIEmbedInterface interface [t=252907]

v6.28 - April 22 2021
  + Accessibility: improve behavior of spacebar in actions list and managers with spacebar-navigation option set
  + API: ListView_GetItemText() supports owner-data listviews on non-win32 [t=241880]
  + FX: allow JS/Channel Mapper-Downmixer to be renamed [p=2431465]
  + FX: fix mapping track channels above 32 using the FX pin connector dialog [t=252332]
  + FX: avoid zombie bridge processes in various corner cases
  + macOS: improve behavior when scanning VST plug-ins that present modal UI
  + Marker/region manager: fix undesired periodic marker deselection
  + Marker/region manager: improve selection behavior when removing markers/regions
  + Marker/region manager: move edit cursor when seeking during playback
  + Media explorer: fix tempo sync when inserting media on new track [p=2433091]
  + Media explorer: improve readability of embedded cues
  + MIDI: enforce some reasonable maximum MIDI message and per-block list sizes (128MB and 256MB initially)
  + MIDI: fix unlikely but possible incorrect handling of PDC+MIDI
  + Razor edit: properly handle moving and pasting areas that contain MIDI items when project timebase is set to "time" [t=252436]
  + Razor edit: restrict selected area to 0.001 rather than 0.01 seconds
  + Render: fix display of render vu meters with panned mono media [t=252026]
  + Track manager: fix keyboard navigation quirks on macOS with mirrored selection enabled

v6.27 - April 11 2021
  + Dither: fix half-bit DC offset [p=2425940]
  + FX: add option to not add undo points when closing FX windows [t=250733]
  + FX: improve pin connector window
  + FX: fix persistence of take FX processing channel count when processing only 2 channels on a track with more than 2 channels
  + JSFX: add Channel Mapper-Downmixer plugin
  + JSFX: add option in pin connector dialog I/O menu to pass through or zero out unused output channels
  + JSFX: support drawing vertical text via gfx_setfont flag 'z' (does not work with system/bitmapped font)
  + JSFX: allow Webdings/Wingdings for gfx_setfont() use on Windows [twtr=1376339905408815106]
  + macOS: add prefs/general/advanced option for window fullscreen button to activate fullscreen rather than OS fullscreen
  + macOS: add prefs/general/advanced options for rounded buttons and big sur listview margins
  + macOS: disable macOS 10.12+ Window tabbing menu items
  + Media item channel mapper: control/command-click a mapping to set exclusively for that channel
  + Media items: add action to set channel mapping for selected media items
  + MIDI: output multiple note-offs for the same channel/pitch if there are multiple sequential note-ons
  + Project time offset: preserve maximum resolution when setting project time offset to edit cursor position [p=2428360]
  + ReaScript: support drawing vertical text via gfx_setfont flag 'z' (does not work with system/bitmapped font)
  + ReaScript: allow Webdings/Wingdings for gfx_setfont() use on Windows [twtr=1376339905408815106]
  + ReaScript: correct documentation for Set/GetMasterTrackVisibility
  + ReaTune: fix potential crash with certain parameter settings and UI open [t=251553]
  + Render: resolve $samplerate wildcard correctly when mixing at one sample rate but rendering to another [t=249713]
  + TCP FX List: fix multi-column scroll extent issue [t=245850]
  + Timeline: truncate seconds and samples display rather than rounding, for consistency with H:M:S and H:M:S:F display
  + Timeline: when displaying minutes:seconds and a project start offset exists, center ruler major tick marks on project time zero rather than the start of the timeline
  + Track manager: display track channel count
  + VST: add I/O menu item in FX pin connector dialog to increase host channel count to match plugin
  + VST: add option in pin connector dialog I/O menu to pass through or zero out unused output channels
  + VST: cache VST3 parameter/bus/latency information for improved performance with yabridge and plug-ins
  + VST: improve behavior when VST3 plug-ins call restartComponent(kReloadComponent)
  + VST: more flexible support for passing MIDI program change messages to VST3 [t=251104]

v6.26 - March 29 2021
  + Actions: action to insert or extend MIDI item does not skip existing looped MIDI items [t=250900]
  + Arrange: prevent possible overdraw with selected track cursor bracket and blinking cursor [t=250398]
  + Dither: use prvhash for rng/tpdf generation (courtesy Aleksey Vaneev)
  + Edit cursor: display slightly subtler bracket on last selected track
  + Envelopes: fix editing automation item edge point using envelope point editor dialog [twtr=1367910774371401731]
  + Glue: actions glue within razor edit areas if they exist
  + Glue: default glue action splits items and glues only within time selection, if all selected items intersect the time selection
  + Glue: former default glue action is renamed "expanding to time selection if any", custom actions and scripts that use it are unaffected
  + JSFX: add output polarity setting for Liteon/nonlinear [t=251346]
  + JSFX: autoexpand soft knee fix from sstillwell
  + JSFX: update 1175 compressor and various others to fix overcompression, add deprecated blown capacitor mode to preserve old behavior
  + JSFX: add improved SRC gain correction mode for guitar/amp-model and guitar/amp-model-dual
  + macOS: add NSMicrophoneUsageDescription to improve audio input support on Big Sur
  + macOS: main window fullscreen button activates REAPER-native fullscreen rather than OS fullscreen
  + macOS: workaround Big Sur MIDISend() bug where 250 or more events at the same timestamp crashes
  + Media items: adjust snap offset when copying selected part of media item [p=2417743]
  + Media items: preserve relative snap offset when copying media items [p=2410077]
  + Metadata: support displaying multiple iXML or xmp metadata keys in media source properties dialog [t=250356]
  + Metadata: support embedding images when rendering FLAC files
  + MIDI: suppress unnecessary prompt on importing MIDI files [t=251349]
  + MIDI: fix issues when launching rename note action via enter key
  + MIDI: improve MIDI editor focus behavior when closing various text editing fields
  + Mute automation: fix potential hang with empty automation item [t=250949]
  + Razor edit: action to crossfade items at time selection acts on razor edits if they exist
  + Razor edit: add action to move nearest razor edit edge to edit cursor
  + Razor edit: add mouse modifiers to remove area from razor edit selection
  + Razor edit: auto-scroll vertically when editing razor edit top/bottom edge
  + Razor edit: fix stretching automation items across tempo changes when razor edit area does not match automation item length [p=2418894]
  + Razor edit: mouse modifer to split items at razor edit edges selects media items and removes the razor edit area
  + Razor edit: respect preference to update video playback position when editing audio items [p=2417843]
  + Razor edit: update display after running action to enclose media items [t=250589]
  + API: add ResolveRenderPattern, to resolve a wildcard pattern into a list of target filenames
  + ReaScript: update Track_GetPeakInfo()/Track_GetPeakHoldDB() to support master track RMS, update documentation
  + ReaSurround: update track channels only when increasing plugin speaker count, not when changing plugin input channel count or decreasing speaker count
  + ReaVerb: add new SRC-normalization mode which should be more consistent for some impulses and samplerate combinations [t=249454]
  + Region/Marker Manager: support exporting take markers, but not importing [t=250605]
  + Render dialog: add support for rendering razor edit areas
  + Video processors: do not warn when loading empty video processor preset [p=2411241]
  + Video processors: fix preset name preservation behavior
  + Video processors: make preset list consistent with other FX types (user presets first)
  + Video processors: support preset navigation API [t=249961]
  + VST: improve bridged plug-in floating window behavior on macOS
  + VST: improve floating window behavior of bridged VST plug-ins

v6.25 - March 5 2021
  + Appearance: add option to hide edit cursor highlight on last selected track
  + Appearance: reduce locked media item darkness slightly (6.24 change)
  + JSFX: fix stack functions that caused Reaticulate to hang (6.24 regression) [t=250317]
  + JSFX: prevent more than 128MB of MIDI per block from being sent
  + ReaScript: support extra parameters in classic extension_api() calls [p=2414474]
  + WAV: fix potential crash when reading wav files written by other applications [t=250134]
  + Windows: fix 32-bit ReaRoute [p=2413463]
  + Windows: prevent ReaRoute crash when using unreasonably-large block sizes

v6.24 - March  2 2021
  + Accessibility: improve labels for FX preset menu and edit comment buttons
  + Accessibility: add textual fade descriptions to menus via accessibility APIs
  + AIFF: support in24/42ni AIFC encoding
  + Arrange view: display bracket over last selected track, to indicate where pasted content will go
  + Envelopes: add "remove envelope" menu item to envelope control panel context menu
  + Envelopes: fix automation potentially lost when moving media items across tracks that do not have a matching FX envelope [p=2409245]
  + Envelopes: fix automation items potentially left behind when moving media items across tracks
  + Envelopes: improve snapping behavior for Audio Unit enum parameters
  + Envelopes: support aliasing FX parameter from envelope control panel context menu
  + FX: add LV2 plug-in support
  + FX: add option to hide FX comment field in chain view
  + FX: support mousewheel processing for embedded FX if they have focus
  + FX: fix undo/redo display issues for FX comments
  + FX: limit generic UI vertical size to better fit on screen
  + FX: improve embedded FX update frequency when mouse-editing
  + FX: improve handling of embedded TCP/MCP controls for non-automatable parameters
  + FX: improve keyboard behavior with Waves StudioRack
  + FX: improve performance when automating parameters with the generic VST/AU GUI is open
  + FX: prevent creating envelopes for non-automatable parameters
  + JSFX: fix bugs in IX/MIDI_KeyMap, MIDI_Variant, and MIDI_DuplicateFilter [t=181206] [p=1918238]
  + JSFX: improve denormal filtering on win32 to match win64/other OSes
  + JSFX: reduce memory use
  + Linux: rename JSFX ix/ subdirectory to IX/ [t=216852]
  + Linux: support JSFX/Video Processors/EEL2 ReaScripts with SELinux heap-execution-prevention
  + Loop sections: fix media item loop section mouse modifiers interaction with reversed media [t=249263]
  + macOS: improve Big Sur compatibility by forcing various windows to use Metal
  + macOS: improve resolution of docker icon progress display during render
  + macOS: fix launching Audio MIDI Setup on Big Sur
  + macOS: fix video output flicker on recent macOS versions
  + macOS: fix possible crash in menu editor on 10.14 [t=249194]
  + Media explorer: action to insert media into active sample player looks for more recently active ReaSamplomatic5000 instance [t=249473]
  + Media explorer: don't automatically re-scan empty files
  + Media explorer: don't re-scan database files that contain no metadata [t=249280]
  + Mixer: fix issues with themes that have very large master.mcp.size set [t=249171]
  + Mute automation: fix automating active mute envelope from controller [t=247330]
  + Razor edit: add Razor Edit feature, to draw, move, copy, and transform selected areas of media and automation
  + Razor edit: support moving/copying envelope areas based on envelope type or envelope lane via mouse modifiers
  + Razor edit: when "move envelope points with items" enabled, hidden envelopes are included in area selections
  + Razor edit: actions to insert MIDI item, empty item, click source use razor edit areas if they exist
  + Razor edit: add option to always trim content behind razor edits
  + Razor edit: add preference to stretch audio media items by either adding stretch markers or splitting
  + Razor edit: display move guides when editing
  + Razor edit: auto-crossfade media items at the edge of the razor edit area [p=2396178]
  + Razor edit: support displaying overlapping media items in lanes
  + Razor edit: when auto-creating FX parameter envelopes, initialize with the current parameter value [p=2409576]
  + Razor edit: add basic API support
  + Razor edit: add settings to snap razor edits to media items and automation items
  + ReaScript: allow EEL2 scripts to call third-party API functions without using extension_api()
  + ReaScript: improve EEL2 extension_api() parameter count checking
  + ReaScript: reduce memory use of EEL2 scripts
  + ReaScript: fix IDE watch refresh when variable name changes [t=249408]
  + ReaScript: support gfx_setcursor() during capture/without mousemove [t=249619]
  + ReaScript: F1 in IDE gives context-sensitive help if possible
  + ReaScript: optimize Lua calls to plug-in API registered functions
  + ReaScript: add support for querying envelope GUID using GetSetEnvelopeInfo_String
  + ReaScript: GetTrackEnvelopeByChunkName will also find envelopes by GUID
  + Retina/HiDPI: improve knob appearance in ReaSamplomatic5000
  + Routing diagram: fix pin display issues with 64 channel tracks
  + ReaSamplomatic5000: support increasing max voices via controller/automation when UI is closed
  + Theme: add cursor_seltrack.png image to customize selected track cursor bracket
  + Theme: add theme color for selected track background
  + Video: fix possible deadlock when rendering video [t=243142]
  + VST3: support third-party embeddable UIs via new IReaperUIEmbedInterface interface
  + Windows: 32-bit build is now compiled with VS2005/ICC
  + Windows: add experimental asynchronous unbuffered disk write mode
  + Windows: ignore hidden/system directories when scanning VST paths

v6.23 - February  4 2021
  + Project bay: fix possible crash with VST that lack developer descriptions (6.20 regression)
  + RS5k: update mode combo box on state change [t=248967]
  + Video: fix floating FX processor window default size [t=249037]

v6.22 - February  4 2021
  + Audio units: improve envelope tooltips for enum parameters
  + FX: fix PDC mode resetting on undo of automation edits [t=248877]
  + FX: import of .rpl automatically disambiguates duplicate preset names within the .rpl file
  + Jump dialog: fix jumping to times greater than one hour when ruler is not in H:M:S:F [t=248559]
  + Media explorer: fix preview length when dragging selected part of media item into REAPER [t=248547]
  + Media explorer: improve behavior when browsing folders with option disabled to auto-expand directory tree [2399243]
  + MIDI export: improve accuracy when exporting gradual tempo changes
  + ReaScript: fix return value for FX_GetParameterStepSizes() on Audio Units
  + Render: support $region(name) wildcard when rendering via region matrix [t=248672]
  + Render: add preference in preferences/audio/rendering to control automatic closing of the render window
  + Save live output: improve behavior when not in 'record only while stopped' mode
  + Video: improve section/reverse behavior when previewing in the media explorer
  + Video: improve seek behavior when previewing in the media explorer
  + Windows: change default worker thread behavior to reduce excessive CPU use on many-core systems
  + Windows: fix tooltip behavior on windows that are pinned topmost

v6.21 - January 23 2021
  + macOS: fix video issues on Catalina/Big Sur and retina displays [t=248278]
  + macOS: fix ReWire-related crash on Catalina
  + macOS: revert to older macOS SDK for Catalina build (matching 6.19)
  + Media explorer: fix possible memory leak when importing or previewing media with embedded transient information
  + Media explorer: fix waveform display when zoomed in to sample level [p=2394935]
  + Media explorer: reverse inserted media if preview playback is reversed
  + Render: fix potentially rendering one sample too short [t=248280]
  + Render: fix preset loading of 2nd pass render [t=248293]
  + Track/mixer panels: improve mouse hit testing for empty FX slots [t=248355]
  + VST: don't show scan window when only built-in plugins are being scanned
  + VST: fix scanning of certain plug-ins on Windows (6.16 regression) [t=246534]

v6.20 - January 20 2021
  + Actions: add actions to render stems on 2nd pass
  + Actions: add action to set horizontal zoom to default project setting
  + Actions: fix action to select take under mouse when empty take lanes are not visible [t=246606]
  + Actions: fix capitalization of ReaScript, ReaEQ, etc in action descriptions
  + Automation items: fix potential crash when splitting automation item under mouse cursor [t=247651]
  + Batch converter: add option to preserve original file metadata when possible
  + Envelope panels: fix redraw glitch when TCP is wider than arrange view
  + Envelope panels: fix lag in display of value text
  + Freeze: disable track playback offset when freezing, re-enable when unfreezing [t=242673]
  + FX: faster preset listing/saving/loading with large preset counts/sizes
  + FX: fix JSFX/AU pin map loading when FX pin count is larger than track channel count
  + FX: FX chains can be set to compensate delay per-FX (REAPER 6.19 or earlier), per-chain, or not at all
  + FX: master FX chain can be set to compensate delay at the hardware-send level for improved performance
  + FX browser: improve selection behavior when renaming FX
  + IDE: autosuggest text (function names, variable names) while typing
  + IDE: add Ctrl+R for replace (after find), Ctrl+Return to insert line after current, Ctrl+J to jump to line number
  + IDE: support searching backwards via Ctrl+Shift+G or Shift+F3
  + IDE: double-clicking a user function name jumps to the function definition
  + IDE: improve search function, up/down changes search modes (substring, word, token)
  + IDE: reduce flicker in watch lists
  + IDE: show viewable line range in status line
  + IDE: add preference for watch window update interval
  + IDE: remember watch-enabled state
  + JSFX: add bit meter
  + JSFX: change IDE behavior to use Ctrl+Shift+R to open an import line (vs Ctrl+R which is now used for search/replace)
  + macOS: improve listview appearance on Big Sur
  + macOS: Catalina/Big Sur builds are now built with the latest Apple toolchain
  + macOS: update table column alignment properly for multi-tab list controls (project bay, etc)
  + Media explorer: action to open arrange view media in media explorer respects autoplay preference
  + Media explorer: add action to calculate peak volume for media files
  + Media explorer: add action to preview random file
  + Media explorer: add action to re-read metadata from media files
  + Media explorer: add action to reverse preview playback
  + Media explorer: add options to normalize preview volume and/or normalize volume when inserting media
  + Media explorer: add track number column (not displayed by default)
  + Media explorer: after deleting files, select the next file
  + Media explorer: apply volume, pitch, rate adjustments properly when inserting multiple files at once
  + Media explorer: clean up shortcut view after reordering shortcuts
  + Media explorer: clear search cache when renaming files or when new files are added
  + Media explorer: display arrow cursor over preview scrollbars on Windows
  + Media explorer: display length, sample rate, audio channels for video media
  + Media explorer: don't re-initialize column order when the a new REAPER release changes the number of defined columns
  + Media explorer: don't search bit depth/bitrate/length metadata (you can sort those columns though)
  + Media explorer: remove actions to delete folders without recycling or confirming
  + Media explorer: fix action to play previously played media (A/B test)
  + Media explorer: fix context menu action to browse to folder for search results or database entry
  + Media explorer: fix customized column order on macos
  + Media explorer: fix displayed length/bit depth for some non-standard .wav files
  + Media explorer: fix possible crash when removing a track/project that is being used for preview
  + Media explorer: if option enabled, don't auto-play when initiating drag of unselected file [p=2384175]
  + Media explorer: improve accessibility of context menus via application-key
  + Media explorer: improve column sorting for length and bits columns
  + Media explorer: improve context menu organization and consistency
  + Media explorer: map reverse/forward preview playback to shift+leftarrow/shift+rightarrow by default
  + Media explorer: more accurate sample drawing when highly zoomed in
  + Media explorer: preserve custom column order when docking/undocking media explorer
  + Media explorer: prevent writing multiple ID3 tags to .wav files
  + Media explorer: respect preferences to apply preview volume/pitch/rate when inserting into reaper via drag or action
  + Media explorer: restore selection after renaming file [p=2389201]
  + Media explorer: optionally select enclosing folder in shortcut list when browsing
  + Media explorer: show filenames that will be removed when prompting to confirm deletion
  + Media explorer: tempo adjust (approximately) visual preview length of media when dragging from media explorer
  + Media explorer: improve behavior of various media format previews going to multichannel tracks [p=2370348]
  + Media items: don't reset manual fades when editing item contents with auto-crossfade enabled [t=239815]
  + Media items: fix actions to move items up/down one lane [t=246887]
  + Media items: fix potential crash when replacing media source for multiple items with empty items in the project [p=2380184]
  + Metadata: add preset support
  + Metadata: fix decoding very large BWF start offset values [p=2380383]
  + MIDI: fix pre-FX fader-scaled volume envelopes transforming MIDI incorrectly
  + MIDI: preserve MIDI port messages when importing MIDI to multiple tracks
  + MIDI: support RIFF-based MIDI files [t=246816]
  + MIDI editor: add undo state when editing MIDI event properties
  + MIDI editor: fix display of linear tempo changes [t=238747]
  + MIDI editor: fix event list delete button [t=242993]
  + MIDI editor: support inserting and editing port messages, though REAPER doesn't do anything with those messages
  + MIDI export: improve exporting linear tempo changes [t=247636]
  + MIDI recording: don't extend existing MIDI items past the start of a following MIDI item [t=226129]
  + MIDI recording: fix behavior where existing MIDI item will not be extended if recording begins, but nothing is actually recorded, before the item start position
  + MIDI recording: in overdub/replace mode, don't extend existing MIDI items until MIDI has actually been recorded on the track
  + Mouse modifiers: add modifier to move item loop section contents respecting snap
  + MP3: encode chapter tags in timeline order [t=246910]
  + Project bay: fix FX developer display when using multi-out VSTs
  + Project bay: fix undo point when dragging from FX tab to project
  + ReaDelay: display formatted parameter value in REAPER envelope edit dialog
  + ReaScript: EnumerateFiles()/EnumerateSubDirectories() allow manual cache invalidation [t=203235]
  + ReaScript: fix EnumerateFiles()/EnumerateSubDirectories() cache bug [t=203235]
  + ReaScript: GetFocusedFX()/GetFocusedFX2() support monitoring FX
  + ReaScript: Lua gmem_attach() returns previous global memory segment name [t=240447]
  + Render: add support for 2nd pass render
  + Render: adjust 2nd pass render for resampler latency [p=2381827]
  + Ruler: when displaying time, always show X:XX.XXX even if the last X is zero
  + Scrub: add actions to temporarily enable looped-segment scrubbing
  + Super8: fix recording bug when in project/playback sync mode and gate was left enabled
  + Timecode: format as HH:MM:SS:FF rather than H:MM:SS:FF everywhere (timeline, transport, wildcards, etc)
  + Timecode: when timeline is in H:M:S:F format, interpret user-entered H.M.S.F as H:M:S:F
  + Track panel: show bypassed indication for embedded FX, shift+click header to toggle bypass
  + Undo: add undo point including filename when using save-as to save to a new project file
  + Undo: display the undo point corresponding with the last save in undo history
  + Video: do not show intermediate frames when seeking playback long distances
  + Video: do not show intermediate frames when stopped and repositioning cursor, unless rapidly repositioning cursor
  + Video: fix playing early frames when looping
  + Video: fix possible parameter automation display flicker while stopped
  + Video: fix timing glitch at playback start [t=239873]
  + Video: improve playback behavior immediately following small backwards seek
  + Video: add optional project settings to have video affected by solo of other tracks, mute of parent tracks
  + Video: add project setting to control where in the frame automation is evaluated [t=244908]
  + Video: spectrogram synthesis preset bugfixes
  + VST: improve plug-in scanning status messages
  + VST: add option to not scan for new/updated plug-ins on startup
  + VST: preferences window lists skipped plug-ins and allows manual re-scan
  + VST: reduce display update frequency when scanning filesystem
  + VST: fix IContextMenu compatibility issue
  + WAV: fix importing embedded transients as transient markers [p=2386342]
  + WAV: import embedded cues properly for certain files [t=247260]
  + WAV: list embedded cues in media source properties dialog
  + Wildcards: add $marker(name) and $region(name) wildcards, for example $marker(ISRC) will resolve to "1234" if there is a project marker named "ISRC=1234" within the rendered time range
  + Wildcards: format timecode as HH.MM.SS.FF rather than H-MM-SS-FF (saved settings in existing projects are not affected)
  + Wildcards: make help text more consistent for start/end/length wildcards (sorry, localizers)
  + Windows: allow longer search strings in Media Explorer [t=245795]
  + Windows: installed executables are now signed

v6.19 - December 18 2020
  + Accessibility: option to allow space key for navigation now affects track/marker managers, track routing window, envelope window, fade editor
  + Accessibility: add option in preferences/fades to display accessible (but cryptic) fade names instead of images in menus
  + API: add PCM_SOURCE_EXT_GETBITRATE
  + API: PCM_SOURCE_EXT_ENUMCUES_EX distinguishes between cues and chapter markers
  + Control surfaces: do not send duplicate CSURF_EXT_SETFXPARAM notifications in response to automation
  + EDL: fix importing media item fade shapes
  + Elastique: update to v3.3.3
  + Localization: support UTF-8 localization of wav render formats
  + Media explorer: add actions to increase/decrease preview volume
  + Media explorer: add action to reset columns to default order and width, re-initialize view when new fields are added in a REAPER update
  + Media explorer: add option to group folders at top
  + Media explorer: add scrollbars to zoom/scroll preview display and increase/decrease waveform height, option to show/hide scrollbars
  + Media explorer: if options enabled, apply volume/pitch/rate adjustments when adding media to ReaSamploMatic5000 [t=246343]
  + Media explorer: by default, pick up beat-synced preview immediately, add option to wait for next measure to start
  + Media explorer: display sample rate, channel count, bit depth/bitrate
  + Media explorer: display media length
  + Media explorer: display media source preferred position (BWF start offset or any other available reference metadata)
  + Media explorer: display preview position/length in milliseconds
  + Media explorer: faster metadata reading for .ogg files
  + Media explorer: fix action to go to parent folder when persistent search is active [t=245677]
  + Media explorer: fix duplicate metadata potentially being written to files
  + Media explorer: fix actions to scan all databases [t=246221]
  + Media explorer: fix synchronized previews being off by a buffer block
  + Media explorer: preserve unrecognized ID3 tag elements when editing metadata
  + Media explorer: set media offline before rewriting metadata, in case the file is already in use in the arrange view
  + Media explorer: snap cursor to media cues/regions/transients/chapters
  + Media explorer: support customizing file last-modified date display format
  + Media explorer: preserve user-edited (but not yet written) metadata when re-reading metadata from media [p=2376166]
  + Media explorer: support temporary marks for files, to track what has been previewed or other uses
  + Media explorer: support updating metadata for multi-stream files where the vorbis stream is not first
  + Media explorer: support writing edited source preferred position back to media files
  + Media explorer: add refresh button, add theme images for back/forward/up/refresh
  + Media items: preserve manual fade shape when clearing auto-crossfade [t=240579]
  + Media items: support applying media source BWF start offset to item ruler [t=236220]
  + Media items: support section/reverse parameters in item ruler [t=245993]
  + Metadata: support time reference (start offset) in mp3, flac, ogg, opus, wavpack in addition to wav
  + Metadata: support iXML metadata in flac, mp3 files
  + Metadata: display mp3 event markers in media explorer and arrange view
  + Metadata: list AXML metadata separately from BWF (they had been combined for legacy reasons)
  + Metadata: list render metadata schemes in alphabetical order
  + Metadata: improve compatibility when editing metadata for OGG files with large page sizes
  + MIDI: fix adjusting imported MIDI length to match project tempo map (6.17 regression) [p=2373470]
  + Peaks: warn if peaks are not buildable because alternate peaks path is configured but invalid
  + Project save: improve behavior when copying symlinked media on macOS/Linux
  + ReaScript: add B_RECMON_IN_EFFECT to GetMediaTrackInfo_Value()
  + ReaScript: add GetMediaFileMetadata
  + ReaScript: support UTF-8 font names on Windows [t=190468]
  + ReaScript: do not run deferred scripts during project load [p=2379014]
  + Recording: fix potentially incorrect fadeout created when recording in item autopunch mode
  + Render: activate "apply" button after browsing to new output folder [p=2375479]
  + Render: respect option to embed project name in BWF when other BWF metadata is provided
  + Subprojects: embed subproject metadata in rendered .rpp-prox file [p=2378755]
  + Subprojects: support embedding source preferred position (BWF start offset)
  + Takes: fix potential crash when replacing multiple occurrences of the same source media [p=2378037]
  + Track manager: fix reordering tracks when using various sort modes [t=246511]
  + Video processors: support UTF-8 font names on Windows [t=190468]
  + VST: decrease latency of VST3 channel context notifications
  + VST: support "REAPER record monitor in effect" attribute for VST3 channel contexts
  + Wavpack: improve interaction between render metadata and BWF settings
  + Wildcards: add $dayname and $datetime wildcards

v6.18 - December 1 2020
  + Media explorer: add option to not apply preview pitch/rate to inserted media item [t=245666]
  + Media explorer: fix UTF-8 support for search field on Windows [p=2371567]
  + Media explorer: fix potential failure editing .ogg file metadata [t=245670]
  + Media explorer: allow editing metadata of multi-stream .ogg files
  + ReaScript: improve documentation for GetAppVersion and GetOS
  + VST: avoid scanning recursive symlinks
  + VST: improve startup scanning user interface, allow cancel of long scans

v6.17 - November 29 2020
  + FX shortcuts: fix issues with shortcut creation [t=245551]
  + JSFX: support file_open() of media with absolute path [t=212561]
  + JSFX: full recompile/reset in IDE closes any files opened via file_open()
  + Linux: fix potential hang on startup [p=2370672]
  + macOS-arm64: add mp3 encoder
  + Media: fix adjusting media based on embedded tempo when not importing from media explorer [t=245407]
  + Media explorer: add actions to set/adjust pitch
  + Media explorer: add action to insert media at BWF source preferred position
  + Media explorer: display error message when unable to create new database file [t=245376]
  + Media explorer: finer control for pitch and rate knobs
  + Media explorer: fix customizing Show and Options menus
  + Media explorer: fix possible macOS crash when removing database
  + Media explorer: if fine pitch adjustment set, further semitone or quartertone adjustments will preserve the fractional adjustment [p=2368875]
  + Media explorer: when replacing source media and stretching to fit, stretch active take only [p=2369092]
  + Media explorer: improve handling of stereo mp3 files played through multichannel tracks [t=242507]
  + ReaScript: GetOS() now returns macOS-arm64 on arm64
  + ReaScript: restore 6.15 OS-subtype GetAppVersion() behavior, add 6.16/macOS-arm64 for arm64
  + ReaScript: fix insertion of ReWire plug-ins by qualified name with TrackFX_AddByName() etc
  + Render queue: order queued items by time
  + Windows: disable DX scanning by default on new installations

v6.16 - November 23 2020
  + Anticipative FX: fix excessive buffering when feedback loop w/ PDC present but feedback is disabled
  + Dynamic split: support ripple editing when removing silence
  + FX: improve display issues when FX resized and chain is displayed on right side of view
  + Linux: improve splash screen update frequency
  + macOS: update to ICC 18.5
  + Media: display all supported metadata in source properties dialog
  + Media: when matching tempo for preview or import, use any suitable metadata, if it matches the beat length of the media [p=2309774]
  + Media explorer: display metadata in filesystem folder view
  + Media explorer: automatically populate metadata for media files that are visible in the browser list
  + Media explorer: support editing metadata in-place in the list
  + Media explorer: support editing media file metadata in filesystem folder view
  + Media explorer: support writing edited metadata to media files
  + Media explorer: support marking media files as favorites, option to group favorites at the top
  + Media explorer: add playback rate control
  + Media explorer: remove option to use Windows Explorer for browsing
  + Media explorer: update databases after moving or renaming files
  + Media explorer: actions to insert on a new track create a new track after the last touched track
  + Media explorer: add actions to go to shortcuts
  + Media explorer: add action to rescan all databases
  + Media explorer: add action to collapse all folders
  + Media explorer: add action to create subfolder
  + Media explorer: add action to delete files immediately (no confirmation/trash/recycle)
  + Media explorer: add action to insert media into sample player on new or last touched track
  + Media explorer: add action to open selected file in explorer/finder
  + Media explorer: add action to randomize file list
  + Media explorer: add option for search string to persist when browsing to new folder or database
  + Media explorer: add option for vertical layout
  + Media explorer: add option to hide file extension when file type column is displayed
  + Media explorer: add option to hide folders in file list
  + Media explorer: add option to hide path dropdown box
  + Media explorer: add option to try to open non-media files
  + Media explorer: add various display zoom actions
  + Media explorer: add option to auto-stop when initiating file drag
  + Media explorer: automatically update filesystem folder view when folder contents change
  + Media explorer: clear preview playback when changing selection in browser
  + Media explorer: column sort preferences persist between REAPER sessions
  + Media explorer: confirm before deleting database
  + Media explorer: display embedded transients in preview
  + Media explorer: display mp3, flac, ogg, opus chapter markers in preview and optionally arrange view
  + Media explorer: double-click preview display to select region between slices, cues, chapters
  + Media explorer: faster metadata reading for various file types
  + Media explorer: fix handling of shortcuts/paths with odd paths ending in space, etc [t=239608]
  + Media explorer: improve macOS retina drawing of peaks
  + Media explorer: insert media on most recently selected track, not first selected track, consistent with copy/paste
  + Media explorer: insert new shortcuts or databases immediately following the currently selected shortcut
  + Media explorer: more efficient duplicate prevention with gigantic databases
  + Media explorer: move pitch knob options to pi

jcsteh:
Yes, as noted above, the new features here (including the changed Peak
Watcher) are not expected to work fully in REAPER 6.29. You need REAPER
6.30, which is not yet released. We are testing with pre-release builds.

Lo-lo78:
Yes, I'm trying. Very interesting!

Lo-lo78:
Weird thing, I have a project that I have at -11.0 lufs and the JS 
loudness_meter gives it to me at -9.6.With SWS: SWS / BR: Analyze 
loudness ... It gives me -11.0 lufs. Very strange!

Lo-lo78:
Try this mix I made at -11.0 lufs.

https://drive.google.com/uc?id=1xWaRKCCpPfwTa77YZ0ljcUtTf-2OL9wl&export=download

With SWS it's ok but if you use the JS loudness_meter it will give you 
other results.

Lo-lo78:
Item: MichaelGarding_Michael'sTheme_ManoloVazquez.flac;
Peak Watcher;
Level type: peak dB: -0.4 -0.0
integrated LUFS: -9.6 -9.6
Level type: momentary LUFS: -6.9 -6.9
Level type: short term LUFS: -7.7 -7.7
Level type: integrated RMS: -10.6 -10.6

SWS/BR: Analyze loudness...; Analyze selected items:
Item: MichaelGarding_Michael'sTheme_ManoloVazquez.flac;
Integrated: -11.0 LUFS;
Range: 4.9 LU;
True peak: -inf; Maximum short-term: -7.9 LUFS;
Maximum momentary: -6.9 LUFS

SWS: Analyze and display item peak and RMS (entire item):

Peak level: Channel 1 = -1.25 dB Channel 2 = -1.25 dB
RMS level: Channel 1 = -13.46 dB Channel 2 = -13.98 dB


Not all results match, what I don't understand is peak level, integrated 
LUFS and RMS.
With SWS there are other numbers.

Lo-lo78:
James, check out these actions, they are very interesting:


•Description: Calculate loudness of master mix via dry run render; 
Command ID: 42440

•Description: Calculate loudness of master mix within time selection via 
dry run render; Command ID: 42441
•Description: Calculate loudness of selected items, including take and 
track FX and settings, via dry run render; Command ID: 42437
•Description: Calculate loudness of selected tracks via dry run render; 
Command ID: 42438
•Description: Calculate loudness of selected tracks within time 
selection via dry run render; Command ID: 42439


The actions are more correct than the JS plugins.

jcsteh:
OSARA can't magically use the functionality behind those actions within Peak Watcher precisely because they have to render to do what they do. I'm not sure why the JSFX is less accurate. Did you find the EBUR128 JSFX (which is used by OSARA main snapshots) to be more accurate?

Lo-lo78:
Unfortunately it is the same.

JS: ebur128_analysis:

-9.6 -9.6.

I don't understand, there must be a reason but I don't have a guess.


Il 09/06/2021 11:08, James Teh ha scritto:

jcsteh:
Could you perhaps have your master trim volume set? The JSFX is going to run where other effects run, which means pre-fader. If your master trim volume is non-0, that might explain the different results, since REAPER's builtin actions probably account for that somehow. There's no way I can fix that in OSARA, though.

Lo-lo78:
It is a virtual render, I noticed that in the dialog box there is the 
button: Dry Run (no output) It is different from the render it is as if 
it were an analysis, nothing is loaded into the HD Maybe it uses RAM , 
I'm not sure!

Lo-lo78:
The project is this:


Master 0.00

1 FX JS: loudness_meter

Track 0.00

1 MichaelGarding_Michael'sTheme_ManoloVazquez 1 item


I ask other people to understand why! Thanks!

jcsteh:
Oh. I think I see why you're seeing different numbers. Integrated LUFS is calculated over the whole segment of audio. It's not the maximum level; it's an average. You probably have Peak Watcher set to "Hold highest level", which is useful for finding peaks but not for calculating integrated loudness. If you turn that off, you should get the same values.

ScottChesworth:
> Not all results match, what I don't understand is peak level, integrated LUFS and RMS. With SWS there are other numbers.
> 
I can't tell you why this happens on a technical level, but it's not unusual for real-time metering to display slightly different results compared to an offline calculation. It's also not unusual for different real-time meters to produce varied results. Advice from the sighted mastering engineers I know who have more metering options available to them is to choose a meter, learn how it behaves with some references, then stick to that one meter as often as you can, factoring its quirks into your decisions.
In other words, keep the points of inconcistency to a minimum, or you'll end up chasing your tail.

jcsteh:
> I can't tell you why this happens on a technical level, but it's not unusual for real-time metering to display slightly different results compared to an offline calculation.

In this case, I very much suspect it's the hold highest level thing, though.

ScottChesworth:
> If you turn that off, you should get the same values.
> 
You'd likely need to turn that off *and* play the project all the way through. An average can only be calculated consistently if fed the exact same material. at which point, it'll almost always be quicker to do an offline calculation anyway unless you're running insane amounts of processing that causes Reaper to render slower than real-time. As I said above, it's not ideal, but it is normal.

Lo-lo78:
I'm testing with an expert tonight. I'll send you feedback right away! 
09/06/2021 12:50, ScottChesworth wrote:

ptorpey:
If I am understanding these new actions correctly, does that mean that we can now get the analysis post fader, i.e., including the impact of any effects and VST instruments on the track?  This wasn’t possible before with the SWS loudness analysis tools.  If that is correct I’m very happy to see this capability.

 

Also, the “Dry run” rendering sounds interesting if one doesn’t have to keep creating extra audio files on disk just to analyze a track.

 

Cool!

 

--Pete

ScottChesworth:
Haven't tested it here yet @ptorpey, but on paper, yep, should be doable. Make sure you've got a high block size specified for rendering in Prefs so those dry run renders take minimal time to complete.

ptorpey:
Thanks for the hint about the large block sizes to speed up redners.  Looking forward to the 6.30 release and testing this out.  Very welcome feature if it does what we think it will.

 

--pete

Lo-lo78:
Ok, now everything is clear. The problem that with the LUFS: integrated, 
momentary and short term it is not possible to read the maximum peak. 
The screen reader must not read the maximum peak but with "alt + f11 and 
alt + f12" but the realtime because the LUFS result is at the end of the 
song. So here's why the mix at -11.0 LUFS looked at -9.6, what matters 
is the final result. It is not possible to reset "alt + f10" because the 
LUFS is valid for all the audio stream from start to finish. If you 
reset in the middle of the song, you won't get the exact result

Therefore, the LUFS of the Peak Watcher cannot be read automatically 
from the summary but it must be the user who reads the result from time 
to time Similar to: OSARA: Report current peak for channel 1 of current 
trackOSARA: Report current peak for channel 2 of current track This way 
you can understand LUFS of the song.

Greetings!

jcsteh:
That's correct and is how integrated levels are designed to work. As I
noted, you can disable "Hold highest level" to prevent the maximum from
sticking. Then, if you wanted to know the final integrated measurement, you
would listen to the entire piece of audio and read the level at the end.

Lo-lo78:
It's ok James, it works fine!

If you set: When level reaches: 0.00, Hold highest level: disabled and 
do ok it works perfectly! Thanks!

Lo-lo78:
Hi James, Reaper version 6.30 came out yesterday. I think you can merge! 
Yes, it is very nice what they added! Greetings!

PR #539, opened 2021-06-09
Make it possible to open the OSARA FX Parameters dialog for the selected effect in the FX chain dialog.

### LINKED ISSUE #538: Is it possible to pass directly to the OSARA: FX parameters to the plugin from the FX chain?
Hello,
98% of the plugins are not accessible and I thought it would be convenient to add a plugin to the chain and go directly into the OSARA: FX parameters.
Guy:
Add FX to Track 2 Filter Dialog:
tal j 8;
VST3i: TAL-J-8 (TAL-Togu Audio Line) (6 out); after sending the effects I will have, with copy paste:
VST3i: TAL-J-8 (TAL-Togu Audio Line) (6 out)
VST3: FETISH (AnalogObsession)
I select the TAL and with the "P" or "shift + P" key you enter directly on FX parameters without doing "ESC" and then "P", look for the FX and then in the FX parameters ..
In this way you enter more quickly in the parameters of the plugin, then for the accessible ones you already do "F6".
Greetings!

jcsteh:
You do this by selecting the effect in the list and pressing alt+p (OSARA: View parameters for current track/item/FX).

I'm not sure if this is going to work on Mac. Assuming it builds, it will need testing.

Fixes #538.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -431,7 +431,7 @@ This is useful for parameters which are tedious or impossible to access otherwis
 
 #### Track/Item Parameters
 To access the parameter list for a track or an item, select the track or item you wish to work with.
-Then, press alt+p (OSARA: View parameters for current track/item).
+Then, press alt+p (OSARA: View parameters for current track/item/FX (depending on focus)).
 
 #### FX Parameters
 Many effects are unfortunately either partially or completely inaccessible.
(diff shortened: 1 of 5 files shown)


pitermach:
Seems to be working great on Mac, not seeing any issues.

Lo-lo78:
Yes, it works fine!

👍

PR #544, opened 2021-06-30
Allow the enabled state/type of ReaEQ bands to be adjusted from the FX Parameters dialog.

### LINKED ISSUE #543: Can ReaEQ band type somehow be inserted into the OSARA Parameters dialog?
I've just had a chat with Cockos about this. Their reason for not exposing band type as a ReaEQ parameter is that it can cause a glitch in the audio when changed. However, I still think it would be useful to have access to this via the Parameters dialog if possible, and Justin offered API details as follows:
It could be done via the TrackFX_SetNamedConfigParm() and TrackFX_GetNamedConfigParm() APIs, with "BANDTYPE0"/"BANDTYPE1" etc.
the values are a string representation of a number, being:
"0" -> "Low Shelf"
"1" -> "High Shelf"
"2" -> "Band (alt 2)"
"3" -> "Low Pass"
"4" -> "High Pass"
"5" -> "All Pass"
"6" -> "Notch"
"7" -> "Band Pass"
"8" -> "Band"
"9" -> "Band (alt)"

@jcsteh, is there any mileage in OSARA being able to manually insert these into the params dialog somehow?

jcsteh:
Fixes #543.
I've done some preliminary testing, but I'd like to get confirmation from a couple of users that this works and doesn't break things (for both ReaEQ and other effects) before merging.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -611,21 +611,33 @@ class ParamsDialog {
 // and with the prefix by passing it and fetching the functions dynamically.
 template<typename ReaperObj>
 class FxParam;
+template<typename ReaperObj>
+class FxNamedConfigParam;
 
 template<typename ReaperObj>
 class FxParams: public ParamSource {
 	friend class FxParam<ReaperObj>;
+	friend class FxNamedConfigParam<ReaperObj>;
 
 	private:
 	ReaperObj* obj;
 	int fx;
+	// Named config params can't be enumerated, so we have to build a list of
+	// these based on the effect and the known named parameters it supports. See
+	// initNamedConfigParams().
+	vector<FxNamedConfigParam<ReaperObj>> namedConfigParams;
 	int (*_GetNumParams)(ReaperObj*, int);
+	bool (*_GetFXName)(ReaperObj*, int, char*, int);
 	bool (*_GetParamName)(ReaperObj*, int, int, char*, int);
 	double (*_GetParam)(ReaperObj*, int, int, double*, double*);
 	bool (*_GetParameterStepSizes)(ReaperObj*, int, int, double*, double*,
 		double*, bool*);
 	bool (*_SetParam)(ReaperObj*, int, int, double);
 	bool (*_FormatParamValue)(ReaperObj*, int, int, double, char*, int);
+	bool (*_GetNamedConfigParm)(ReaperObj*, int, const char*, char*, int);
+	bool (*_SetNamedConfigParm)(ReaperObj*, int, const char*, const char*);
+
+	void initNamedConfigParams();
 
 	public:
 
@@ -633,12 +645,21 @@ class FxParams: public ParamSource {
 			obj(obj), fx(fx) {
 		// Get functions.
 		*(void**)&this->_GetNumParams = plugin_getapi((apiPrefix + "_GetNumParams").c_str());
+		*(void**)&this->_GetFXName = plugin_getapi(
+			(apiPrefix + "_GetFXName").c_str());
 		*(void**)&this->_GetParamName = plugin_getapi((apiPrefix + "_GetParamName").c_str());
 		*(void**)&this->_GetParam = plugin_getapi((apiPrefix + "_GetParam").c_str());
 		*(void**)&this->_GetParameterStepSizes = plugin_getapi((apiPrefix +
 			"_GetParameterStepSizes").c_str());
 		*(void**)&this->_SetParam = plugin_getapi((apiPrefix + "_SetParam").c_str());
 		*(void**)&this->_FormatParamValue = plugin_getapi((apiPrefix + "_FormatParamValue").c_str());
+		*(void**)&this->_GetNamedConfigParm = plugin_getapi(
+			(apiPrefix + "_GetNamedConfigParm").c_str());
+		*(void**)&this->_SetNamedConfigParm = plugin_getapi(
+			(apiPrefix + "_SetNamedConfigParm").c_str());
+		if (fx >= 0) {
+			this->initNamedConfigParams();
+		}
 	}
 
 	string getTitle() {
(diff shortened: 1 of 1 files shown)


ScottChesworth:
The Mac build passes and Windows fails? What madness is this? Is ice cream even cold anymore? :)

jcsteh:
Fussy bloody compiler. Looks like the version of Visual Studio on AppVeyor
is slightly fussier than mine. I guess I need to update.

ScottChesworth:
Working here on Win10 with Reaper 6.29. Got two minor nitpicks and two UX questions for ya though:

1. Should we be following Cockos' style for parameter names? That being the case, first letter of each word would be uppercase, "Band 1 Type".

2. Does past tense make sense for the new Enabled parameters? I think this would be clearer for newbs without the final D. "Enabled" could be misinterpreted as a report of current status, whereas "Enable" suggests it's a setting you have control of.

I did warn ya they were nitpicky... :)

3. Would it be possible to slot the new Enabled and Type params in above each band? Ideally, the params list would look something like:
band 1 enabled (1)
band 1 type (2)
Freq-Low Shelf (3)
Gain-Low Shelf (4)
Q-Low Shelf (5)
Repeat for subsequent bands?

4. If I instantiate ReaEQ at its default settings, I have 4 bands. The lowest is a Low Shelf type, then 2 bands, and a High Shelf.
Now if I use the "band 1 type" param to change band 1 to be a High Pass, the Freq, Gain and Q params still report as "Freq-Low Shelf", "Gain-Low Shelf" and "Q-Low Shelf".
If I close and reopen the Parameters dialog, the labeling for band 1 is updated to reflect the new band type.
Any chance this labeling could be done on the fly without needing to close and reload the dialog? Happy to take this back to Cockos if it's something that needs to be handled by the API itself.

jcsteh:
> 1. Should we be following Cockos' style for parameter names? That being the case, first letter of each word would be uppercase, "Band 1 Type".

Cockos are somewhat inconsistent. You're correct that here, it's title case:
Freq-Band 2
But here, they capitalise only the first letter of the first word:
1: Shift (full range)
To be consistent with ReaEQ, we should probably do:
Type-Band 1
However, I feel like that's less user friendly. Also, ReaEQ calls band 1 low shelf if it's set that way:
Freq-Low Shelf
That doesn't make sense for band type, which you might be changing.
Long story short: Cockos are inconsistent, but I think it makes sense to capitalise the first letter.

> 2. Does past tense make sense for the new Enabled parameters? I think this would be clearer for newbs without the final D. "Enabled" could be misinterpreted as a report of current status, whereas "Enable" suggests it's a setting you have control of.

This would be inconsistent with "enabled" in, say, ReaPitch... but again, things aren't consistent here. Still, your reasoning makes sense.

GarthHum:
Agree with the order of parameters proposed by Scot. 
I had the same issue with the reporting of the parameter names not updating until closing and reopening the parameters dialog on Mac. 
I am now a bit confused about the proposed change of name for the new band enable parameter. As it is, I assumed it would enable or disable that particular band, and this is what it seems to do. What am I missing?

Lo-lo78:
Hi, very nice; I know that it is not possible to do this but I say it: 
(rofl) it would be interesting to also enable:

1.

Add band

2.

Remove band

3.

This is very important:

Output gain

4.

also interesting:

Show phase

Greetings!

ScottChesworth:
> I am now a bit confused about the proposed change of name for the new band enable parameter. As it is, I assumed it would enable or disable that particular band, and this is what it seems to do. What am I missing?
> 
Seems like you're looking at it after the change I proposed had already been made. It used to be "enabled" past tense, now it's "enable" present tense, to avoid the potential for the word "enabled" to be confused with a status report.

Lo-lo78:
Sorry, I wanted to point out that in the Band type after the high shelf 
there is Band, the Band (alt) comes after the Band Pass.

For example:

Low Shelf

High Shelf

Band

Low Pass

High Pass

All Pass

Notch

Band Pass

Band (alt)

Band (alt 2)

jcsteh:
Yeah, the order of the band types is different in the API for some reason compared to what REAPER shows in its UI. OSARA is using the API order. I don't know why it's different. I don't really think it matters, but I can reorder them easily enough.

jcsteh:
I can't slot the band enable and type parameters in with the other ones, sorry. That would basically require a separate Parameters dialog just for ReaEQ. At that point, I'd rather throw this code away and just say use the ReaEQ UI which already allows you to change all of this.

The issue with the parameter names not updating is also tricky. To fix that, I'd have to rebuild the entire parameters list every time any parameter changes, since we don't know which parameters might cause other parameter names to change. I might be able to come up with some way around that one though.

jcsteh:
Blah, I think I shouldn't have wasted time on this. It's raising too many issues.

ScottChesworth:
> Blah, I think I shouldn't have wasted time on this. It's raising too many issues.
> 
As things stand in the most recent AppVeyor build, it's working. If that's as far as you can go with it, that's fair enough, it's still an excellent addition and worth merging IMO.

Lo-lo78:
Hi, I think you are already informed. Band 1 enable, Band 1 type up to 
Band 4 enable, Band 4 type everything works. Is it normal that the 
frequency, gain and Q have no effect? ​​They don't work for me.
Greetings!

ScottChesworth:
> Is it normal that the frequency, gain and Q have no effect?
> 
Yikes, guess I forgot to check whether the parameters actually work! @jcsteh, all sliders beyond band 4 enable seem to be snapping to -100% here too. Sorry I didn't spot this earlier, embarrassing oversight trophy unlocked.

jcsteh:
> Yeah, the order of the band types is different in the API for some reason compared to what REAPER shows in its UI. OSARA is using the API order.

I reordered these so they're the same order as the ReaEQ UI.

> Is it normal that the frequency, gain and Q have no effect?

Oops. Rookie error; I had a plus instead of a minus somewhere. Fixed.

Lo-lo78:
Hi, yes James, it works fine! 👍

PR #554, opened 2021-07-11
Catch possible exception when trying to determine the note for MIDI event list preview.

### LINKED ISSUE #545: Critical bug with Osara in MIDI event editor
- [x] 👋Hello everyone! How's life going First hoping that it is good, but if we ask Osara I think she is a bit sad hehe.

After going through the figurative, let's get to the heart of the matter.
When the Midi event editor is opened from obviously Reaper, and CC carriles such as Chorus, Rever, Tremolo, etc. are added, and the following critical message starts to be manipulated with arrows, the following critical message appears, closing Reaper and deleting everything you have in your project if unfortunately you have not saved.
Microsoft Visual C ++ Runtime Library Dialog Error!

Program: C: \ Program Files \ REAPER (x64) \ reaper.exe

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.

It should be noted that in the community where we handle Osara several users have reported the same bug to me and if you ask, yes, we have the complete collection of Microsoft Visual C ++ Redistributables Libraries, 2005-2008-2010-2012-2013-2015-2019
With Osara 2020 the error does not happen, only with 2021. I imagine that something must be there.
Even trying to replicate the error, it was enough to add a carril CC and move with arrows to give me the mentioned error. He gives other teammates the error when they select more than 2 CC carriles at the same time to eliminate. I would have liked to report the error much earlier, but between work and many things I had not given the time and also that I hardly use the MIDI event editor.
I hope and my message is understood, forgive my bad English.
Who wants to have a beer? Because I do. Or maybe a coffee, greetings from Ecuador brothers. To continue with this project.

jcsteh:
It seems that REAPER language packs can translate note names.
If a translated note name matches the start of an English note name (e.g. "Fa" will match "F"), stoi (used to get the octave) will then barf because "a" isn't a number.
For now, just catch this exception and gracefully ignore it to prevent a crash.
In future, we're going to have to figure out how to deal with translated note names.
Fixes #545.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1092,6 +1092,9 @@ void maybePreviewCurrentNoteInEventList(HWND hwnd) {
 		try {
 			octave = stoi(noteNameWithOctave.substr(noteNameLength, 2));
 		} catch (invalid_argument) {
+			// If a REAPER language pack translates note names, we might get something
+			// unexpected here. There's nothing we can do to compensate, so just
+			// gracefully ignore this.
 			continue;
 		}
 		note.pitch = ((octave + 1) * 12) + i;
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1088,7 +1088,12 @@ void maybePreviewCurrentNoteInEventList(HWND hwnd) {
 		// Therefore, the number counts either one or two characters in noteNameWithOctave.
 		// If the note is explicitly named, the name comes after the octave and a space.
 		// As stoi simply ignores whitespace or the end of a string, all possible appearances should be covered.
-		int octave = stoi(noteNameWithOctave.substr(noteNameLength, 2));
+		int octave;
+		try {
+			octave = stoi(noteNameWithOctave.substr(noteNameLength, 2));
+		} catch (invalid_argument) {
+			continue;
+		}
 		note.pitch = ((octave + 1) * 12) + i;
 		break;
 	}


jcsteh:
@ElCrackLoko, please test a build when it appears here in a few minutes and report if this fixes the crash for you. Note that MIDI event list preview for note names that have been translated (e.g. F -> Fa) isn't going to work. I don't know how we're going to deal with this, but I figure preview not working for some notes is far better than preview crashing for some notes. :)

ElCrackLoko:
Hello my bro. The preview of the notes is not something that keeps me awake, I think that as you say, it is worse to lose the whole project when it gives us a RunTime error lol. I just tried the Osara compiled in this thread, the one that says osara_pr554-577, e7e73c42 and everything has gone great with this one. Although SOL# does not work for the preview either, I comment, and it has no similarity with the English note system since it seems to me that SOL# in your system would be G#
After testing this version I entered
[The OSARA Development Snapshots](https://osara.reaperaccessibility.com/snapshots/)
And with that version, the 2021.1pre-579,3f19b953 the error keeps happening. I thought that after compiling here it would update on the OSARA Development Snapshots page
Either I consult out of ignorance, why are there 2 Osara from July 11th, with completely different names, or how could I always get the latest Osara compiled like I got it from this thread like this from @AppVeyorBot How could I do that?
I have seen that sometimes some community partners pass an Osara that says something like Osara_AppVeyor and something else that I do not remember but the number is always different from the one on the Development Snapshots page
Thanks for fixing the bug, bro.

ScottChesworth:
> I thought that after compiling here it would update on the OSARA Development Snapshots page
> 
The builds that AppVeyorBot provides are for testing. For the change to be reflected in the downloads offered from the OSARA Snapshots page, @jcsteh has to merge the changes he's made in this pull request.
Now that you've confirmed it works, I'm sure he'll merge soon. :)

ElCrackLoko:
Thank you very much for answering my question.
What I still do not understand is how there are people who obtain Osara with the prefix (PR) and with the name of AppVeyorBot, or only AppVeyor, I do not remember and then they pass those files, mysteries of life. However, the error in this version no longer appears and that is something that makes me happy.
Thank you again.

ScottChesworth:
> What I still do not understand is how there are people who obtain Osara with the prefix (PR) and with the name of AppVeyorBot, or only AppVeyor, I do not remember and then they pass those files
> My guess is that they're just keeping a close watch on the GitHub threads and getting the PR builds the same way you did to test this fix. Maybe they're using GitHub search to display open pull requests.

PR #581, opened 2021-10-30
Add specific post command for selected events

### LINKED ISSUE #521: MIDI Editor Suggestion: Instead of Reporting How Many Notes Have Been Selected, Report How Many Events Have Been Selected
It's all in the title, really, but to clarify further, I think the changing of the feedback the pressing of CTRL + A results in could benefit from a little adjustment. The problem is that Osara doesn't take into account that CTRL + A (and related actions) don't just select notes, but all types of MIDI events. So, maybe reporting how many MIDI events have been selected would make more sense instead of just focusing on note events.

leonardder:
Closes #521

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -736,14 +736,19 @@ void postMidiSelectEvents(int command) {
 	int count=0;
 	for(;;){
 		evtIndex = MIDI_EnumSelEvts(take, evtIndex);
-		if (evtIndex == -1)
+		if (evtIndex == -1) {
 			break;
+		}
 		++count;
 	}
-	// fakeFocus = FOCUS_NOTE;
-	ostringstream s;
-	s << count << " event" << ((count == 1) ? "" : "s") << " selected";
-	outputMessage(s);
+	if (fakeFocus != FOCUS_NOTE && fakeFocus != FOCUS_CC) {
+		fakeFocus = FOCUS_NOTE;
+	}
+	// Translators: Reported when selecting events in the MIDI editor. {} will be replaced with
+	// the number of events; e.g. "2 events selected".
+	outputMessage(format(
+		translate_plural("{} event selected", "{} events selected", count),
+		count));
 }
 
 const string getMidiControlName(MediaItem_Take *take, int control, int channel) {
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -729,6 +729,23 @@ void postMidiSelectNotes(int command) {
 	outputMessage(s);
 }
 
+void postMidiSelectEvents(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int evtIndex=-1;
+	int count=0;
+	for(;;){
+		evtIndex = MIDI_EnumSelEvts(take, evtIndex);
+		if (evtIndex == -1)
+			break;
+		++count;
+	}
+	// fakeFocus = FOCUS_NOTE;
+	ostringstream s;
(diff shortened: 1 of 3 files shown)


src/midiEditorCommands.cpp
```diff
@@ -729,6 +729,23 @@ void postMidiSelectNotes(int command) {
...
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int evtIndex=-1;
+	int count=0;
+	for(;;){
+		evtIndex = MIDI_EnumSelEvts(take, evtIndex);
+		if (evtIndex == -1)
+			break;
+		++count;
+	}
+	// fakeFocus = FOCUS_NOTE;
```
leonardder, line 743: I think this line needs some debate. Do we want to introduce a specific focus for events? I don't think so, as either ccs or notes are events.

I propose:

1. Is focus is set to CC or Note, leave it alone
2. If not, set it to note

src/midiEditorCommands.cpp
```diff
@@ -729,6 +729,23 @@ void postMidiSelectNotes(int command) {
...
 
+void postMidiSelectEvents(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int evtIndex=-1;
+	int count=0;
+	for(;;){
+		evtIndex = MIDI_EnumSelEvts(take, evtIndex);
+		if (evtIndex == -1)
+			break;
```
jcsteh, line 740: nit: There's still a lot of older code that doesn't do this, but I'm trying to ensure newer code braces statement blocks even if there's only a single statement.

src/midiEditorCommands.cpp
jcsteh: Your proposal sounds good to me.

src/midiEditorCommands.cpp
```diff
@@ -729,6 +729,23 @@ void postMidiSelectNotes(int command) {
...
+		++count;
+	}
+	// fakeFocus = FOCUS_NOTE;
+	ostringstream s;
+	s << count << " event" << ((count == 1) ? "" : "s") << " selected";
```
jcsteh, line 745: This needs to be translatable. A lot of the MIDI editor code isn't translatable yet, but I'm not landing any new strings that aren't translatable. Search for "{} track copied" in reaper_osara.cpp for an example of how to do this, including the translators comment.

jcsteh (changes requested):
Thanks!

jcsteh (approved):
Perfect. Thank you.

RDMurray:
This counts each note twice, which is different to the count returned by MIDI_CountEvts(), as used by cmdMidiDeleteEvents().
I don't know why this is the case.  Is there any reason not to use MIDI_CountEvts here?

RDMurray:
Forget the last part of the above comment, of course MIDI_CountEvts doesn't count selected events.

leonardder:
I guess every note has a note off that is counted as an event? That doesn't clarify why delete events doesn't count it.

RDMurray:
Yes, that is it, MIDI_EnumSelEvts counts note off events where as MIDI_CountEvts doesn't.  I think it would be less confusing for users if note offs were not counted.  Especially as they are not counted in the event list.

We could probably get round this by subtracting the number of selected notes from the number of selected events.

leonardder:
> We could probably get round this by subtracting the number of selected notes from the number of selected events.

I'd probably prefer filtering out the note off events while enumerating.

ScottChesworth:
First bug report has landed out in the wild. Oops.

MatejGolian:
Any fix for the inclusion of note-off events in Osara messages on the horizon? It's quite misleading when Osara reports the double amount of notes being selected after pressing CTRL + A in the MIDI editor.

PR #582, opened 2021-11-01
Respect the Project Start Time Offset Set in Project Settings

### LINKED ISSUE #566: Respect measure offsets set in project settings
Project settings (alt enter) allows an offset to be made in measures.  It is quite rare but it would be nice to be able to start from a measure of zero so pressing control home gets you to 0.1.0.
I think Reaper supports this but Osara is not.
However, I'd put this request pretty low down on the priority list and if it risks complicating a load of other stuff I wouldn't say it is worthwhile.

### LINKED ISSUE #567: Respect the Project Start Time Offset Set in Project Settings
Osara currently doesn't honor the project start time offset when reporting the edit/play cursor position. In case #566 gets implemented, it would be nice if Osara had support for the project start time feature as well. Of course, provided that's possible at all.
      Thanks a lot.

leonardder:
fixes #566
fixes #567

### DIFF
--- a/src/osara.h
+++ b/src/osara.h
@@ -143,6 +143,8 @@
 #define REAPERAPI_WANT_GetMediaTrackInfo_Value
 #define REAPERAPI_WANT_GetTrackMIDINoteName
 #define REAPERAPI_WANT_get_config_var
+#define REAPERAPI_WANT_projectconfig_var_addr
+#define REAPERAPI_WANT_projectconfig_var_getoffs
 #define REAPERAPI_WANT_EnumProjects
 #define REAPERAPI_WANT_GetProjectName
 #define REAPERAPI_WANT_GetProjectTimeOffset REAPERAPI_WANT_GetProjectTimeOffset
--- a/src/osara.h
+++ b/src/osara.h
@@ -145,6 +145,7 @@
 #define REAPERAPI_WANT_get_config_var
 #define REAPERAPI_WANT_EnumProjects
 #define REAPERAPI_WANT_GetProjectName
+#define REAPERAPI_WANT_GetProjectTimeOffset REAPERAPI_WANT_GetProjectTimeOffset
 #define REAPERAPI_WANT_plugin_register
 #define REAPERAPI_WANT_GetTrackUIVolPan
 #define REAPERAPI_WANT_GetGlobalAutomationOverride
@@ -246,7 +247,7 @@ typedef enum {
 	TF_SAMPLE
 } TimeFormat;
 const TimeFormat TF_RULER = TF_NONE;
-std::string formatTime(double time, TimeFormat format=TF_RULER, bool isLength=false, bool useCache=true, bool includeZeros=true);
+std::string formatTime(double time, TimeFormat format=TF_RULER, bool isLength=false, bool useCache=true, bool includeZeros=true, bool includeProjectStartOffset=true);
 void resetTimeCache(TimeFormat excludeFormat=TF_NONE);
 std::string formatNoteLength(double start, double end);
 std::string formatCursorPosition(TimeFormat format=TF_RULER, bool useCache=true);
--- a/src/osara.h
+++ b/src/osara.h
@@ -147,7 +147,7 @@
 #define REAPERAPI_WANT_projectconfig_var_getoffs
 #define REAPERAPI_WANT_EnumProjects
 #define REAPERAPI_WANT_GetProjectName
-#define REAPERAPI_WANT_GetProjectTimeOffset REAPERAPI_WANT_GetProjectTimeOffset
+#define REAPERAPI_WANT_GetProjectTimeOffset
 #define REAPERAPI_WANT_plugin_register
 #define REAPERAPI_WANT_GetTrackUIVolPan
 #define REAPERAPI_WANT_GetGlobalAutomationOverride
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -153,9 +153,11 @@ string formatTime(double time, TimeFormat timeFormat, bool isLength,
 			timeFormat = TF_MEASURE;
 		}
 	}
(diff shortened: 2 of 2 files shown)


MatejGolian:
I can report that it works nicely. It's a bit of a bummer though that there's no way of retrieving the project start measure - at least not currently.

nofishonfriday:
Project start measure can be retrieved using SWS' [SNM_GetIntConfigVar()](https://www.extremraym.com/cloud/reascript-doc/#SNM_GetIntConfigVar) with varname = [projmeasoffs](https://mespotin.uber.space/Ultraschall/Reaper_Config_Variables.html#projmeasoffs).

jcsteh:
Thanks @nofishonfriday!

We probably want to do this in a separate PR, but OSARA does make use of one SWS function already (NF_GetSWSTrackNotes), so I don't see any harm in using another. It just needs to fail gracefully if SWS isn't installed. That said, SWS ends up using projectconfig_var_getoffs/projectconfig_var_addr or get_config_var, so we could just use that ourselves and avoid the SWS dependency here. I'm unclear as to when you need projectconfig_var_getoffs vs get_config_var (is projectconfig_var_getoffs project specific and get_config_vra global?).

nofishonfriday:
Hi,

seems you already peeked the relevant SWS code. I already thought about suggesting rolling your own GetIntConfigVar version, if you don't want to depend on SWS. :)
However I didn't write that function, so couldn't give you any details about it, sorry.
[cfillion](https://github.com/cfillion) would probably be the one to ask (same name on the Reaper forum).

That said, from what I observed, Cockos seems to be quite responsive on feature requests regarding accessibility, so I could imagine [this](https://forum.cockos.com/showthread.php?t=259173) being implemented soon(ish).
Just a guess though of course.

MatejGolian:
Hello again,
    I made more tests with this build today and unfortunately I think that this issue needs more work by you clever programming people. I've attached a zipped sample project below. In short, 'odd' BPMs seem to be causing problems.
    The attached project is set to a 6/8 time signature at 102 BPM. The time signature probably has no effect on this, but as I'm currently working on a 6/8 piece, I wanted to see how it all would behave in real life.
    I have  placed a marker where the project should start, but in stead of reporting 0 min 0.000 sec there, on pressing CTRL + Shift + J Osara reports 0 min -0.004 sec. The jump to time/marker/region dialog correctly displays 0:00.000 at that position.
    The start time offset in project settings is -0:03.529, but for some reason instead of that value, Osara reports 0 min -3.533 sec at the very beginning of the file.
    In contrast, at 120 BPM everything works as one'd expect.
    [Start Time Offset Test.zip](https://github.com/jcsteh/osara/files/7461171/Start.Time.Offset.Test.zip)

leonardder:
Thanks for this great test project, I will have a look,

leonardder:
I think I fixed this particular issue in the last commit. I can not reproduce the issue with the test project with that commit applied.

MatejGolian:
@leonardder, don't know how you did it, but confirmed: It all works nicely now. Many thanks. Everything I threw at it gave expected results - at least I haven't found a scenario that would make your code confused. :D

leonardder:
When looking into start offset in measures, I noticed that length reporting is probably broken this way. I think it makes sense to yet bundle them in one project as the area of code is strongly related.

jcsteh:
> When looking into start offset in measures, I noticed that length reporting is probably broken this way. I think it makes sense to yet bundle them in one project as the area of code is strongly related.

Fair enough. If you think it's best to do them in the same PR, that's fine.

You should be able to get the project start measure like this (untested):

```
int size = 0;
int offset = projectconfig_var_getoffs("projmeasoffs", size);
assert(size == sizeof(int));
int measure = *(int*)projectconfig_var_addr(nullptr, index);
```

src/reaper_osara.cpp
```diff
@@ -153,6 +153,12 @@ string formatTime(double time, TimeFormat timeFormat, bool isLength,
 			timeFormat = TF_MEASURE;
 		}
 	}
+	if (includeProjectStartOffset) {
+		if (timeFormat != TF_MEASURE && timeFormat != TF_SAMPLE) {
+			ReaProject* proj = EnumProjects(-1, nullptr, 0);
+			time += GetProjectTimeOffset(proj, false);
```
jcsteh, line 159: For most REAPER functions, you can just pass nullptr to choose the current project. Does that work here?

src/reaper_osara.cpp
```diff
@@ -153,6 +153,9 @@ string formatTime(double time, TimeFormat timeFormat, bool isLength,
 			timeFormat = TF_MEASURE;
 		}
 	}
+	if (!isLength && includeProjectStartOffset && timeFormat != TF_MEASURE && timeFormat != TF_SAMPLE) {
```
leonardder, line 156: I'm not 100% sure about TF_SAMPLE here, but I don't think applying the project start time to samples is a good idea.

Lo-lo78:
Hi Leonard, it works perfectly!

👍

src/osara.h
```diff
@@ -143,8 +143,11 @@
 #define REAPERAPI_WANT_GetMediaTrackInfo_Value
 #define REAPERAPI_WANT_GetTrackMIDINoteName
 #define REAPERAPI_WANT_get_config_var
+#define REAPERAPI_WANT_projectconfig_var_addr
+#define REAPERAPI_WANT_projectconfig_var_getoffs
 #define REAPERAPI_WANT_EnumProjects
 #define REAPERAPI_WANT_GetProjectName
+#define REAPERAPI_WANT_GetProjectTimeOffset REAPERAPI_WANT_GetProjectTimeOffset
```
jcsteh, line 150:
```suggestion
#define REAPERAPI_WANT_GetProjectTimeOffset
```

jcsteh (approved):
Thanks!

PR #586, opened 2021-11-03
Add action feedback for toggle delta solo

### LINKED ISSUE #584: Is it possible to vocalize when activating or deactivating the Delta solo from the action?
Hello,
there is action to activate or deactivate the Delta solo, is it possible for OSARA to send information to the screen reader?
The action is:
FX: Toggle delta solo for last focused FX
Very convenient, as for the bypass it would also be useful for the delta solo.
Greetings!

leonardder:
Fixes #584

Note: nothing needed around fx param ident stuff. Also, as this action is supported in REAPER 6.37 and above, do we need to bump the minimum version?

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -870,8 +870,8 @@ void postToggleAllTracksFxBypass(int command) {
 
 void postToggleLastFocusedFxDeltaSolo(int command) {
 	outputMessage(GetToggleCommandState(command) ?
-		translate("delta solo on") :
-		translate("delta solo off"));
+		translate("enabled delta solo") :
+		translate("disabled delta solo"));
 }
 
 bool shouldReportScrub = true;
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -868,6 +868,12 @@ void postToggleAllTracksFxBypass(int command) {
 		translate("all tracks FX active"));
 }
 
+void postToggleLastFocusedFxDeltaSolo(int command) {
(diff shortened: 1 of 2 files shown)


Lo-lo78:
It's ok Leonard!

👍

src/reaper_osara.cpp
```diff
@@ -868,6 +868,12 @@ void postToggleAllTracksFxBypass(int command) {
 		translate("all tracks FX active"));
 }
 
+void postToggleLastFocusedFxDeltaSolo(int command) {
+	outputMessage(GetToggleCommandState(command) ?
+		translate("delta solo on") :
+		translate("delta solo off"));
```
jcsteh, line 874: Given the length of the words "delta solo", what do you think about "delta on"/"delta off"? Alternatively, "enabled delta solo"/"disabled delta solo"? My aim here is to make sure the users get the important information (the on/off state) as early as possible.

jcsteh (commented):
Thanks!

I don't think we need to bump the minimum version. This won't break anything on older versions; the functionality just can't be triggered. I usually only bump the minimum version when something would break on an older version.

src/reaper_osara.cpp
jcsteh: Or "soloed delta" and "unsoloed delta", but that seems weird.

leonardder: I think Enabled/disabled delta solo is ok. The word solo seems pretty relevant in this context.

jcsteh (approved):

PR #588, opened 2021-11-05
Windows installer: Ask about key map replacement using a message box  instead of showing it as an optional component.

### LINKED ISSUE #579: Windows Installer: can we clarify the UX around key map replacement?
It's been pointed out on RWP that in the Windows installer, it's easy to miss the presence of our "Replace existing key map with OSARA key map" check box. I reckon the below would be a neat solution that would also clarify the choice for users.

On Mac, the terminal script offers a Yes/No choice that has to be answered before installation proceeds. Would it be possible to implement this approach in the Windows installer?

I'm thinking the prompt would go something like:
"Do you want to replace the existing key map with the OSARA key map?
New users are advised to answer Yes, which will place a clean copy of the OSARA key map including all latest assignments.
Answering No will install this snapshot of OSARA without modifying your local key map, which may be preferable for experienced users who have prior alterations that they'd like to preserve."

Any mileage? All thoughts welcome.

jcsteh:
This makes it more obvious to users and should avoid accidental key map overwriting.
The answer defaults to no.
Fixes #579.

### DIFF
--- a/installer/osara.nsi
+++ b/installer/osara.nsi
@@ -1,7 +1,7 @@
 ; OSARA: Open Source Accessibility for the REAPER Application
 ; NSIS installer script
 ; Author: James Teh <jamie@jantrid.net>
-; Copyright 2016 NV Access Limited
+; Copyright 2016-2021 NV Access Limited, James Teh
 ; License: GNU General Public License version 2.0
 
 !include "MUI2.nsh"
@@ -29,7 +29,6 @@ RequestExecutionLevel user
 Page custom portablePage portablePageLeave
 !define MUI_PAGE_CUSTOMFUNCTION_PRE directoryPagePre
 !insertmacro MUI_PAGE_DIRECTORY
-!insertmacro MUI_PAGE_COMPONENTS
 !insertmacro MUI_PAGE_INSTFILES
 
 !insertmacro MUI_UNPAGE_CONFIRM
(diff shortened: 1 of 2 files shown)


jcsteh:
I want you all to know that working on this sucked and I derived no satisfaction from it whatsoever. :)

Someone want to take this for a spin before I merge?

@ScottChesworth , I have this defaulting to no. Are you happy with that? It could result in support requests from users just hitting enter and wondering why they have no OSARA keys. On the other hand, since the whole point of this was to avoid unintentional overwrites, defaulting to yes seems counter-productive. A user can always reinstall and hit yes after actually reading the prompt.

ScottChesworth:
> @ScottChesworth , I have this defaulting to no. Are you happy with that? 
> 
I'd prefer it default to Yes for two reasons:
1. The text suggests that anyone other than more experienced users will probably want to choose Yes. It seems weird to explain that, then offer No by default.
2. I feel quite strongly that the responsibility to make an active choice about which option is most appropriate for this installation should remain with more experienced users, because they'll likely be more familiar with how all the pieces of the puzzle fit together.

> Someone want to take this for a spin before I merge?
> 
Done. It worked. Thanks for pushing through the tedium.

> I want you all to know that working on this sucked and I derived no satisfaction from it whatsoever. :)
> 
If it's any consolation, it's arrived less than 24 hours after I exported what I thought would be the final final very finished version of a new audio walkthrough of installation on Windows. Deep breaths...

jcsteh:
> I'd prefer it default to Yes for two reasons:

Okay. Changed.

> If it's any consolation, it's arrived less than 24 hours after I exported what I thought would be the final final very finished version of a new audio walkthrough of installation on Windows. Deep breaths...

Oh blah. Sorry. :(

Issue #580, opened 2021-10-30
Advanced Peakwatcher Functionality: True Peak, at least, not working


Justinmac84:
Hi,

I'm sure this worked back along, but, when I try to set the Peakwatcher to true peak mode, the meters set to 0.0 and stay there. 

Scott informs me that OSARA no longer adds a plugin for this advanced functionality because Reaper does it natively, but OSARA still is, although, when you highlight a track, it is not reported as being there. it only shows up when you enter the fx chain window and is undeletable, which would make sense if everything was working.

What's more, both Kenny from Reapermania and John from the Reaper Blog still seem to be adding loudness plugins to their fx chain, so I'm wondering if Reaper really does do this stuff natively or if the plugin is still required.

I'm noticing that there are two loudness plugins, the one that OSARA still automatically adds and one with a chattier name including the types of info it will give you. I was wondering if the plugin that OSARA is using was deprecated?

I'm running the latest version of OSARA and Reaper 6.40 and have done experiments with this setup using a project I was working on and a simpler one, just involving a tone generator with the same results.

Oh yes and, when I tried to switch from true peak back to peak in my mullti-track project, Reaper crashed entirely. 

Thanks for any help.

Justinmac84:
Hi,

Hopefully this replies to the right place. 

Spoke to Robbie, who has diagnosed the problem as being that OSARA is currently incompatible with the post 6.32 version of the loudness meter. As a temporary work around, he has provided the old version of the meter for those requiring it to copy in.

ScottChesworth:
@RDMurray, can you give us any nudges on what needs to be tweaked?

jcsteh:
My guess is that Cockos changed the order of parameters or added new ones in the middle. Since OSARA gets the parameters by number, that would break things surely enough. So the parameter numbers in the code will need tweaking again. Blerg.

jcsteh:
I vaguely recall reading something in the REAPER changelog very recently about being able to get parameters by some sort of unique id in some cases now. If that's a thing, that would make this a lot more reliable. However, I think that might only be internal params like bypass, delta, etc.

jcsteh:
Ah, TrackFX_GetParamIdent and TrackFX_GetParamFromIdent.

RDMurray:
They also added a new parameter "Output values as automation" which has to be set before the output parameters will work.  

@jcsteh I am happy to work on this issue if you aren't already.

jcsteh:
[...]

PR #589, opened 2021-11-05
Make peak watcher work with the loudness meter included with Reaper 6.32

RDMurray:
Update parameter numbers for the loudness meter. Fixes: #580


jcsteh:
Thanks!

leonardder:
I think strictly spoken this needs a minimum version bump, right?

jcsteh:
Yes it does. Fixed. Good catch.

Issue #560, opened 2021-07-22
Make remaining untranslatable strings translatable


ElCrackLoko:
Hello everybody. How are they?
In the translation I saw something that says that it can be enabled / disabled so that the automation points move when we move the items, just as if we were talking about one Riple per track but for the automation points. If this is the case, I would like to check it but I would like to know the command since I have not seen it or have not known how to look for it.
I would also like to know where I could send Osara lines that are not yet part of the translation but are obviously part of Osara.
Here are some for example:
"Change Item Selection"
(This, as perhaps many already know, is when we scroll through items with control plus the right or left arrows)
When configuring the peak meter with Alt + W there are also a couple of options like
"none"
Y
"Follow current track"
I take this opportunity to leave another one that I found, in the Midi event editor,
"1 event removed"
As well as in the singular and in the plural, as
"3 events removed"
Maybe they are trifles but I wanted to report it
Thank you very much and sorry for the inconvenience.

jcsteh:
> In the translation I saw something that says that it can be enabled / disabled so that the automation points move when we move the items, just as if we were talking about one Riple per track but for the automation points. If this is the case, I would like to check it but I would like to know the command since I have not seen it or have not known how to look for it.

Could you provide the exact string you're unsure about? That'll make it a lot easier for me to find it.

> I would also like to know where I could send Osara lines that are not yet part of the translation but are obviously part of Osara.

I'm aware there are quite a few (at least 50 strings!) I haven't gotten around to making translatable yet. The process of going through them all and making them translatable is super tedious, mind-numbing and time consuming and I just haven't gotten to the rest of them yet.

> "Change Item Selection"

This one wasn't in my list and should be an easy fix. Thanks for pointing it out.

> When configuring the peak meter with Alt + W there are also a couple of options like
> "none"
> "Follow current track"

These weren't in my list either and should also be pretty easy. Thanks.

> "1 event removed"
> As well as in the singular and in the plural, as
> "3 events removed"

These ones are in my list, but are somewhat trickier to fix.

jcsteh:
If anyone wants to work on the code for this, I explained what needs to be done in https://github.com/jcsteh/osara/issues/573#issuecomment-925427115.

RDMurray:
I've started working on this. Currently going through midiEditorCommands.cpp.

jcsteh:
Amazing. Thank you.

PR #594, opened 2021-11-10
Make strings in midiEditorCommands.cpp translatable (#560)

RDMurray:
I think this covers all of the strings that should be translated in midiEditorCommands.cpp.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -19,6 +19,7 @@
 #endif
 
 using namespace std;
+using namespace fmt::literals;
 
 // Note: while the below struct is called MidiControlChange in line with naming in Reaper,
 // It is also used for other MIDI messages.
@@ -242,11 +243,12 @@ void cmdMidiMoveCursor(Command* command) {
 			notes.push_back({chan, pitch, vel, 0, start, end});
 		}
 	}
-	auto count = notes.size();
+	int count = static_cast<int>(notes.size());
 	if (count > 0) {
 		previewNotes(take, notes);
 		fakeFocus = FOCUS_NOTE;
-		s << " " << count << (count == 1 ? " note" : " notes");
+		s << " " << format(
+			translate_plural("{} note", "{} notes", count), count);
 	}
 	outputMessage(s);
 }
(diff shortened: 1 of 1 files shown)


src/midiEditorCommands.cpp
```diff
@@ -1443,17 +1547,18 @@ void postMidiSwitchCCLane(int command) {
 }
 
 void postToggleMidiInputsAsStepInput(int command) {
-	ostringstream s;
-	s << (GetToggleCommandState2(SectionFromUniqueID(MIDI_EDITOR_SECTION),
-		command) ? "enabled" : "disabled");
-	s << " MIDI inputs as step input";
-	outputMessage(s);
+	bool enabled = GetToggleCommandState2(SectionFromUniqueID(MIDI_EDITOR_SECTION), command);
+	if(enabled){
```
jcsteh, line 1551: Out of curiosity, why the bool variable here but not later?

jcsteh (changes requested):
Thank you very much.

RDMurray:
I don't know how to get the individual reply fields above to work.  

with regard to cmdMidiNoteSplitOrJoin,
Currently those are the only two actions that can call this function.  Maybe in the future there could be others, but we could add them at the time as "transformed" might not make sense anyway.

src/midiEditorCommands.cpp
```diff
@@ -960,7 +976,9 @@ void cmdMidiMoveToTrack(Command* command) {
 	char* trackName = (char*)GetSetMediaTrackInfo(track, "P_NAME", NULL);
 	if (trackName)
 		s << " " << trackName;
-	s << " item " << itemNum << " " << GetTakeName(take);
+	s << " " << format(
```
jcsteh, line 979: This one needs a translators comment.
```suggestion
	// Translators: Used when reporting activation of the next/previous track in
	// the MIDI editor. {num} will be replaced with the item number. {name} will
	// be replaced with its name. For example: "item 2 chorus".
	s << " " << format(
```

jcsteh (approved):
Thanks! I particularly appreciate this given how tedious it is. :)

src/midiEditorCommands.cpp
```diff
@@ -1443,17 +1553,18 @@ void postMidiSwitchCCLane(int command) {
 }
 
 void postToggleMidiInputsAsStepInput(int command) {
-	ostringstream s;
-	s << (GetToggleCommandState2(SectionFromUniqueID(MIDI_EDITOR_SECTION),
-		command) ? "enabled" : "disabled");
-	s << " MIDI inputs as step input";
-	outputMessage(s);
+	bool enabled = GetToggleCommandState2(SectionFromUniqueID(MIDI_EDITOR_SECTION), command);
+	if(enabled){
```
jcsteh, line 1557: Your update seems to have gotten lost here somehow...
```suggestion
	if (GetToggleCommandState2(SectionFromUniqueID(MIDI_EDITOR_SECTION), command)) {
```

PR #595, opened 2021-11-11
Support translation for more strings in reaper_osara.cpp

### LINKED ISSUE #560: Make remaining untranslatable strings translatable

RDMurray:
I think this should be the remaining strings from reaper_osara.cpp 
Fixes #560

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1073,10 +1073,25 @@ void postGoToSpecificMarker(int command) {
 			continue;
 		fakeFocus = reg ? FOCUS_REGION : FOCUS_MARKER;
 		ostringstream s;
-		if (name[0])
-			s << name << (reg ? " region " : " marker ");
-		else
-			s << (reg ? "region " : "marker ") << num << " ";
+		if (name[0]){
+			if(reg){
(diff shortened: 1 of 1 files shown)


RDMurray:
making draft for now because of mistakes.

jcsteh:
Again, thanks so much for bearing the tedium. :)

PR #596, opened 2021-11-12
Play notes in midi event list based on MIDIEditor_GetSetting_str                                                                                                                                             

### LINKED ISSUE #439: Note previews in the event list only work properly if their length is one beat or longer
When previewing notes in the MIDI event list, the previews only play correctly if a note has a length of one beat or longer. If the length is even one tick shorter than a beat, the note is only played for about a quarter of the displayed length. This is not affected by the "ticks per quarter note for new MIDI items" setting in Reaper preferences.

leonardder:
Fixes #439 

Note: I haven't yet tested the test case from #439, marking this as draft for now.

I also recall an issue regarding translated note names, but I can't find one. That should have been fixed now as well.


leonardder:
Ugh, appveyor refuses to build this somehow

jcsteh:
Weird. I just tried redelivering the webhook notification and it didn't trigger then either. Can you try `git commit --amend` and force push to see if it gets picked up as a new commit?

leonardder:
Just did, no luck unfortunately.

jcsteh:
Thanks for this, as always.

I can't get AppVeyor to build this no matter what I do. Not sure what's going wrong.

The translated note names issue is #545. I hacked around that, but this is indeed a proper fix.

We can probably also use this technique to improve cmdFocusNearestMidiEvent, but that can be done separately.

jcsteh (approved):
I see this is still marked as draft, but the code looks fine to me, so approving on the condition that the test case works (and the PR is taken out of draft).

leonardder:
Thanks for approving. I can confirm that #439 does no longer occur.

leonardder:
> Thanks for this, as always.

You're most welcome!

> We can probably also use this technique to improve cmdFocusNearestMidiEvent, but that can be done separately.

I looked into this and I had to move stuff around, so I figured I might as well just do it in here. It's not a big deal.

src/midiEditorCommands.cpp
```diff
@@ -1060,17 +1061,38 @@ void cmdMidiNoteSplitOrJoin(Command* command) {
...
 	double cursorPos = GetCursorPosition();
-	for (int i = 0; i < ListView_GetItemCount(focus); ++i) {
-		char text[50];
-		// Get the text from the position column (1).
-		ListView_GetItemText(focus, i, 1, text, sizeof(text));
-		// Convert this to project time. text is always in measures.beats.
-		double eventPos = parse_timestr_pos(text, 2);
+	HWND editor = MIDIEditor_GetActive();
+	assert(editor == GetParent(focus));
+	for (int i = 0; i < MIDIEditor_GetSetting_int(editor, "list_cnt"); ++i) {
```
jcsteh, line 1081: Please assign the count to a variable before looping. Otherwise, we'll keep asking REAPER for the count for every iteration of the loop, which might be expensive with a lot of events. I do see I made the same error in my original code. :)

jcsteh (changes requested):

jcsteh (approved):
Thanks.

PR #598, opened 2021-11-13
Allow silencing note preview by pressing control

### LINKED ISSUE #597: Worrying keyboard hook behavior
## Reproduce
Below these lines:

```
	HWND focus = GetFocus();
	if (!focus) {
		return CallNextHookEx(NULL, code, wParam, lParam);
	}
```

Add:

```
	ostringstream s;
	s << wParam;
	outputMessage(s);
```

Then, start REAPER. Observe that the if statement at the top (i.e. the one that checks for keys to ignore doesn't seem to be executed when it actually should, e.g. when pressing a key that's not in the if statement.

leonardder:
Fixes #597 

This pr provides a handy way to silence long running note previews. This can be helpful when previewing long notes, such as string sections.

I also tidied up duplicate code to stop a note preview. Furhtermore, the keyboard hook is much more often skipped early in the process.


Lo-lo78:
Leonard, yes, it's ok!

👍

ScottChesworth:
I haven't tested firsthand yet, but that's a nifty solution. Thanks @leonardder!

src/midiEditorCommands.cpp
```diff
@@ -184,13 +183,8 @@ void previewNotes(MediaItem_Take* take, const vector<MidiNote>& notes) {
...
-	if (previewDoneTimer) {
-		// Notes are currently being previewed. Interrupt them.
-		// We want to turn off these notes immediately.
-		KillTimer(NULL, previewDoneTimer);
-		previewDoneTimer = 0;
-		previewNotesOff();
-	}
+	// Stop the current preview.
+	cancelMidiPreviewDoneTimer(true);
 	// Queue note on events for the new notes.
```
jcsteh, line 188: The reason I didn't do this originally is that there's no guarantee concerning the timing or sync/async behaviour of PlayTrackPreview. In the original code, we only call PlayTrackPreview once for both note offs and note ons here. In your new code, PlayTrackPreview will be called twice in quick succession. I worry that might cause problems if PlayTrackPreview is async. I guess you could get around this by having another argument to the cancel function to only queue.

src/midiEditorCommands.cpp
```diff
@@ -212,10 +206,13 @@ void previewNotes(MediaItem_Take* take, const vector<MidiNote>& notes) {
 		(UINT)(minLength ? minLength * 1000 : DEFAULT_PREVIEW_LENGTH), previewDone);
 }
 
-void cancelMidiPreviewNotesOff() {
+void cancelMidiPreviewDoneTimer(bool notesOff) {
```
jcsteh, line 209: This function name seems specific to the timer, but it also queues/sends note offs. Perhaps we could just call it cancelMidiPreview?

src/reaper_osara.cpp
```diff
@@ -2285,17 +2285,19 @@ bool maybeFixTabInSaveDialog(bool previous) {
 // Handle keyboard keys which can't be bound to actions.
 // REAPER's "accelerator" hook isn't enough because it doesn't get called in some windows.
 LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
-	if (code != HC_ACTION && wParam != VK_APPS && wParam != VK_RETURN &&
-			wParam != VK_F6 && wParam != 'B' &&
-			wParam != VK_TAB) {
+	const bool isKeyDown = !(lParam & 0x80000000);\
+	if (!isKeyDown || code != HC_ACTION || (
```
jcsteh, line 2289: Extraneous trailing `\`.

jcsteh, line 2289: 1. On HC_ACTION: Oops. My bad. :) Nice catch.
2. Are you absolutely certain there are no cases left in this code that want a key up? I know there was at least one obscure key up case at one point, but maybe that got eliminated with some refactor (perhaps the context menu refactor).

jcsteh (changes requested):
Neat idea. Thanks for that, as well as the cleanup.

src/midiEditorCommands.h
```diff
@@ -10,9 +10,10 @@
 
 const int MIDI_EDITOR_SECTION = 32060;
 
-// This should be called when playback starts, as otherwise, pending note off
+// This can be called when playback starts, as otherwise, pending note off
```
jcsteh, line 13: I think this comment needs further tweaking. Calling this when playback starts is not optional; it is still a "should" at least, perhaps even a "must", definitely more than a "can". Perhaps it should be described more in terms of what it does and then how it is used.

> Cancel MIDI preview. If notesOff is true, note off messages are sent immediately. This is used to silence the preview. If notesOff is false, the pending note off messages are never sent. This is used when playback starts, as otherwise, pending note off messages for OSARA MIDI preview might interfere with REAPER MIDI playback.

leonardder:
I thought about your comments and figured it would be clearer to separate the timer cancel from sending the notes off, as it was before. Otherwise, the cancelNotePreview function would be really nasty with two arguments that also somehow depend on each other and passing one of the arguments to another function.

src/reaper_osara.cpp
leonardder: > 2. Are you absolutely certain there are no cases left in this code that want a key up? I know there was at least one obscure key up case at one point, but maybe that got eliminated with some refactor (perhaps the context menu refactor).

I checked all the if clauses several times and can't find any case where isKeyDown should be false.

src/midiEditorCommands.cpp
```diff
@@ -142,27 +142,30 @@ PreviewSource previewSource;
...
 		return; // Cancelled.
-	previewNotesOff();
-	previewingNotes.clear();
-	// Send the events.
-	previewReg.curpos = 0.0;
-	PlayTrackPreview(&previewReg);
-	previewDoneTimer = 0;
-	KillTimer(nullptr, event);
+	}
+	if (cancelMidiPreviewDoneTimer()) {
```
jcsteh, line 166: I assume there should never be a case where cancelMidiPreviewDoneTimer returns false here, since we already returned early if the timer was cancelled? I guess there's nothing wrong with checking the result anyway, but thought I'd check.

src/midiEditorCommands.h
```diff
@@ -10,9 +10,16 @@
...
-// This should be called when playback starts, as otherwise, pending note off
+// Stops the notes currently being previewed.
+// If sendNoteOff is true, note off messages are sent immediately.
+// This is used to silence the preview.
+// If sendNoteOff is false, note off messages are queued but not yet sent.
+// This is used when stopping a note preview that is immediately followed by a new preview.
+void previewNotesOff(bool sendNoteOff=true);
+
+// This must be called when playback starts, as otherwise, pending note off
 // messages for OSARA MIDI preview might interfere with MIDI playback.
```
jcsteh, line 21: Please document the return value here, as it isn't immediately obvious what it means.

src/midiEditorCommands.h
```diff
@@ -10,9 +10,16 @@
...
-void cancelMidiPreviewNotesOff();
+bool cancelMidiPreviewDoneTimer();
```
jcsteh, line 22: I'm still not super fond of this function name. It's very implementation specific and code outside of midiEditorCommands shouldn't need to be concerned with that. Do you have concerns with the original name of cancelMidiPreviewNotesOff? Perhaps cancelMidiPreviewPendingNotesOff would work better for you?

jcsteh (changes requested):

src/midiEditorCommands.cpp
leonardder: I wonder whether it could be theoretically possible that the preview is canceled at the same time previewDone is called. In that case, previewDoneTimer would be nullptr when the event occurs.

src/midiEditorCommands.h
leonardder: I think `cancelMidiPreviewNotesOff` is ok.

leonardder: Actually, when calling cancelMidiPreviewNotesOff followed by midiPreviewNotesOff, ain't that weird? how about cancelPendingMidiPreviewNotesOff. I think I prefer that over cancelMidiPreviewPendingNotesOff, as it more resembles the intend that a pending call of midiPreviewNotesOff is canceled.

src/midiEditorCommands.cpp
jcsteh: It's entirely possible for the callback to still run after the timer is cancelled, yes. However, we set previewDoneTimer to nullptr in that case, and the early return above this (`event != previewDoneTimer`) will happen if previewDoneTimer is nullptr (or a subsequently set timer).

jcsteh (approved):
Perfect! Thanks.

Issue #434, opened 2020-12-22
Unable to properly navigate between CCs at the same position


Batworx:
When using the "OSARA: Move to previous CC" and "OSARA: Move to next CC" actions, if multiple CC messages are on the same MIDI tick, the actions don't seem to know which one of them to move to. Could this problem be solved by treating all CC messages at the same position as a chord or group, and having additional actions to move between individual messages in the group in ascending or descending order, much like Osara is already doing with notes?

leonardder:
I looked briefly into this. It looks like CC events are ordered arbitrarily, and the order can even change between invocations of the next/previous cc actions or the event list. I added [an example project](https://github.com/jcsteh/osara/files/5743125/ccIndex.zip)

@jcsteh Do you have any thoughts about this? Should we do our own ordering? I understand that cc indices can change when new events are inserted between orders, but I'd assume that they ought to be stable when you don't change anything. If not, that's something Cockos (not we) should fix.

leonardder:
@jcsteh Just a friendly reminder that there's a question open for you in https://github.com/jcsteh/osara/issues/434#issuecomment-751356276 if you don't mind.

jcsteh:
Does MIDI_Sort do anything useful to help here?

When you say they're ordered arbitrarily, do you mean just CCs at the same tick or all CCs? That is, are they at least ordered by position?

I agree it's pretty weird they shift even when you don't insert or delete. That said, sorting probably isn't that difficult if they're at least sorted by position. We already do that for notes. We could just sort by CC number, which is kinda arbitrary but should serve well enough.

I can't quite remember how the sorting for chords works. I guess we'll have to track the sorted number of the CC at the current position or something.

leonardder:
> Does MIDI_Sort do anything useful to help here?

I'm afraid not. I added it to MoveToCC, just above `auto cc = findCC(take, direction);`, but that doesn't make any difference.

> When you say they're ordered arbitrarily, do you mean just CCs at the same tick or all CCs? 

All cc's at the same tick are ordered arbitrarily. The order can change by 
That is, are they at least ordered by position?

> I agree it's pretty weird they shift even when you don't insert or delete. That said, sorting probably isn't that difficult if they're at least sorted by position. We already do that for notes. We could just sort by CC number, which is kinda arbitrary but should serve well enough.

That makes sense, I will see how to do this.

leonardder:
[...]

Issue #393, opened 2020-11-25
A technical question about the parameter view


davidkreynolds:
Hi,
How does OSARA process data for the parameter list as invoked by pressing P?
When using CSI to control plugins, they use parameter numbers. It would be so convenient if you both used the same methodology to create the parameter list.
Not really an issue, but I'll close it if you can give me some kind of answer.
Thanks.

jcsteh:
OSARA queries all parameters in the order they're exposed via the REAPER API. I guess you could say that means it uses "parameter numbers" internally, though I guess that depends if CSI has its own concept of parameter numbers or whether it uses the REAPER API numbers unmodified.

Note that the numbers shown after the parameter name in OSARA's dialog have 1 added to them for user friendliness, since REAPER parameter numbers are 0 based. That is, if OSARA says (1), the internal parameter number is actually 0.

PR #599, opened 2021-11-13
Let edit cursor follow focus in MIDI event list

leonardder:
It has always been a thorn in the eye for me that changing focus in the event list doesn't move the edit cursor, since you're basically moving focus in the event list to edit things, right? I added an option to allow that.
The new "Edit cursor follows focus in MIDI event list" option is disabled by default, as it modifies REAPER's default behavior in such a way that it could be potentially annoying/confusing for someone.

This pr is a draft until #596 is merged, as it depends on it and. Therefore a rebase is necessary after merging.


leonardder:
I didn't realize that this does not work on Mac. I now hid the option on Mac, which will add a gap in the dialog. I think this shouldn't be a big problem.

ScottChesworth:
Is there any chance of getting this to fly on Mac too? Ideally, I'd prefer to avoid features being platform-specific. The ability to jump platforms and keep workflows in tact is unique to OSARA in this space, I'd like to preserve that wherever possible.

leonardder:
We do not support note preview on event list focus either. I'm afraid 
this is simply not possible unless Cockos gives us a way to be modified 
about focus changes in the REAPER interface.

ScottChesworth:
> We do not support note preview on event list focus either.
> 
Ah, somehow I didn't know that. Well if we're already down the rabbit hole and you've already done the work, then I retract my gripe. :)

Lo-lo78:
👍👍👍

jcsteh:
Does this force the edit cursor to go to the start of the MIDI item when you open the event list? If I remember correctly, the first event gets focused in that case. That wouldn't be ideal.

Even if it doesn't, if the user presses down arrow without first pressing control+f to focus the nearest event to the cursor, they'll get thrown to the start of the item then. That said, I don't know that there's a good solution to this. I wondered about auto focusing the nearest event when opening the event list, but I think that would be difficult to manage without getting into nasty conflict situations.

leonardder:
> Does this force the edit cursor to go to the start of the MIDI item when you open the event list? If I remember correctly, the first event gets focused in that case. That wouldn't be ideal.

Ah, that's a good catch ineed.

> I wondered about auto focusing the nearest event when opening the event list, but I think that would be difficult to manage without getting into nasty conflict situations.

Could you elaborate on what conflict situations might arise?

jcsteh:
I think my thoughts about conflicts related to the fact that if we incorrectly focus the nearest event for a focus other than the initial focus, we could end up in a situation where we focus the nearest event, that triggers a focus event, we respond to that by focusing the nearest event. However, I now realise it'd stop there rather than getting into a loop because the focus wouldn't actually change in the final part. Still, that would be pretty nasty regardless.

leonardder:
As far as I can see, on initial focus, no actual item in the list gets focus, so that case should be easy to detect.

Lo-lo78:
Hi Leonard, I just tried your latest build and with midi files with many 
events in one item there is a latency. I think it depends on the number 
of events in a midi item; if I create smaller items there is no latency.

Greetings!

jcsteh:
Are you seeing latency when using the cursor keys or only when first opening the event list? I can understand latency when opening (since focusing the nearest event has to search many events), but I'd be surprised by latency when using the cursor keys.

jcsteh:
If the latency is caused by focusing the nearest event, we should be able to improve that using a binary search. I implemented binary search for the piano roll in #391. I never landed it because no one ever confirmed that it helped them in real usage. However, with the framework I implemented there, it should be possible to create a template to do this for the event list.

Lo-lo78:
Yes, there is latency when I open the event list, when it ends it can be 
put into play, which causes an interruption for the play as well.

ScottChesworth:
@Lo-lo78, can you link to a MIDI file we can use for consistency during testing? Many events will mean different things to different users.

Lo-lo78:
Here is the link:

https://drive.google.com/uc?id=11sVbS5wxNbzJqKmZqIClh17UN_GVDBs1&export=download

ScottChesworth:
Thanks. Now, can you test to see if the event list is still responding slowly loading this data with OSARA pre621 from the snapshots page?
Here, the event list is laggy irrespective of whether I'm using pre621 or the most recent AppVeyor build in this thread. On my computer, @leonardder's contribution doesn't make the situation noticeably better or worse. Are you seeing the same thing?

leonardder:
I did a force push because it looks like appveyor is hanging once again.

leonardder:
> If the latency is caused by focusing the nearest event, we should be able to improve that using a binary search. I implemented binary search for the piano roll in #391. I never landed it because no one ever confirmed that it helped them in real usage. However, with the framework I implemented there, it should be possible to create a template to do this for the event list.

I think the binary search can also help with fixing #434 right? Anything against merging that and let it prove itself?

ScottChesworth:
> Anything against merging that and let it prove itself?
> 
Just tried to download the windows build in #393 to test whether it improves latency with the MIDI file in this thread. It's been more than a month, so AppVeyor doesn't have the build anymore. How do we rebuild that?
BTW, sorry I missed it first time round @jcsteh.

leonardder:
Think you meant #391? I merged latest master into it. However, note that binary search is only implemented for notes/chords and not for events.

Lo-lo78:
Hi, now with the screen reader there is no latency but if I switch from 
the piano roll to the event list the play is slowed down the first time 
always with large items. With smaller items there is no problem.

jcsteh:
> I think the binary search can also help with fixing #434 right? Anything against merging that and let it prove itself?

As you realised, the binary search isn't currently implemented for CCs. However, the framework should make implementing that a lot easier. I've merged it now because it seems to help @Lo-lo78 and so we can base other improvements on that framework.

leonardder:
@jcsteh is there anything holding this back currently?

jcsteh:
As I understand it, there are still performance concerns as reported in https://github.com/jcsteh/osara/pull/599#issuecomment-970673460. Fixing this would require chanhing to a binary search here. If you did that already, I didn't see any commit notifications here...

leonardder:
@Lo-lo78 could you please try the try build that will be posted shortly after this message to see how performance is going compared to the build of two days ago?

leonardder:
Ah, it failed on mac. Never mind, here is the windows build to test with: https://ci.appveyor.com/api/buildjobs/s0gusns4l1v4fjsf/artifacts/installer%2Fosara_pr599-758%2C3508b6cd.exe

Lo-lo78:
@leonardder 
Great, the piano roll motor is very snappy and the first time it is perfect it synchronizes.
However, I do not think it is intended, without exiting the midi editor, if I reopen the piano roll and I move and re-enter the event list, I go back to the previous point.
For instance:
Piano Roll: event 307
Event List: event 307
Piano Roll: event 106
Event list: event 307.

If I do it for the second time from the piano roll and I go to the event list it doesn't sync.
I have to move the event list cursor to update the position otherwise it returns to event 307.
;))
Am I wrong something?

leonardder:
Could you provide the midi file to reproduce?

Lo-lo78:
Hi, try this midi file.
[Diana Est - Tenax3.zip](https://github.com/jcsteh/osara/files/7906022/Diana.Est.-.Tenax3.zip)

leonardder:
I think I covered all edge cases now and indeed, the binary search is really fast and snappy!

@Lo-lo78 I think I fixed your case as well, would be great if you could verify. I also made sure that no movement takes place when the current focus is at the cursor position. The binary search iterator together with equal_range makes thas pretty elegant if I may say so.

Lo-lo78:
Very good!

Congratulations, you have improved the situation a lot!

My opinion is that you have solved, look forward to the opinions of others.

For me it's perfect!

👍👍👍

src/reaper_osara.rc
```diff
@@ -63,8 +63,11 @@ BEGIN
 	CONTROL "Report &FX when moving to tracks/takes", ID_CONFIG_REPORT_FX, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 10, 126, 140, 14
 	CONTROL "Report MIDI &notes in MIDI editor", ID_CONFIG_REPORT_NOTES, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 10, 146, 140, 14
 	CONTROL "Report changes made via &control surfaces", ID_CONFIG_REPORT_SURFACE_CHANGES, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 10, 166, 155, 14
-	DEFPUSHBUTTON "OK", IDOK, 10, 186, 30, 14
-	PUSHBUTTON "Cancel", IDCANCEL, 137, 186, 40, 14
+#ifdef _WIN32
```
jcsteh, line 66: Does the PHP script which generates SWELL resources respect ifdefs? In other words, do we know for certain that this actually excludes the control on Mac?

jcsteh (commented):
This is looking great.

src/reaper_osara.rc
leonardder: AH that's actually pretty unlikely looking at the script. I wonder how to fix this.

jcsteh: I think you'll need to do this at runtime, after the dialog is created but before it is shown. If Mac, get the HWND for the checkbox and hide it. You should be able to use ShowWindow with the right arguments to hide it, though I don't recall the specifics. You could test that on Windows easily enough. Failing that, you could just disable it with EnableWindow(hwnd, false) which I know for a fact works because I do that elsewhere.

leonardder:
I first started with EnableWindow(.., FALSE), then realizing that it still shows the window text. ShowWindow works much better indeed.

jcsteh (approved):
Thanks so much for working on this.

jcsteh:
> The binary search iterator together with equal_range makes thas pretty elegant if I may say so.

Agreed. When I wrote the original version of the binary search code, I spent a fair bit of time on it and it then languished for ages without feedback. I'm glad it has proven to be useful and grateful that you've taken it to the next level. :)

leonardder:
@RDMurray is this mac or Windows?

leonardder:
Ugh you are correct. Pretty sure this didn't work before. Anyhow, I guess we can revert this.

Lo-lo78:
Bye!
Yes. It works even without OSARA.
Too bad that Reaper has not completely resolved the latency between the transition from the piano roll to the event list. With large items that have large lists, Reaper has some problems.
the Actions List is Reaper's only quick list. When you add a plugin there is latency. Maybe they'll fix these big lists too. ;))
👍

PR #601, opened 2021-11-15
Add action to mute the next message from OSARA

### LINKED ISSUE #600: Is it possible not to make the screen reader speak when it is in play for actions that have OSARA voice messages?
Hello!
The messages that OSARA sends to the screen reader are very useful, but when I'm in play and I adjust the time selection, I prefer to listen to the audio content without hearing the screen reader telling me the position all the time.
Perfect when we are not in play but some actions with OSARA messages could interfere in the work when we have an audio stream below that we would like to hear.
I don't know how to proceed, I could make some suggestions:
1.
we could put a checkbox in the OSARA configuration:
Disable OSARA messages for time selection in play. Check box disabled(default)  
2.
It might be interesting to create an OSARA action for custom actions or cycle actions; this action should only change the message for that custom action or cycle action event. Alone this OSARA action is useless but combined with other actions it could be very useful to customize a custom action or a cycle action.

For instance.
Custom Action:
OSARA: Mute OSARA message for this event when in play
Time selection: Nudge left edge left

In this way the action to adjust the left edge of the time selection will not have the message and we can hear better what we are doing in play.

Greetings!

RDMurray:
For use in custom actions.  Suppresses OSARA output for the next command handled by OSARA.  See comments in issue #600.
Fixes #600.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -694,6 +694,9 @@ OSARA also includes some other miscellaneous actions.
 #### MIDI Event List Editor
 - OSARA: Focus event nearest edit cursor: control+f
 
+### Muting OSARA messages in custom/cycle actions
+The action "OSARA: Mute next message from OSARA" can be used in custom/cycle actions to mute OSARA feedback for the next action. It should be placed before each action to be muted.
+
 ## Support
 If you need help, please subscribe to the [Reapers Without Peepers discussion group](https://groups.io/g/rwp) and ask your questions there.
 
--- a/readme.md
+++ b/readme.md
@@ -694,8 +694,9 @@ OSARA also includes some other miscellaneous actions.
 #### MIDI Event List Editor
 - OSARA: Focus event nearest edit cursor: control+f
 
-### Muting OSARA messages in custom/cycle actions
-The action "OSARA: Mute next message from OSARA" can be used in custom/cycle actions to mute OSARA feedback for the next action. It should be placed before each action to be muted.
+### Muting OSARA Messages in Custom/Cycle Actions
+The action "OSARA: Mute next message from OSARA" can be used in custom/cycle actions to mute OSARA feedback for the next action.
+It should be placed before each action to be muted.
 
 ## Support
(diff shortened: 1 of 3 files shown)


Lo-lo78:
Hi RDMurray, yes, it's very good, great! I've tried it with both custom 
actions and cycle actions. It always works! 👍

Lo-lo78:
Yes, it works very well!

👍

Lo-lo78:
Here is an example:


Cycle Action; Increase and decrease the left border and of an item, 
increase and decrease the right border of an item

https://drive.google.com/uc?id=1zPCmGV3AODmESrIxQ39IM1Mvtgp25_Eb&export=download

Cycle Action; Time selection; works on the edges of a selection without 
feedback in play

https://drive.google.com/uc?id=16VvG3AtTFkz4wBci_tGpxuAUASN90j5d&export=download


Greetings!

Lo-lo78:
I find it very clean and efficient; a user who does not know Reaper well 
cannot go wrong!

jcsteh (changes requested):
This is perfect. Thank you. It needs to be documented in the readme, though. Perhaps you could add a tiny sub-section at the end of the "Usage" section (after Miscellaneous Actions) which explains how this is used. That sub-section could be called "Muting OSARA Messages in Custom/Cycle Actions" or something like that.

readme.md
```diff
@@ -694,6 +694,9 @@ OSARA also includes some other miscellaneous actions.
 #### MIDI Event List Editor
 - OSARA: Focus event nearest edit cursor: control+f
 
+### Muting OSARA messages in custom/cycle actions
+The action "OSARA: Mute next message from OSARA" can be used in custom/cycle actions to mute OSARA feedback for the next action. It should be placed before each action to be muted.
```
jcsteh, line 698: nit: Split sentences onto separate lines.
```suggestion
The action "OSARA: Mute next message from OSARA" can be used in custom/cycle actions to mute OSARA feedback for the next action.
It should be placed before each action to be muted.
```

jcsteh (approved):
Great! Thanks!

PR #602, opened 2021-11-17
Fix capitalisation error in cmdMidiNoteSplitOrJoin

RDMurray:
oops! :)


RDMurray:
@jcsteh  do you mind me merging really small fixes like this?

jcsteh:
Thanks. Sure, but please note the following for future reference:

1. In general, any merges to OSARA should be squash merges. I make very rare exceptions if a branch contains commits where it's useful to see the individual pieces, but this would never be the case for a fix like this.
2. Where possible, please have the commit title specify the user visible effect of a change. It's okay to include technical detail about the function you're fixing, etc. in the commit message, but I want users to be able to look at the commit log to understand what's new, and this commit message might not make any sense to some users.

RDMurray:
Yes, sorry about that.  I should have asked first.  It does make the log look a bit messy.

PR #604, opened 2021-11-18
Improve binary search iterator

leonardder:
Merely as a personal studying case, I tried to expand on the binary search algorithm idea #391. The idea is to pass a lambda to the constructor of the iterator that creates a MidiNote or MidiControlChange, fetching only those properties from the REAPER API that we actually need, such as the start position for the note. However, in case of note preview (e.g. in `cmdMidiMoveCursor`), we need to know more.

I looked at [std::ranges::equal_range](https://en.cppreference.com/w/cpp/algorithm/ranges/equal_range) in C++20, which should allow us to do something like:

```
auto range = ranges::equal_range(begin, end, now, ranges::less{}, &MidiNote:;start);
```

That looks pretty awesome, the downside is that it does not work.

I at least did the following:

1. Create an iterator for control changes (currently not used)
2. Implemented ReqParams struct to provide to the iterator constructor so it will fetch the appropriate properties from REAPER when dereferencing.
3. Cache the current value on the iterator when dereferencing, clearing it when moving it around. As std::optional is used, this should be lazy.


leonardder:
Ugh, this pr is getting a bit messy, sorry for that. It is really a bit of a play ground.

I'm trying to abstract things a bit, also to make it easier to get midi note or event data from REAPER without passing all these nullptrs, doing manually conversion from ppq pos to project pos, etc.

Furthermore, I discovered that the ranges namespace is not yet available for Clang, which is a pitty.

leonardder:
Ugh Apple Clang is really no fun. It doesn't support passing iterators to the constructor of vector.

leonardder:
@jcsteh I think this code can help with #599. Could you have a look at the proposal please?

src/midiEditorCommands.cpp
```diff
@@ -24,14 +25,51 @@ using namespace fmt::literals;
...
 	int index;
 	int message1;
 	int message2;
 	int message3;
 	double position;
-} MidiControlChange;
+	bool selected;
+	bool muted;
+
+	struct GetParams {
```
jcsteh, line 38: I'm not a big fan of this GetParams thing for two reasons:

First, even though you're only using it when constructing the iterator, it will be available inside instances of the outer struct (e.g. MidiControlChange). Worse, the values in the instance of the outer struct will actually be wrong. I think this problem could be solved by moving this struct out and having the MidiEventIterator template take two arguments: the event type (e.g. MidiControlChange) and the requested params (e.g. MidiControlChangeReqParams).

Second, incantations like `MidiNoteIterator begin(take, MidiNote::GetParams{true, true, true, true, true})` aren't super readable. I guess we could comment the parameter names before each boolean (`/* position */ true`), but I wonder whether bit vectors would be easier and more readable:

```
enum class MidiControlChangeReqParams {
position = 1 << 0,
message = 1 << 1,
};
```

Then you could do:

`MidiNoteIterator begin(take, MidiControlChangeReqParams::position | MidiControlChangeReqParams::message1)`

MidiEventIterator's constructor would then just take an int with the bit vector. The only disadvantage there is that it's less type-safe - you could pass in CC flags for a note iterator - but I'm not sure how much of a real problem that is and the readability advantage might justify that sacrifice.

If you don't like the bit vector idea, I'm happy to go with the structs, but please move them out of the outer struct and comment the param names as described in my first point.

src/midiEditorCommands.cpp
```diff
@@ -436,11 +522,11 @@ pair<int, int> findChord(MediaItem_Take* take, int direction) {
 		return {-1, -1};
 	}
 	// Find the last note of the chord.
-	double firstStart = *firstNote;
+	double firstStart = (*firstNote).start;
```
jcsteh, line 525: You should be able to do `firstNote->start` instead of `(*firstNote).start`. There are a few other instances of this too.

jcsteh (changes requested):
I did look at this a while ago, but I was somewhat confused by it. Looking at it again, it makes a bit more sense now. :)

Thanks for working on this. I think it will inded make some things a bit easier.

This is still failing to build on Mac, which is obviously a problem. Some of the errors look easily fixable. One of them looks a bit more complicated. It seems std::vector takes an iterator, but it isn't satisfying `!__is_cpp17_forward_iterator` for some reason. I didn't look into this much, but I did find [this StackOverflow thread](https://stackoverflow.com/questions/57596150/why-does-clang-not-like-boosttransform-iterator) which might be of interest.

Beyond that, here are some thoughts.

jcsteh:
Um... I think my comments about the inner struct might be entirely bogus. Just ignore that until I have a chance to think about this with a less silly brain.

src/midiEditorCommands.cpp
jcsteh: > even though you're only using it when constructing the iterator, it will be available inside instances of the outer struct (e.g. MidiControlChange).

I'm so sorry. I'm completely wrong. For some reason, I was thinking this was defining the struct as a data member of the outer struct, but it isn't; it's just declaring a type, not a data member. This is totally fine. 😳

src/midiEditorCommands.cpp
```diff
@@ -468,19 +554,15 @@ bool compareNotesByPitch(const MidiNote& note1, const MidiNote& note2) {
 // This updates curNoteInChord.
 MidiNote findNoteInChord(MediaItem_Take* take, int direction) {
 	auto chord = findChord(take, 0);
-	if (chord.first == -1)
+	if (chord.first == -1) {
 		return {-1};
+	}
+	MidiNoteIterator begin(take, MidiNote::GetParams{true, true, true, true, true});
```
jcsteh, line 560: 1. I still think the readability here is a concern. Please comment the param names; e.g. `/* position */ true`.
2. Will the compiler deduce the type if you remove the `MidiNote::GetParams` here, so it's just: `MidiNoteIterator begin(take, {...})`? It might not, but if it does, let's do that.

leonardder:
Well, I do like your enum class suggestion. I think it makes sense to go 
that route.

src/midiEditorCommands.cpp
leonardder: So I did this, but that requires the -> operator to be implemented on the iterator. I did this now, but that's not going to work with temporary objects I'm afraid. Unless there's a workaround I do not know of.

leonardder:
Never mind, I followed the approach in https://stackoverflow.com/questions/4050429/stl-iterator-dereferencing-iterator-to-a-temporary-is-it-possible
let's see how this works now.

leonardder:
@jcsteh Could you please give some quick hints as how to debug OSARA in a debugger? I guess I could consider making a visual studio project?

jcsteh:
I've never used Visual Studio as an IDE. :) I use WinDBG for dbugging.
WinDBG commands are super unintuitive, though.

You can install WinDBG by "modifying" the Windows 10 SDK installation in
Apps and Features or Add/Remove Programs. You want to install Debugging
Tools for Windows.

Once you start REAPER, open up WinDBG. Attach to the REAPER process by
pressing f6 and choosing the right process.

I can't remember what the default GUI settings are these days, but the
command window might be floating. Enabling MDI emulation from the Window
menu might fix that. Failing that, hunt around with object nav until you
find it.

Anyway, once you have the command window open, type "g" (and press enter;
it's a command line) to resume the REAPER process. Then, repro your crash.

Once it crashes, you can use the command "kp" to get a stack trace.

Other useful commands are: ".frame nn" to switch to a particular stack
frame nn, "dv" to show variables on the stack, "?? expr" to evaluate a C++
expression expr (useful for inspecting variables).

leonardder:
Thanks a lot.

I wonder whether a switch to cmake might be beneficial here, as it can 
automatically build visual studio projects. Debugging in Visual Studio 
allows you to set break points, which is of course a bit easier than 
looking at the stack trace of a crash. There's also pretty nice VS code 
integration. Having said that, it's probably harder to get mac and 
windows differences right.

jcsteh:
WinDBG can set break points too. You can either do it with the bp command
or I believe you can use the source window and set one from there, though
I've never actually done it with the source window. You can also add
conditional breakpoints (e.g. break in a function if a certain argument has
a certain value).

I know cmake is a lot more popular than scons, but I'm pretty reluctant to
consider a switch. It's basically a whole other language to learn.
Especially with some of the weird stuff OSARA has to do (building 32 and 64
bit on Windows, Win/Mac differences, etc.), it wouldn't be simple. I
realise this is probably largely me being backwards and resistant to
change, but... 🤷‍♂️

leonardder:
Interesting enough, as soon as I repro the crash in REAPER, everything 
seems to freeze, including NVDa. Hitting alt+f4 for several times will 
bring NVDA back but that of course means that WinDBG is closed.

jcsteh:
Welcome to the joy of debugging as a screenreader user. I'm surprised you've not hit this before with VS debugging, since the principles are the same, just with a different UI.

In reality, the debugger isn't frozen, though it does of course suspend REAPER. The problem is that suspending a process sometimes takes a screen reader with it depending on what query the screen reader was making at the time. NVDA's watchdog should prevent this, but as you no doubt know, it doesn't catch everything, especially UIA stuff.

Anyway, alt+tabbing a few times might help to try to trigger the watchdog. Or you might have to restart NVDA.

leonardder:
I found the cause of the crash among several other issues:

1. Operator+ and operator- were wrong, they were behaving like operator+= and operator-=, mutating the current iterator instead of returning a new one.
2. Also, using std::optional to persist the value for operator-> wasn't working either. I now used the simplest approach advertised in https://stackoverflow.com/questions/4050429/stl-iterator-dereferencing-iterator-to-a-temporary-is-it-possible
3. I was initializing a vector from iterators of a chord, not realizing that findChord returns inclusive values. Therefore every end iterator for chord.second should be made exclusive with an additional + 1.

src/midiEditorCommands.cpp
```diff
@@ -257,15 +339,19 @@ class MidiEventIterator {
...
+		return this->getEvent(this->index + index);
 	}
 
-	const double operator*() const {
+	reference operator*() const {
 		return (*this)[0];
 	}
 
+	pointer operator->() const {
+		this->currentValue = (*this)[0];
```
jcsteh, line 351: Ug. I'm kinda regretting suggesting you use `->` now. I didn't realise it'd be this painful or ugly. I honestly thought it'd be the same as `(*foo).bar` at the end of the day, but apparently not. Anyway, it's done now and I think it probably does make things more readable in future.

src/midiEditorCommands.cpp
```diff
@@ -404,7 +498,6 @@ const string getMidiNoteName(MediaItem_Take *take, int pitch, int channel) {
 	}
 	return s.str();
 }
-
 // Returns the indexes of the first and last notes in a chord in a given direction.
 pair<int, int> findChord(MediaItem_Take* take, int direction) {
```
jcsteh, line 502: I'm wondering whether this should eventually return a pair of MidiNoteIterators. I guess that'd require passing in ReqParams in case the caller wants something else. And there's the challenge of what to return instead of -1. We can deal with that some other time, though.

src/midiEditorCommands.cpp
```diff
@@ -539,6 +634,15 @@ vector<MidiNote> getSelectedNotes(MediaItem_Take* take, int offset=-1) {
 	return notes;
 }
 
+using MidiControlChangeIterator = MidiEventIterator<MidiControlChange>;
+
+template<>
+int MidiControlChangeIterator::getCount() const {
```
jcsteh, line 640: I assume you're planning on using this pretty soon? if not, I'd suggest removing from this PR and adding this in future.

src/midiEditorCommands.cpp
```diff
@@ -42,10 +80,53 @@ struct MidiNote {
...
+			note.start = start;
+		}
+		if (params.end) {
+			end = MIDI_GetProjTimeFromPPQPos(take, end);
+			note.end = end;
+		}
+		return note;
 	}
+
+
```
jcsteh, line 129: nit: Extraneous blank lines.

src/midiEditorCommands.cpp
```diff
@@ -223,22 +304,23 @@ bool cancelPendingMidiPreviewNotesOff() {
 }
 
 // A random access iterator for MIDI events.
-// The template parameter is arbitrary and is only used to specialise the
+// The template parameter is used to specialise the
 // template for different event types (note, CC, etc.). We do this instead of
 // subclassing because iterators need to return their exact type, not a base
```
jcsteh, line 309: nit: Remove the final sentence of this comment about subclassing because now that this iterator returns different types (not always double), subclassing could never work in any way, shape or form.

jcsteh (approved):
Thanks for persevering with this. Approved with nits fixed.

src/midiEditorCommands.cpp
leonardder: It is needed for #599, so yes.

src/midiEditorCommands.cpp
leonardder: I thought about that too when having lunch today, haha. Point is that for the chords, we just need the start for the note, whereas for preview, we need much more parameters from REAPER (vel, chan, etc.)

Having said that, I wonder whether REAPER's `MIDI_GetNote` performance is affected by requesting more parameters. Most likely it is just getting a note and writing values that are already in memory to the provided output variables. That doesn't apply to start and end, though.

leonardder:
Ugh, I decided to test drive this once more when editing a midi project and discovered that selection is broken. Let's fix that first.

src/midiEditorCommands.cpp
jcsteh: Sure, but the function could take ReqParams and we could pass in different ReqParams depending on whether we need preview.

PR #605, opened 2021-11-19
Update appveyor builds to newest tooling

leonardder:
Updates to Visual Studio 2022 / XCode 13.0. The latter requires us to use the big sur image.
This finally allows us to pinpoint C++ support to C++20 for both environments. Very welcome because yesterday, I bumped into several case where code didn't compile on Mac. I'm not sure whether C++20 is fully covered though.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,7 @@ branches:
 environment:
  matrix:
   - job_name: Windows
-    appveyor_build_worker_image: Visual Studio 2019
+    appveyor_build_worker_image: Visual Studio 2022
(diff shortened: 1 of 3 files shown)


leonardder:
Not sure why this is still a draft, since I think it's ready 😉

jcsteh:
Would the change to the build flags mean this won't run on Visual Studio 2019 or Xcode 12? I'm happy enough to update AppVeyor, but I don't want to require VS 2022 just yet, since it was only just released.

leonardder:
XCode 13 is really required for these flags whereas VS2022 isn't.

PR #607, opened 2021-11-20
Ensure we don't ignore keys in the keyboard hook we shouldn't ignore

### LINKED ISSUE #606: Alt+DownArrow fails to open presets combo box with OSARA pre-638 and Reaper 6.42
This got reported on RWP. I can reproduce here.

1. Instantiate an effect that provides presets in stock library format, let's say ReaEQ.
2. In the FX window, navigate to the Presets combo box.
3. Hit Alt+DownArrow. Expected behaviour is that the combo box pops open and OSARA's search filter is revealed. Actual behaviour here is that neither of those things happen. Instead, the Alt+DownArrow keystroke still decreases track volume as if I were in the TCP.

Any idea what's up?

leonardder:
Fixes #606

When refactoring the keyboard hook in #598 , there was one thing I forgot about, namely checking whether the initial if clause that ignored keys we weren't interested in was complete. That wasn't the case, therefore shift+f10 and alt+downArrow didn't work. It should be fixed with this pr.

The keys are now listed in the order in they are used later on in the code

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -2294,9 +2294,10 @@ bool maybeFixTabInSaveDialog(bool previous) {
 LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
 	const bool isKeyDown = !(lParam & 0x80000000);
 	if (!isKeyDown || code != HC_ACTION || (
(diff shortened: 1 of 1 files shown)


ScottChesworth:
Working wonderfully here. Thanks @leonardder

PR #611, opened 2021-12-04
Shorten FX names reported when navigating tracks

### LINKED ISSUE #608: Refining what OSARA reports when effects are present (this needs input from translators)
In the comments of issue #592, I proposed a shorter reporting format when users move to a track or item that has effects instantiated. This would involve splitting the string we get from Reaper based on punctuation and spacing. The goal is to cut back on extraneous speech and relay all useful information to the user in a shorter time.
If you're a translator, or you use Reaper and OSARA in a language other than English, please examine the format of the two reports below and let us know the following things:
1. Are the spacing and punctuation I've used here consistent in your translation?
2. Would the second, shorter report still make sense in your language?
3. Can you think of any examples of effects that aren't covered here, IE, where the spacing and/or formatting of how they're reported differs?

Here's our current report when moving to a track:
1 Guitar Riff 1 item; FX: VST: ReaEQ (Cockos), VST3: Ultimate Serious Pro (Waves), JS: Soft Clipper/Limiter

Here's the shorter, more productive version I'm proposing. Note that this example includes notification that the second effect in the chain is bypassed, which is what we're hoping will be implemented in #592.
1 Guitar Riff 1 item; FX: ReaEQ, Ultimate Serious Pro bypassed, Soft Clipper/Limiter

Thanks in advance for any feedback.

jcsteh:
Strip the type prefix (VST, JS, etc.) and strip the vendor suffix in parentheses.
One problem with this is that not all JS effects have a vendor suffix, but some use a parenthesised suffix for other reasons; e.g.
JS: Transient-Driven Auto-Pan (Transmitter)
With this PR, we strip (Transmitter) from that last one, and given that there's an accompanying (Receiver) plugin, that is not ideal.
Still, this is a rare edge case, and it may be worth sacrificing it for the overall benefit this brings.
Fixes #608.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -768,7 +768,22 @@ void postGoToTrack(int command, MediaTrack* track) {
 			if (f > 0)
 				s << ", ";
 			TrackFX_GetFXName(track, f, name, sizeof(name));
-			s << name;
+			const regex RE_FX_NAME("^([A-Z]+): (.+?)( \\(.*?\\))?$");
+			cmatch m;
+			regex_search(name, m, RE_FX_NAME);
+			if (m.empty()) {
+				s << name;
+			} else {
+				// Group 1 is the prefix, group 2 is the FX name, group 3 is the
+				// parenthesised suffix.
+				s << m.str(2);
+				if (m.str(1) == "JS") {
+					// For JS, not all effects have a vendor name. Therefore, we always
+					// include the parenthesised suffix to avoid stripping potentially
+					// useful info.
+					s << m.str(3);
+				}
+			}
 			if (!TrackFX_GetEnabled(track, f)) {
 				s << " " << translate("bypassed");
 			}


ScottChesworth:
Thanks for going ahead with this, I'm pretty sure it's gonna be a welcome change.

I wonder if perhaps js should be excluded from the stripping? It's only just occurred to me that consistency is nonexistent with many third-party js FX. VST, VST3 and their I counterparts seem to be more widely used anyway, so the middle ground would still benefit most folk without the loss of useful information.

Seems I didn't examine enough js FX, sorry about that!

jcsteh:
Disabling stripping for JS is a little trickier to implement, but I'll see
what I can do.

ptorpey:
Just playing around with the ReaLibrary and I wonder if this update affected how some of the ReaLibrary plug-ins are spoken by Osara.

For example, I loaded a Rhodes piano from the ReaLibrary onto a track.  When I arrow to the track, Osara reports:
FX: Track 3 "ReaLibrary - 08 - Electric Pianos - Fender Rhodes - Mello Rhodes EP" dialog 

So it is very clear what the loaded instrument is.

However, when I open the FX dialog and arrow to the Rhodes instrument, Osara reports:
enabled VSTi: BassMidiVsti (Falcosoft)

which doesn't clearly say what the loaded instrument is.

I do notice that the more complete name can be found in the FX dialog when navigating by object in the dialog, so maybe this is just how the list view of plug-ins works with long names and a screen reader.

--pete

ScottChesworth:
That's all normal. BassMIDI is what's being used to host the EP patch.

jcsteh:
That's a surprisingly accurate effect name. I wonder if the FX instance was
renamed? I've not tried that, but I guess it probably impacts what OSARA
sees.

In any case, OSARA doesn't touch the name reported in the FX list view.

PR #614, opened 2021-12-06
Go to first and last track actions

ScottChesworth:
I might have screwed up rebasing this on the current master. Thought I'd done it right, but I'm seeing a warning here that this can't be merged automatically. Any thoughts?
Also gonna mark this as draft for now, because I still haven't got the Mac key map updated. Jenny K says she'll make the changes, but first I need to learn how to make AppVeyor build this sucker (I don't have a local Mac build environment at the mo). Any advice very, very welcome!


jcsteh:
Nice work! The conflict is because I've modified osara.h since you did and they've thus diverged. You'd need to either git merge master or git rebase master, but either way, you'd have to manually resolve the conflict. If you want to try your hand at that, [this documentation might help](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#_basic_merge_conflicts). Otherwise, I can take care of this easily.

That said, I'm wondering whether this is the appropriate behaviour where the first or last track isn't visible; e.g. because it is hidden or (in the case of the last track) because it's inside a collapsed folder. I'm honestly not sure; first and last intuitively suggest first and last regardless of anything else, but on the other hand, you deliberately can't navigate to hidden tracks with up/down arrows, so it feels weird that you could navigate to them this other way. If you do want to skip invisible tracks, that's going to be quite a bit trickier than the code you have here.

jcsteh:
I just realised the custom actions don't handle hidden tracks either, so perhaps we should deal with that question at some other point. That is, this isn't a regression and it does fix other problems.

ScottChesworth:
Think I've merged properly now. Also learned my question about AppVeyor was redundant, it just builds automatically. I'll pass the Mac build along to Jen and have her update the key map asap.

ScottChesworth:
> If you do want to skip invisible tracks, that's going to be quite a bit trickier than the code you have here.
> 
I'd love to say challenge accepted, but I'm not == to it yet. @RDMurray handheld me through most of this. He's been very patient, like a softer-spoken, Scottish Yoda.

ScottChesworth:
Seems like this is ready. Over to you @jcsteh

Issue #617, opened 2021-12-14
Gain Reduction meter for Dynamic Plugins


Nils31011992:
It would be great to have a Kind of Gainreduction Meter for Compressors and other dynamic tools, especially for less experienced users or people who can not see the Meters in Plugins.

jcsteh:
I don't see how this would be possible for OSARA. In order for OSARA to do that, the plugin would need to expose that information programmatically. The only way to do that as far as I know is via a parameter. And if it exposed a parameter, you'd be able to see it in OSARA's Parameters dialog.

Nils31011992:
you can get a gain reduction value in db from the api, so it would work at least for vst3 plugins I think

RDMurray:
Where did you find this in the API?  I couldn't find anything about gain reduction.

A way it could be done is by inserting a custom jsfx before and after the effect to be measured, which could expose a parameter that could be read by OSARA, similar to how the peak watcher works.  

I do think something like this would be useful.  Did you have an idea for how the UX would work for this?

Nils31011992:
Here,

https://forums.cockos.com/showthread.php?t=232577

Von meinem iPhone gesendet

RDMurray:
This is only documented in the changelog:
+ VST: support TrackFX_GetNamedConfigParm() with GainReduction_dB to retrieve the current GR from plug-ins that support the Presonus GR API 


I wonder how widely supported that API is?  

We could make an action to report the current gain reduction for the focused effect using that API fairly easily.  The problem as always with new actions is where to put them on the keymap?

jcsteh:
Ah, I knew about that API, but I incorrectly thought it was only for Cockos effects which I believe already expose gain reduction in the GUI in an accessible way. It we indeed be interesting to know how many other effects support that.

ScottChesworth:
The forum thread that Nils linked to indicates it's supported by Waves VST3. Beyond that, I don't know. How much work would be involved in whipping up that action to start testing in situ Robbie? I think that's probably the only way to find out how useful it is. Doing the gruntwork first isn't something I'd usually consider, but given how useful gain reduction on demand would be, maybe this is an exception.

jcsteh:
We could use the key hook to hard-code a shortcut which only works in the FX chain dialog, since this can only work for a focused effect anyway. The only problem with that is that it would be Windows only.

ScottChesworth:
> The only problem with that is that it would be Windows only.
> 
[...]

PR #618, opened 2021-12-15
add action to report gain reduction for plugins that support it

RDMurray:
Add the action "OSARA: Report current gain reduction for focused effect if supported".

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -685,6 +685,8 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Toggle global automation override between latch preview and off: control+alt+shift+l
 - OSARA: Cycle through midi recording modes of selected tracks: alt+shift+\
 - OSARA: Report groups for current track
+- OSARA: Report current gain reduction for focused effect if supported
+ - If no effect is focused, reports the gain reduction of the last focused effect.
 - OSARA: About
 
 #### Midi editor
--- a/src/osara.h
+++ b/src/osara.h
@@ -194,6 +194,8 @@
 #define REAPERAPI_WANT_GetSetProjectInfo_String
 #define REAPERAPI_WANT_SetOnlyTrackSelected
 #define REAPERAPI_WANT_MIDI_GetEvt
+#define REAPERAPI_WANT_TrackFX_GetNamedConfigParm
+#define REAPERAPI_WANT_TakeFX_GetNamedConfigParm
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
 
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3697,6 +3697,57 @@ void cmdMuteNextMessage(Command* command){
 	muteNextMessage = true;
 }
 
+void cmdReportGainReductionFocusedFX(Command* command){
+	auto invalid = []() {
+		outputMessage(translate("No gain reduction value available"));
+	};
+	static int typeCache, trackNumCache, itemNumCache, fxCache;
+	int trackNum, itemNum, fx;
+	bool valid = false;
+	int type = GetFocusedFX(&trackNum, &itemNum, &fx);
+	if (type == 0 && typeCache != 0) { // no fx focused, so use last focused FX
+	type = typeCache;
+	trackNum = trackNumCache;
+	itemNum = itemNumCache;
+	fx = fxCache;
+	}
+	MediaTrack* track = trackNum == 0 ?
+		GetMasterTrack(nullptr) : GetTrack(nullptr, trackNum - 1);
+	if (!track) {
+		invalid();
+		return;
+	}
+	const int buffsize = 10;
+	char buff[buffsize];
+	if (type == 1) { // Track
+		valid = TrackFX_GetNamedConfigParm(track, fx, "GainReduction_dB", buff, buffsize);
+	} else if (type == 2) { // Item
+		MediaItem* item = GetTrackMediaItem(track, itemNum);
+		if (!item) {
+			invalid();
+			return;
+		}
+		int takeNum = HIWORD(fx);
+		fx = LOWORD(fx);
+		MediaItem_Take* take = GetTake(item, takeNum);
+		if (!take) {
+			invalid();
+			return;
+		}
+		valid = TakeFX_GetNamedConfigParm(take, fx, "GainReduction_dB", buff, buffsize);
+	}
+	if (!valid) {
+		invalid();
+		return;
+	}
+	typeCache = type;
+	trackNumCache = trackNum;
+	itemNumCache = itemNum;
+	fxCache = fx;
+	double gr = stod(buff);
+	outputMessage(format("{:.1f}", gr));
+}
+
 // See the Configuration section of the code below.
 void cmdConfig(Command* command);
 
@@ -3838,6 +3889,7 @@ Command COMMANDS[] = {
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: About"}, "OSARA_ABOUT", cmdAbout},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Report groups for current track"}, "OSARA_REPORTTRACKGROUPS", cmdReportTrackGroups},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Mute next message from OSARA"}, "OSARA_MUTENEXTMESSAGE", cmdMuteNextMessage},
+	{ MAIN_SECTION, {DEFACCEL, "OSARA: Report current gain reduction for focused effect if supported"}, "OSARA_REPORTGR", cmdReportGainReductionFocusedFX},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Go to first track"}, "OSARA_GOTOFIRSTTRACK", cmdGoToFirstTrack},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Go to last track"}, "OSARA_GOTOLASTTRACK", cmdGoToLastTrack},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, "OSARA: Enable noncontiguous selection/toggle selection of current chord/note"}, "OSARA_MIDITOGGLESEL", cmdMidiToggleSelection},


RDMurray:
fixes: #617

ScottChesworth:
I've just taken this for a spin, tried to get some gentle gain reduction going without hearing the source audio (partly as a test of whether it can emulate the mainstream workflow of watching the needle move, mostly because it was the only way I could think of to rely on the GR readings. I tried 3 different pieces of source material, a good spread of dynamic ranges. Unfortunately, in all cases the results were a hot mess of over-compression, which leads me to believe that even banging away on the key with screen reader rate cranked, I was probably missing important data. Gonna get some eyes on that test later today to see what I'm missing, have them watch the meters, tell me what and where the max gain reduction was happening. My initial suspicion is that speech on demand just isn't a good method for reporting information as dynamic and fast-changing as this.
Testing it in combination with Delta being active mode was better. There, I felt like it was adding some insight. Trouble is, in that scenario I could hear what the compressor was doing and that's inevitably going to influence my next tweak, so I still don't know if I can ascertain smooth compression settings just based on the readings we're providing via speech. Rightly or wrongly, I guarantee that's what users are going to try to do.
So right now, my hunch is that for this to be useful, it would ideally need to be an option within Peak Watcher where the maximum gain reduction could either be held for a given time or reported when a desired threshold is reached. I'm aware that presents some UX questions that will need to be thoroughly thought through before we start considering implementation.
More later, once I've got some comparisons and more time under my belt.
And thanks for whipping this up so fast @RDMurray!

jcsteh:
> my hunch is that for this to be useful, it would ideally need to be an option within Peak Watcher where the maximum gain reduction could either be held for a given time or reported when a desired threshold is reached.

That makes sense. I think I'm going to end up needing to rewrite Peak Watcher from the ground up or at least give it a very serious makeover. We just can't support use cases like this or multiple tabs with the current code. Also, I didn't really design it with the kind of complexity it now has to deal with, so the code is becoming pretty ugly. That rewrite/refactor is going to be a massive chunk of work, though, with lots of GUI mangling.

ScottChesworth:
> I think I'm going to end up needing to rewrite Peak Watcher from the ground up or at least give it a very serious makeover. 
> 
Urgh. If there's anything I can do to help with that from here (accountability partner to make sure you show up and bang away at it at a regular time etc), just shout. Also, seems like they're pretty close to releasing this year's Christmas fundraiser, so there'll be some funding available to cover those hours.

jcsteh:
Thanks. It's mostly just that I don't often have the big blocks of
concentrated time that I need for this kind of work these days, and when I
do, I end up working on actually writing/recording music. Which is an
improvement for me personally since I previously spent years working on
OSARA and rarely actually using it, but not so good for the OSARA
community, haha.

Still, I wonder whether it might be easier for me to just rewrite the code
from scratch rather than trying to massage it. PeakWatcher is only just
over 500 lines of code. How hard can it be? <gulp>

RDMurray:
Closing this PR as it isn't useful in it's current state.

jcsteh:
Thinking about doing this in Peak Watcher:

1. Even if we had an FX combo box that was hidden until you chose gain reduction, "follow current track" doesn't make any sense, so we'd have to remove that option when gain reduction was chosen.
2. Would you want to support take FX as well as track FX? That would complicate the UX a lot. The user would then have four things to choose - a track, an item, a take and an effect - and whenever an earlier choice changed, we'd have to update all the later ones. And following current take doesn't make any sense here either.
3. I could see navigating a list of items being really unwieldy. First, there can be many items. Second, unless the user is really good at naming their items, they're going to have no idea what item they're choosing, since they can't hear it when choosing from a list.
4. One possibility would be to use menus and sub-menus instead of combo boxes. That way, the flow is linear: if a user chooses an item and realises they wanted another track, they have to back out of the take and item sub-menus to choose a different track. This also means we can generate the menus on the fly based on the user's selection, rather than having to continually update multiple combo boxes. However, this might not be intuitive - we'd have to have a button or something to bring up the menu - and that would also mean a couple more keystrokes.

jcsteh:
Actually, now I'm starting to wonder about a completely alternative UX for Peak Watcher where the user chooses the track/take by moving to it in the track view, then brings up the Peak Watcher dialog to enable stuff for that track/take. That avoids the issue of choosing or multiple tabs. Though that doesn't solve the problem of a user wanting to monitor two things for the same track. Arrrgh! Oh, and nor does it allow the user to choose an effect.

ScottChesworth:
Does this redesign get any easier if we think of Peak Watcher 2.0 as a menu hierarchy instead of a dialog with multiple tabs? That way, it wouldn't have to be super dynamic, options that are unavailable can just be reported as unavailable. Of course, it'd still need careful planning to keep it intuitive, but would that lighten the load implementation-wise?

RDMurray:
How about adding it to context menus? for example:
* Select track -> context menu -> OSARA meters -> meter 1 -> dialog pops up with meter options for the current track
* focus effect in FX chain window -> context menu -> OSARA meters -> meter 2 -> dialog pops up with metering options relevant to the effect such as Gain reduction.

RDMurray:
> Actually, now I'm starting to wonder about a completely alternative UX for Peak Watcher where the user chooses the track/take by moving to it in the track view, then brings up the Peak Watcher dialog to enable stuff for that track/take. That avoids the issue of choosing or multiple tabs. Though that doesn't solve the problem of a user wanting to monitor two things for the same track. Arrrgh! Oh, and nor does it allow the user to choose an effect.

I think this would work.  Monitoring multiple things could be handled by asking the user to choose channel 1 or channel 2 before showing the dialog.  Effects could be selected simply by focusing them in the FX chain.  
To handle follow current track there could be a checkbox in the dialog when focus is on a track.  I think it could also be useful to have follow current fx available for gain reduction.

Lo-lo78:
@RDMurray it's a very interesting idea; if you I couldkeep the settings for each type of selection:
track, take, that would be great.
For example, when we enter the FX chain window Reaper always remembers the selected plugin. If your system allowed settings to be retained even if there were 200 tracks, there would be 200 settings per track and more per take.
Uau, if you do you are a genius!
👍

Lo-lo78:
@RDMurray You could put this data into Reaper's .rpp project like the DAW does for the FX chain.
What do you think, is it a utopia?

jcsteh:
> Does this redesign get any easier if we think of Peak Watcher 2.0 as a menu hierarchy instead of a dialog with multiple tabs?

Yeah, that's what I was suggesting when I mentioned menus and sub-menus.

> How about adding it to context menus?

I think we can add stuff to the track context menu, but I don't think we can to the FX context menu.

> Monitoring multiple things could be handled by asking the user to choose channel 1 or channel 2 before showing the dialog.

I think that could be confusing because the user might think that refers to stereo channels, and even if we use a different name, it might be construed to mean that each track can have two watchers, rather than two watchers total.

> I think it could also be useful to have follow current fx available for gain reduction.

Again, that can't be done in a cross-platform way.

jcsteh:
> I think it could also be useful to have follow current fx available for gain reduction.

Actually, maybe it can be done cross-platform since Peak Watcher uses a timer, but only while the FX actually has focus. As soon as you close the FX chain dialog, you'd lose it.

jcsteh:
> @RDMurray it's a very interesting idea; if you I couldkeep the settings for each type of selection: track, take, that would be great. ... If your system allowed settings to be retained even if there were 200 tracks, there would be 200 settings per track and more per take.

I think this is getting way too complicated and unrealistic. We don't have enough spots on the key map to on demand query hundreds of peak watchers. Furthermore, the potential announcement flooding that would boggles the mind.

> @RDMurray You could put this data into Reaper's .rpp project like the DAW does for the FX chain. What do you think, is it a utopia?

I think this is out of scope for this work.

Lo-lo78:
@jcsteh In your opinion the experience with:
JS: EBUR128 Loudness Measurement V2.11 (TBProAudio) [osara / ebur128_analysis]
Could this be a cue for the FX Chain Window?
When we analyzed LUFS with this JS we could work in TCP without problems while working.
Did I say nonsense?
Sorry if it's mega galactic bullshit!
(rofl)

jcsteh:
Even though we do add an FX in this case, that isn't related to keeping track of other focused FX that we didn't add ourselves. So no, it's not related to this work.

jcsteh:
Some more precise thoughts on a new UX.

We could have the user begin configuring Peak Watcher in one of two ways:

1. They focus the track or effect they want to monitor, then press alt+w. A context menu would appear where they must specify whether they're configuring the first watched value or the second. Advantage: Keeps the existing alt+w command. Disadvantage: Choosing which watched value might be unintuitive?
2. They focus the track or effect they want to monitor. They then double press the reset action for the watcher they want to configure. Advantage: Makes it clearer what they're configuring and reduces keystrokes. Disadvantage: unfamiliar for existing users, who are used to alt+w.

We could potentially implement both, but I don't love having multiple entry points; that in itself creates confusion.

A disadvantage of both these approaches is that to watch any track, even the master, you have to focus it first. Personally, I think this is reasonable, but I know a bunch of people wanted master specific actions in the past because focusing the master track first is apparently a big deal for them.

Anyway, after that initial entry point, the dialog would be similar to how it is now, except without the option to choose a first or second track. There would be a follow current track/FX checkbox, though.

Thoughts? Anything I've missed?

jcsteh:
The other risk with doing this on a double press of Peak Watcher reset is that some users might be used to tapping that in quick succession for whatever reason. If that is true, binding an action to a double press will be bad for them.

RDMurray:
> I Think this proposal would be perfect @jcsteh  
> 
If people use the double tap shortcut by mistake, they could just cancel the dialog.

ScottChesworth:
> focus the track or effect they want to monitor, then press alt+w.
A context menu would appear where they must specify whether they're configuring the first watched value or the second. Advantage: Keeps the existing alt+w command. Disadvantage: Choosing which watched value might be unintuitive?
> 
I'd prefer this approach to avoid upheaval of existing resources. A double-press of reset doesn't feel any more intuitive here.

> I know a bunch of people wanted master specific actions in the past because focusing the master track first is apparently a big deal for them.
> 
I'm yet to see a strong argument for why master-specific actions have to be preserved. Since last time that topic flared up, we've added a keystroke to move straight to master. Conclusion: they can whistle. :)

jcsteh:
Sitting with it a bit more, I still have two concerns with my proposed UX:

1. Because you focus the target first and then get presented with the menu to configure the first or second watched value, users might think this refers to being able to watch two values for every target, not two values globally.
2. Users wouldn't have a way to check what is already being watched, since they no longer choose that from the dialog and the choice is based on prior focus.

One way to solve both of these would be to include the currently watched value in the name of the context menu item; e.g.
1: First watched value (current: track 1 piano peak dB)
That seems pretty verbose and maybe a bit ugly, though.

Thoughts?

jcsteh:
FYI, I've started on this rewrite, but it's going to be a long slog. I've spent several hours on it already and I'm still nowhere near done. Hopefully, it does actually work out...

Issue #479, opened 2021-02-22
The following error occurs when building osara with Visual Studio Community 2019 version 16.8.5:


dingpengyu:
hi all
The following error occurs when building osara with Visual Studio Community 2019 version 16.8.5:
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
cl /Fobuild\x86\controlSurface.obj /c src\controlSurface.cpp /TP /EHsc /std:c++l
atest /W3 /WX /nologo /Ibuild\x86 /Isrc /Iinclude /Iinclude\WDL /Iinclude\tinyge
ttext\include /Iinclude\fmt\include /Z7
controlSurface.cpp
cl /Fobuild\x86\dictionary.obj /c include\tinygettext\src\dictionary.cpp /TP /EH
sc /std:c++latest /W3 /WX /nologo /Ibuild\x86 /Isrc /Iinclude /Iinclude\WDL /Iin
clude\tinygettext\include /Iinclude\fmt\include /Z7
dictionary.cpp
cl /Fobuild\x86\envelopeCommands.obj /c src\envelopeCommands.cpp /TP /EHsc /std:
c++latest /W3 /WX /nologo /Ibuild\x86 /Isrc /Iinclude /Iinclude\WDL /Iinclude\ti
nygettext\include /Iinclude\fmt\include /Z7
envelopeCommands.cpp
E:\Source\Repos\reaper\osara\include\fmt\include\fmt/format.h(1): error C2220:
以下警告被视为错误
E:\Source\Repos\reaper\osara\include\fmt\include\fmt/format.h(1): warning C4819:
 该文件包含不能在当前代码页(936)中表示的字符。请将该文件保存为 Unicode 格式以防
止数据丢失
scons: *** [build\x86\envelopeCommands.obj] Error 2
scons: building terminated because of errors.
Compiler Environment:
Lenovo Saver ISK15 Notebook
windows10 20h2 64-bit
Visual Studio Community 2019 16.8.5
python3.7.9+virtualenv Latest Stable Version
scons Latest stable version
ps:
The compilation is done with the latest master partition.
And the repository was re-cloned.
thanks

jcsteh:
Does this error go away if you modify line 36 of src/archBuild_sconscript to include /wd4819 after /WX? Like this:
`	env.Append(CXXFLAGS="/EHsc /std:c++latest /W3 /WX /wd4819")`

dingpengyu:
[...]

PR #636, opened 2022-01-08
Build OSARA with LLVM (Clang/LLD) on Windows

leonardder:
This starts building OSARA with LLVM. This should at least bring C++20 coverage on Windows closer to that on Mac, though still not fully as pointed out in https://github.com/jcsteh/osara/pull/636#issuecomment-1008027322

### DIFF
--- a/src/archBuild_sconscript
+++ b/src/archBuild_sconscript
@@ -33,11 +33,26 @@ sources = [
 ]
 
 if env["PLATFORM"] == "win32":
-	env.Append(CXXFLAGS="/EHsc /std:c++20 /W3 /WX "
-		# #479: fmt/format.h includes UTF-8 characters in comments. However, cl
-		# treats the file as ANSI unless it has a BOM. This causes a warning (and
-		# thus an error) on systems with certain locales. Just disable this warning.
-		"/wd4819")
+	# On Windows, OSARA is build with LLVM to have a toolchain that's closer to what's used on Mac
+	env["CC"] = "clang-cl"
+	env["LINK"] = "lld-link"
+	env.Append(CCFLAGS=[
+		"/W3",
+		"/WX",
+		# Clang warns about unused functions in WDL
+		"/clang:-Wno-unused-function",
+	])
+	env.Append(CXXFLAGS=[
+		"/clang:-std=c++20",
+		"/EHsc",
+		# Ignore suggestions to add braces around initialization of subobjects. We might want to fix this at some point
+		"/clang:-Wno-missing-braces",
+	])
+	env.Append(CFLAGS=[
+		# Ignore warnings in WDL/win32_utf8.c, not our code
+		"/clang:-Wno-deprecated-declarations",
+		"/clang:-Wno-unused-variable"
+	])
 	# We always want debug symbols.
 	env.Append(PDB="${TARGET}.pdb")
 	# having symbols usually turns this off, but we have no need for unused symbols.
--- a/readme.md
+++ b/readme.md
@@ -756,8 +756,10 @@ To build OSARA, you will need:
 	* [Download Build Tools for Visual Studio 2022](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022)
 	* Visual Studio 2022 Community/Professional/Enterprise is also supported.
 	* Whether installing Build Tools or Visual Studio, you must enable the following:
-		- On the Workloads tab, in the Windows group: Desktop development with C++
-		- On the Workloads tab, in the Optional group: C++ ATL for latest v143 build tools (x86 & x64)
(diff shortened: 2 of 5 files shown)


jcsteh:
Good idea. What are your thoughts on just doing this always, rather than manually enabling it? I can't see any reason against using clang across the board. We should also use lld-link, since it's significantly faster (though that probably doesn't matter so much for a code base this small).

leonardder:
I actually hoped you'd propose the former. Let's see how that works.

jcsteh:
Regarding your note in the readme, I suspect some of the issues we're seeing with clang aren't so much it being less compliant, but more due to clang being stricter about certain parts of the spec. Also, in my experience, clang's errors and warnings tend to be a hell of a lot more readable. Regardless, having a consistent compiler should help alleviate unnecessary pain for us.

jcsteh:
Of course, this won't solve problems with MacOS not supporting certain parts of later C++ standards on earlier versions of MacOS (e.g. std::get and std::filesystem); those problems will continue to haunt us. :(

jcsteh:
Finally (sorry for the comment spam), when we use clang-cl, we're still using different C++ standard libraries across Windows and Mac, so thinking about it more, I'm not sure this will solve as much as we think it will. My previous comment about MacOS is a prime example of that. Still, the less variables, the better.

leonardder:
I removed the note from the readme. As you said, the less variables, the better. In #604, using clang on Windows at least catches one issue with spec compliance.

leonardder:
Note @ScottChesworth and @RDMurray, merging this will probably require you guys to install the Clang component in Visual Studio. Are you both comfortable with this change?

leonardder:
I'm also curious what do do with the workaround in #479. @xingkong0113 it would be greatly appreciated if you could attempt building this branch.

jcsteh:
> I'm also curious what do do with the workaround in #479.

Yeah, I was wondering about that. I think we have one or two other places where we disable specific warnings/errors too.

ScottChesworth:
> Note @ScottChesworth, merging this will probably require you to install the Clang component in Visual Studio. Are you comfortable with this change?
> 
So long as you're comfortable with the potential increase of dumb questions while I'm doing that, it's a yes from me. Whatever makes things easier.

RDMurray:
> 
> ﻿
> Note @ScottChesworth and @RDMurray, merging this will probably require you guys to install the Clang component in Visual Studio. Are you both comfortable with this change?
> 

No problem.

leonardder:
Ugh, sorry for the spam but we finally got there.
Windows builds are now around 20 seconds slower, but I guess that's something we can live with.

jcsteh:
Wow. I'm really surprised by the speed decrease. Mozilla builds got a lot faster when we switched to clang. That kinda sucks.

src/archBuild_sconscript
```diff
@@ -33,11 +33,21 @@ sources = [
...
+	env["CC"] = "clang-cl"
+	env["LINK"] = "lld-link"
+	env.Append(CXXFLAGS=[
+		"/clang:-std=c++20",
+		"/EHsc",
+		"/W3",
+		"/WX",
+		# Ignore some warnings (treated as errors)
+		# Clang warns about unused functions in WDL/wdltypes.h
+		"/clang:-Wno-unused-function",
```
jcsteh, line 46: I don't love that we're disabling this for the entire project due to WDL. I think we might want to restrict this by doing this with a pragma at include time in future, but we can deal with that later.

src/archBuild_sconscript
```diff
@@ -33,11 +33,21 @@ sources = [
...
+		# Ignore suggestions to add braces around initialization of subobjects. We might want to fix this at some point
+		"/clang:-Wno-missing-braces",
+	])
+	env.Append(CPPDEFINES=["_CRT_SECURE_NO_WARNINGS", "_CRT_NONSTDC_NO_WARNINGS"])
```
jcsteh, line 50: Hmm. Why didn't we need this with MSVC? Any idea?

leonardder: I've looked into it but I have no idea. I wonder whether msvc automatically choses the secure equivalents, in which case switching to Clang would make this less secure.

leonardder:
I found out why clang whined about deprecation warnings while msvc didn't. Actually, we only used CXXFLAGS, which is not used for c code, therefore W3 and WX weren't used for the c code. Therefore, MSVC didn't complain about these warnings. It looks like the default warning level of Clang is different.
I now used CFLAGS (C), CXXFLAGS (C++) and CCFLags (both C and C++).

leonardder:
Re #479, according to the clang-cl docs, it handles all source files as utf-8, so I guess this doesn't apply any longer.

jcsteh (approved):
Thanks!

PR #637, opened 2022-01-08
When moving to a note in the MIDI Editor, report its velocity after its length if View -> Piano roll notes -> Show velocity numbers on notes is enabled.

### LINKED ISSUE #626: Can We Have Note Velocities Reported In The Piano Roll?
Hello,
I think it would be nice if we could get note velocities reported in the piano roll when moving between notes. Right now the only ways of finding out what velocity a given note is at is either opening the event dialog or pressing the V and Shift + V shortcuts. The V/Shift + V method is quick, but it is kinda annoying. Having this feature would make drum programming more comfortable.
Perhaps the most intuitive approach would be appending the velocity after the note length and separating these 2 pieces of information with a semicolon - so, similarly to how notes get reported in the event list editor.
Any thoughts?

jcsteh:
Fixes #626.
@ScottChesworth, @MatejGolian, I'd like some feedback here as to whether just reporting the value after a comma is intuitive enough:
d sharp 4, 25%, 96
The alternative would be to scrap the comma and report "velocity" instead:
d sharp 4, 25% velocity 96
Or we could keep the comma, and/or we could shorten velocity to "vel".
Thoughts? Trying to balance efficiency with intuitiveness here.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -741,6 +741,10 @@ void moveToNoteInChord(int direction, bool clearSelection=true, bool select=true
 	}
 	if (shouldReportNotes) {
 		s << formatNoteLength(note.start, note.end);
+		if (GetToggleCommandState2(SectionFromUniqueID(MIDI_EDITOR_SECTION), 40632)
+				) { // View: Show velocity numbers on notes
+			s << ", " << note.velocity << " " << translate("velocity");
+		}
 	}
 	outputMessage(s);
 }


MatejGolian:
@jcsteh, thanks for this, it works like a treat so far. I vote for keeping the comma and shortening velocity to 'vel'. If it wasn't me who requested this feature, I feel that without it I wouldn't understand what those numbers at the end mean. And should someone turn on that option by mistake, at least they will know that they have to look for something related to velocity. The word 'velocity' feels too much, but I think that leaving that comma in there is useful as it keeps things a bit more separated and more intelligible.

ScottChesworth:
Any chance of moving the word velocity after the number? That way the report would be super clear for newbies and the dynamic info would be spoken sooner for more confident users. seems like the best of both.

MatejGolian:
@ScottChesworth and @jcsteh, if having the word velocity after the numeric value feels good to you both as native English speakers, naturally, I have no objections.

jcsteh:
It's not natural English at all, but I'll generally take efficiency over
natural English as long as it doesn't confuse users. I guess we'll find out.

PR #640, opened 2022-01-10
Build OSARA with multiple concurrent jobs

leonardder:
SCons allows you to specify the number of concurrent jobs used to build. In my case, I measured build times with PowerShell measure-command. The build time decreased from 66 to 18 seconds, definitily a boost isn't it?

### DIFF
--- a/sconstruct
+++ b/sconstruct
@@ -6,6 +6,7 @@
 
 import os
 from makePot import makePot
+import multiprocessing
 
 vars = Variables()
 vars.Add("version", "The version of this build", "unknown")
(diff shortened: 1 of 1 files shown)


jcsteh:
Thanks! I was literally thinking about this a couple of days ago. I was going to hard-code -j2 in appveyor.yml (since AppVeyor documents 2 cores), but this approach is much nicer and gives benefits locally too. (Locally, I had hard-coded -j16). It does seem to speed up AppVeyor builds a bit, though not as much as I'd hoped.

PR #648, opened 2022-01-15
Do not assume the default path for the PHP interpreter, poll for it

vick08:
Do not assume the default path for the PHP interpreter but poll for it. Starting with Mac OS 12.x users will have to install 'PHP', e.g. using Homebrew, in order to compile OSARA.

### DIFF
--- a/src/archBuild_sconscript
+++ b/src/archBuild_sconscript
@@ -98,8 +98,10 @@ else: # Mac
 	env["SHLIBPREFIX"] = ""
 
 	# Generate Mac resources.
+	import shutil
+	php = shutil.which("php")
 	res = env.Command(["reaper_osara.rc_mac_dlg", "reaper_osara.rc_mac_menu"], "reaper_osara.rc",
-		[["include/WDL/WDL/swell/swell_resgen.php", "$SOURCE"]])
+		[[php, "include/WDL/WDL/swell/swell_resgen.php", "$SOURCE"]])
 	env.Depends("reaper_osara.cpp", res)
 
 def addExternalSources(dir, files):


jcsteh:
It amazes and amuses me that in a project this small (relatively speaking), we can have two changes hours apart that conflict... in the build system! :) The WDL update earlier tonight broke your change. I've rebased it now.

Is there any chance you might be able to submit some updated readme guidance for Mac dependencies concerning Homebrew; links to install, how to install php, etc.? I know next to nothing about Mac and I don't have a usable Mac right now to verify anything myself.

jcsteh:
Someone (rather aggressively) noted that Python might be easier with Homebrew on Mac too, rather than installing the official Python distro as recommended in the readme. Is that true?

tbdalgaard:
If the certificates are installed via brew too I would go that route anytime. Brew is a little hard to grasp if you nothing about the terminal at all, but I guess most folks will prefer that.
I am on Catalina, would you mind me testing the tips in the readme out and make a new issue about changes?
I am not a programmer at all, so if I can get it right 99 % of the users who understand programming can do this too, grin..

Best regards Thomas

ScottChesworth:
If you decide to start building from source there, definitely report back on where the readme worked well for you and where it could've been more helpful. I think we're in a spot where the Windows guidance is decent now, but I'm less sure about the Mac side.

jcsteh:
I'm happy with brew if that's easiest. The readme just needs to be updated to include correct guidance. It doesn't need to be a detailed primer on how to install brew - surely brew has it's own documentation we can link to - but links would be good, along with the correct commands to install php and python with brew.

vick08:
Yeah, let me write up something quick for the Readme unless soemone else gets to it first!

vick08:
> Someone (rather aggressively) noted that Python might be easier with Homebrew on Mac too, rather than installing the official Python distro as recommended in the readme. Is that true?

I can't speak on the level of aggression referred to here, <grin>, but I've been always installing Python using Homebrew. For me, it is simply a matter of consistency, i.e. installing into the same directory structure. I know where Homebrew puts its things and I know where to find them if I need to!

jcsteh:
Using brew for both Python and PHP seems reasonable, then. It certainly means less messing around, as the install process for both dependencies will be the same.

Issue #574, opened 2021-09-20
Suggestion/Request: Automatically Select MIDI CCs Under Selected Notes


MatejGolian:
Hi there,
I think it would be nice if underlying CC messages got automatically selected when moving to and selecting notes in the piano roll. Can you think of a scenario where having this behavior as default would pose a problem? Personally I can't think of anything.
Rather than changing the way Osara currently works, perhaps creating custom actions would be a better and more elegant way to do it.
It should just be a matter of running the following action after the Osara ones:
Name: Edit: Select all CC events under selected notes
ID: 41735
I'll create custom actions for this and upload them to Dropbox, but I'd really love this behavior to be the default in the future.
Any thoughts?

MatejGolian:
Update: I've tested my idea and while it works, the drawback is that since CC events also get selected, one always has to use the up and down arrow keys if one wants to edit note events. It's pretty obvious - don't know why I haven't realized it... I assume that this would make things complicated and unintuitive for most people.
So, if you think it's a good idea, would it be possible to add the Edit: Select all CC events under selected notes action on its own to the key map as an alternative?

ScottChesworth:
> Would it be possible to add the Edit: Select all CC events under selected notes action on its own to the key map as an alternative?
> 
Sure. Any suggestion of a logical mapping?
Ideally we'd have action feedback implemented for that. Dunno if everything is already in place for a newb like me to be able to add that (I'm not familiar with how it all works in the MIDI editor yet), but I can try.

Flameborn:
Not sure if there has been any change regarding this, but in the fairly up-to-date OSARA version I am using, the left/right arrows move between notes only, and the Next/Previous CC event actions are unmapped.

While I don't see editing both midi notes and CC events at the same time, it would be nice to have a unified interface.

This way, the left/right arrows would select the Next/Previous event similarly to how it works now, except with the addition that a mapped action could toggle between all/midi notes only/CC events only when moving around, acting as a filter.

This way, there would be hardly any reason to switch the midi editor to all midi events view, as OSARA would essentially handle all events, possibly defaulting to selecting midi notes only, so that the existing behavior is preserved. To achieve what this issue requests/suggests, one would need to toggle to all events, and move around as expected.

Having these unified would be a huge timesaver, because CC 1 (usually mod wheel), CC 2 (breath controller), as well as CC 11 (expression) and CC 64 are used extensively, especially for orchestral instruments.

MatejGolian:
[...]

PR #650, opened 2022-01-18
Add feedback for "Edit: Toggle selection of all CC events under selected notes" in MIDI editor

RDMurray:
fixes: #574 
@MatejGolian  can you test the below build?

Draft for now as I haven't updated the README and translators comments yet.


src/reaper_osara.cpp
```diff
@@ -3782,6 +3782,8 @@ Command COMMANDS[] = {
 	{MIDI_EVENT_LIST_SECTION, {{0, 0, 40835}, NULL}, NULL, cmdMidiMoveToTrack}, // Activate next MIDI track
 	{MIDI_EDITOR_SECTION, {{0, 0, 40836}, NULL}, NULL, cmdMidiMoveToTrack}, // Activate previous MIDI track
 	{MIDI_EVENT_LIST_SECTION, {{0, 0, 40836}, NULL}, NULL, cmdMidiMoveToTrack}, // Activate previous MIDI track
+	{MIDI_EVENT_LIST_SECTION, {{0, 0, 40836}, NULL}, NULL, cmdMidiMoveToTrack}, // Activate previous MIDI track
```
jcsteh, line 3785: It looks like we already had a duplicate of 40836 and this adds another duplicate. :) Can you remove both please?

jcsteh (approved):
Thanks. Approved with comment addressed.

src/reaper_osara.cpp
RDMurray: duplicate removed.  The existing duplicate isn't really a duplicate, it is necessary for the command to work in the event list,  However it is not currently assigned in the keymap for the event list.

PR #656, opened 2022-01-22
Rewrite CC movement using the binary search iterator

### LINKED ISSUE #434: Unable to properly navigate between CCs at the same position

leonardder:
Fixes #434 

This pr uses the binary search iterator to navigate CCs. Similar to notes in chords, notes are stably sorted. Note however that sorting is unlikely to be equal to sorting in the event list, though sorting in there is unstable anyway.

Additionally, there are two seeminly unrelated changes in this pr, though I think they make sense here:

1. The CompareByPitch function for notes is moved to the MidiNote struct in order to keep consistency with MidiControlChange sorting.
2. I stumbled at the suggestion to use std::stable_sort instead of std::sort, that keeps order of adjacent vector items in their original order if they can't be sorted (e.g. when the pitch of two notes is equal). This can happen in chords when multiple midi tracks are merged into one item. I guess this is a theoretical thing that can easily be reverted. Wonder what you think @jcsteh .

@Batworx your feedback would be greatly appreciated. While this should fix your issue, it should also make cc navigation snappier in general.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -54,15 +54,20 @@ struct MidiControlChange {
 	static bool compareForSortAtPosition(const MidiControlChange& cc1, const MidiControlChange& cc2) {
 		if (cc1.message1 < cc2.message1) {
 			return true;
-		} if (cc1.message1 > cc2.message1) {
+		}
+		if (cc1.message1 > cc2.message1) {
 			return false;
-		} if (cc1.channel < cc2.channel) {
+		}
+		if (cc1.channel < cc2.channel) {
 			return true;
-		} if (cc1.channel > cc2.channel) {
+		}
+		if (cc1.channel > cc2.channel) {
 			return false;
-		} if (cc1.message2 < cc2.message2) {
+		}
+		if (cc1.message2 < cc2.message2) {
 			return true;
-		} if (cc1.message2 > cc2.message2) {
+		}
+		if (cc1.message2 > cc2.message2) {
 			return false;
 		}
 		return cc1.message3 < cc2.message3;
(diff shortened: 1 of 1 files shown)


src/midiEditorCommands.cpp
```diff
@@ -605,15 +627,12 @@ pair<int, int> findChord(MediaItem_Take* take, int direction) {
...
-// Used to order notes in a chord by pitch.
-bool compareNotesByPitch(const MidiNote& note1, const MidiNote& note2) {
-	return note1.pitch < note2.pitch;
-}
+// Keeps track of the CC to which the user last moved at a particular position.
+// The first entry of the pair is the position.
+// The second entry of the pair is the number of the CC at the position; e.g. 0 is the first CC.
+// It is not a REAPER CC index.
+// -1 means no position/CC.
+pair<double, int> currentCCAtPos = {-1, -1};
```
jcsteh, line 635: I'm not quite sure about the name currentCCAtPos. On one hand, I understand this now includes a position. On the other hand, it feels like currentCC is still a better name. It's not the current CC at a particular position; it's the current CC for the user, including both its position and index at that position. I think I'd prefer either currentCC (which would require renaming some local variables below) or currentCCAndPos.

jcsteh (changes requested):
Thanks! Wow. Reviewing this really hurt my head. It's not your code - I can't think of a way to do it better - but I just struggled to follow it for some reason.

leonardder:
I'm sorry for making it so difficult to grasp. I addressed your comments and added some additional comments in the process in the hope to make it more understandable now.

jcsteh (approved):
Nice commenting. That's super helpful.

PR #657, opened 2022-01-22
Use binary iterator in more cases for midi notes

leonardder:
This pr implements binary iteration for the following functions:

* cmdMidiMoveCursor. Note that this one was moved in the code to be able to just play the chord at the cursor
* findChord now returns iterators as suggested earlyer
* getSelectedNotes uses MidiNote::get


jcsteh (approved):
Thank you.

PR #659, opened 2022-01-26
Fix navigation for filtered events

### LINKED ISSUE #625: Show Only Events That Pass Filter Causing Problems With Osara's Chord Navigation
Hello all,

This seems a biggie to me, but I hope it's fixable.
OSARA: Move to next chord and OSARA: Move to previous chord do not work when a filter with the 'Show only events that pass filter' option enabled is active. This likely affects the command 'and add to selection' variants as well.
Checking that option in the filter dialog causes Osara to get stuck and instead of navigating between the filtered notes when pressing the left and right arrows, event navigation gets completely broken and Osara only reports things like the following:
"bar 4 beat 4 99% 252 notes
bar 1 beat 1 0% 186 notes"

If you filter all note events out completely, something similar happens. Furthermore, this behavior seems to have an impact on the Osara 'move to note in chord' commands, because pressing up/down results in Osara reporting 'C-1' (although no C-1 note gets actually selected) - as it shouldn't.

leonardder:
Fixes #625

In short, here is what's done:

1. Fetching the count has been moved to the several structs. This makes it easier to cound items when not using the iterator, e.g. in cmdMidiInsertNote
2. Fetching count for MidiNote and MidiControlChange is now done using MIDI_GetNote and MIDI_GetCC. This is still a concept, but if I understood information from Cockos correctly, MIDI_CountEvts is just as costly and doesn't care about filtered notes/CCs.
3. Struct values are now initialized to negative values.
4. When MIDI_GetNote or MIDI_GetCC fails, we now act accordingly. Position/start is set to FLT_MAX so if the iterator ever fails, the ghost notes are always at the end.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -438,7 +438,7 @@ class MidiEventIterator {
 		reqParams(ReqParams),
 		index(index)
 	{
-		this->count = this->getCount();
+		this->count = EventType::getCount(this->source);
 	}
 
 	bool operator==(const MidiEventIterator& other) const {
@@ -511,10 +511,6 @@ class MidiEventIterator {
 	}
 
 	protected:
-	int getCount() const {
-		return EventType::getCount(this->source);
-	}
-
 	value_type getEvent(difference_type index) const {
 		return EventType::get(this->source, index, this->reqParams);
 	}
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -186,7 +186,7 @@ struct MidiEventListData {
 	string message;
 	int offVel = -1;
 	double length = -1.0;
-	bool selected = -1;
+	bool selected;
 
 	struct ReqParams {
 	};
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -514,6 +514,7 @@ class MidiEventIterator {
 	int getCount() const {
 		return EventType::getCount(this->source);
 	}
+
 	value_type getEvent(difference_type index) const {
 		return EventType::get(this->source, index, this->reqParams);
 	}
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -13,6 +13,7 @@
 #include <map>
 #include <cassert>
 #include <functional>
+#include <float.h>
 #include "midiEditorCommands.h"
 #include "osara.h"
 #include "translation.h"
@@ -26,12 +27,12 @@ using namespace fmt::literals;
 // Note: while the below struct is called MidiControlChange in line with naming in Reaper,
 // It is also used for other MIDI messages.
 struct MidiControlChange{ 
-	int channel;
-	int index;
-	int message1;
-	int message2;
-	int message3;
-	double position;
+	int channel = -1;
+	int index = -1;
+	int message1 = -1;
+	int message2 = -1;
+	int message3 = -1;
+	double position = -1.0;
 	bool selected;
 	bool muted;
 
@@ -54,7 +55,7 @@ struct MidiControlChange{
 	static const MidiControlChange get(MediaItem_Take* take, int index, ReqParams params) {
 		MidiControlChange cc;
 		double position;
-		MIDI_GetCC(take, index,
+		if (MIDI_GetCC(take, index,
 			params.selected ? &cc.selected : nullptr,
 			params.muted ? &cc.muted : nullptr,
 			params.position ? &position : nullptr,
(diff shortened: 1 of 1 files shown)


leonardder:
@MatejGolian could you please test this build re #625?

MatejGolian:
@leonardder, I don't know what kind of sorcery you are doing, but it works!!! 😊
I only conducted a very quick test, because I don't have much time right now. I'll do some more testing tomorrow, but I assume that it'll be fine.
Thanks again!!!

leonardder:
> @leonardder, I don't know what kind of sorcery you are doing, but it works!!! 😊 

That's good news. I changed how we count events according to information received from Cockos. They turn out to be very cooperative!

MatejGolian:
I haven't managed to trick your code into misbehaving today either, so as far as I can tell it's all great. 👍
Thanks once again - the user experience and editing possibilities in the MIDI editor have tremendously improved and expanded with this fix. To be honest, I was afraid that this was a Reaper bug and that we'd have to wait several weeks for the Reaper update to arrive, but I'm happy that my worries have been unfounded and that this bug wasn't as horrible as it seemed. And of course, thanks for being so quick!

leonardder:
@jcsteh Could you have a look? I sent you some information by mail yesterday that should clarify why I'm using GetNote/GetCC to count. May be we should do a performance test with a huge midi item to be really sure.

leonardder:
@Lo-lo78 You tend to work with large files isn't it? May be you could test drive this build to see how it behaves performance wise for you?

Lo-lo78:
@leonardder Yup! It's perfect, everything works!
It all seems much easier to use.
Good job!
👍

src/midiEditorCommands.cpp
```diff
@@ -134,35 +146,46 @@ struct MidiNote {
...
-	double position;
+	int index = -1;
+	double position = -1.0;
 	string message;
-	int offVel;
-	double length;
-	bool selected;
+	int offVel = -1;
+	double length = -1.0;
+	bool selected = -1;
```
jcsteh: We shouldn't be putting -1 in a bool even if C++ allows it. We could just let this one be false by default - there are plenty of other ways to tell it is invalid - or we could use int. I'd prefer the former personally.

src/midiEditorCommands.cpp
```diff
@@ -482,7 +514,9 @@ class MidiEventIterator {
 	}
 
 	protected:
-	int getCount() const;
+	int getCount() const {
+		return EventType::getCount(this->source);
+	}
```
jcsteh, line 516:
```suggestion
	}

```

leonardder:
As far as I understand from the comments, performance isn't affected significantly and this pr offers great gain for people working with filtered events. I think this is ready to go.

src/midiEditorCommands.cpp
```diff
@@ -479,7 +511,9 @@ class MidiEventIterator {
 	}
 
 	protected:
-	int getCount() const;
+	int getCount() const {
```
jcsteh, line 514: We don't seem to use this anywhere except the iterator's own constructor. Can we get rid of this and call `EventType::getCount` directly in the constructor?

jcsteh (commented):
Thanks for working on this.

> 1. Fetching the count has been moved to the several structs. This makes it easier to cound items when not using the iterator, e.g. in cmdMidiInsertNote

Unless I'm missing something, I don't see the `::getCount` calls used anywhere outside of the iterator. I don't mind this change, but I'm just wondering whether there's something missing here.

> 2. Fetching count for MidiNote and MidiControlChange is now done using MIDI_GetNote and MIDI_GetCC. This is still a concept, but if I understood information from Cockos correctly, MIDI_CountEvts is just as costly and doesn't care about filtered notes/CCs.

Just to clarify, counting with MIDI_GetNote/CC does respect filtering though, which is why you're switching here, correct? Which leads me to this question:

> 3. Struct values are now initialized to negative values.
> 4. When MIDI_GetNote or MIDI_GetCC fails, we now act accordingly. Position/start is set to FLT_MAX so if the iterator ever fails, the ghost notes are always at the end.

I assume this is just caution, though, and this shouldn't ever happen in reality because you're now counting differently as above such that filtering is respected?

leonardder:
> > 1. Fetching the count has been moved to the several structs. This makes it easier to cound items when not using the iterator, e.g. in cmdMidiInsertNote
> 
> Unless I'm missing something, I don't see the `::getCount` calls used anywhere outside of the iterator. I don't mind this change, but I'm just wondering whether there's something missing here.

Nope, this was intentional. I wanted this pr to focus on the fix for filtered events. With the movement of the count logic, I wanted to pre-sort on a follow up pr that will deal with this.
In short, I'm intending to file the following after this pr:

1. A PR that will tackle the remaining CC cases that don't use the iterator or MidiControlChange::get
2. A PR that replaces all remaining calls of Midi_CountEvts with their MidiControlChange/MidiNote::getCount equivalents.

I have to do that in this order because if I wouldn't, I'd probably do double work.

> > 2. Fetching count for MidiNote and MidiControlChange is now done using MIDI_GetNote and MIDI_GetCC. This is still a concept, but if I understood information from Cockos correctly, MIDI_CountEvts is just as costly and doesn't care about filtered notes/CCs.
> 
> Just to clarify, counting with MIDI_GetNote/CC does respect filtering though, which is why you're switching here, correct? 

Correct indeed.

> > 3. Struct values are now initialized to negative values.
> > 4. When MIDI_GetNote or MIDI_GetCC fails, we now act accordingly. Position/start is set to FLT_MAX so if the iterator ever fails, the ghost notes are always at the end.
> 
> I assume this is just caution, though, and this shouldn't ever happen in reality because you're now counting differently as above such that filtering is respected?

This is also correct. Point 4 was on the urgent advice of cockos and deals with the case where MidiNote::get would retrieve a note that is out of range. In the old situation, it would return a struct with zeroes.

jcsteh (approved):
Thanks.

leonardder:
Thanks for approving. You might have missed https://github.com/jcsteh/osara/pull/659/commits/0e136fc88f1ccca7db92638ba10026216bcdfe46 but it was on your request, so I guess it will be OK 😉

jcsteh:
Damn it. Accidentally hit merge before updating the commit message. That's not ideal, as there's a fair bit of context there. :(

For the record, the message was supposed to be:

Fix navigation for filtered events. (issue #625, PR #659)

1. Fetching the count has been moved to the various structs. This will make it easier to count items in future patches when not using the iterator, e.g. in cmdMidiInsertNote
2. Fetching count for MidiNote and MidiControlChange is now done using MIDI_GetNote and MIDI_GetCC, which respects filtering. MIDI_CountEvts is just as costly and doesn't care about filtered notes/CCs.
3. Struct values are now initialized to negative values.
4. When MIDI_GetNote or MIDI_GetCC fails, we now act accordingly. Position/start is set to DBL_MAX so if the iterator ever fails, the ghost notes are always at the end. This shouldn't happen in reality due to 2), but this is just to be safe.

leonardder:
I wouldn't mind a revise of the commit, I guess master isn't fetched every minute and even if it is, people probably know how to reset.

jcsteh:
The problem is that the pull request won't link the right commit if I do that. Still, it's probably worth it in this case.

leonardder:
It will have a reference from the commit description at least.

PR #666, opened 2022-02-01
Attempt to support move to nearest event on Mac

leonardder:
It looks like this should be possible now. Let's see whether this builds.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1372,7 +1372,6 @@ void cmdMidiNoteSplitOrJoin(Command* command) {
 	}
 }
 
-#ifdef _WIN32
 using MidiEventListDataIterator = MidiEventIterator<MidiEventListData, HWND>;
 
 template<>
@@ -1425,6 +1424,7 @@ void cmdFocusNearestMidiEvent(Command* command) {
 	focusNearestMidiEvent(hwnd);
 }
 
+#ifdef _WIN32
 void cmdMidiFilterWindow(Command *command) {
 	HWND editor = MIDIEditor_GetActive();
 	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
--- a/src/midiEditorCommands.h
+++ b/src/midiEditorCommands.h
@@ -48,8 +48,8 @@ void cmdMidiMoveToPrevItem(Command* command) ;
 void cmdMidiMoveToTrack(Command* command);
 void cmdMidiSelectSamePitchStartingInTimeSelection(Command* command) ;
 void cmdMidiNoteSplitOrJoin(Command* command);
-#ifdef _WIN32
 void cmdFocusNearestMidiEvent(Command* command);
+#ifdef _WIN32
 void cmdMidiFilterWindow(Command* command);
 void maybeHandleEventListItemFocus(HWND hwnd, long childId);
 #endif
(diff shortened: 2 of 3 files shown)


leonardder:
@ScottChesworth could you please give this a spin? Note that this doesn't mean that we can add edit cursor follows event list focus as well, unfortunately.

ScottChesworth:
I can and will, but @pitermach or @jennykbrennan might be able to get you feedback quicker.

pitermach:
I can certainly take a look when I leave work in a few hours. However I don’t work in the midi editor that often so if you can tell me what this should do and how to test that’d be great ☺️

ScottChesworth:
Try these steps:
1. In the MIDI Event List Editor section of the action list, you'll find a new action called "OSARA: Focus event nearest edit cursor". Bind it to Command+F.
2. Insert a MIDI item, select it and open the MIDI Editor by hitting Command+Option+E.
3. When the editor opens, hit Option+1 to ensure that you're in piano roll view to start with, then use RightArrow to jump forward through notes and chords until you've moved a few bars along the timeline. For this test, it's important to remember the bar/beat/percent position of where you've stopped in the piano roll.
4. Now hit Option+3 to switch to Event List view. This step will illustrate the issue Leonard is trying to provide a workaround for. Hit DownArrow to start navigating through events, and you should find that the edit cursor position hasn't synced, so you'll be navigating through events from the start of the file, not from bar 3 or wherever you were when you left the piano roll view. Doh!
5. Retry steps 3 and 4 leaving the piano roll in the same place again, but this time, when you switch to the Event List view, hit Command+F before starting to navigate. If the new action works as hoped, it will move your cursor to the correct event and you'll be able to pick up navigation from the same place you were at when you switched away from the piano roll view.
Let us know whatcha find.

jennykbrennan:
I took this for a spin but when switching between piano roll and event list, the cursor followed the notes without fail. So it stayed in synk. At least now I don't appear to have an issue. I' don't work much in the midi editor so it's not something I've really worked with. I  simply tested  according to Scott's steps and found no problems. I did not use the new action as I didn't need it. :)

pitermach:
Thanks for the clear instructions!

I gave this build a try, recorded some midi and followed the steps. If I run the action, then the selection in the event list gets cleared and a press of down arrow starts me back at the top.

However, if I move around in the piano role with the arrow keys, then return to the event list with OPT+3, it puts me on the correct event - which judging by what you said was supposed to be not working.

jennykbrennan:
Update. Here, when running the action in the event list, the cursor  either stays where it is or focuses on the next event. It doesn't  start me back at the top or clear the selection.

pitermach:
Update from me as well, on more testing now the shortcut is just usually doing nothing. I'm guessing this is because the edit cursor and event list are following each other pretty well, doesn't matter which one I move but if I check my position in the project with command+Shift+J it stays in sync.

leonardder:
To put it the other way around, if you move in the event list and move back into the piano roll, does it also sync in that case?

pitermach:
It does here, yes

leonardder:
That is pretty interesting. May be REAPER on Mac is already doing something that Windows doesn't do, or this is something new.

leonardder:
@ScottChesworth Could you have a final word about this? It would help if we'd have sort of a comparison between Mac and Windows of how OSARA behaves currently (i.e. a normal snapshot branch). I think this pr can safely be closed as it doesn't offer us any value.

ScottChesworth:
Oops, only just remembered I was meant to report back about this. Yeah, the Mac seems to already be more stable when switching between views, so closing as suggested. Sorry about the extraneous work dude.

PR #668, opened 2022-02-02
add action "OSARA: Report regions, most recent marker and items at current position", fixes #253

### LINKED ISSUE #253: Requesting new OSARA action to report currently playing item or items
@vick08 has been trying to figure out how to identify which item is currently playing on RWP, and the only workflow I can think of to identify it is extremely button-bashy (pausing the transport, navigating through tracks hitting Shift+A on each one until an item is selected, then hitting Control+Shift+Space to report selection). Perhaps OSARA could provide a more elegant solution?

Thanks in advance.

RDMurray:
New action to report regions, the most recent marker, and items on selected tracks under the cursor.  Uses the edit cursor if stopped, otherwise the play cursor.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3820,8 +3820,8 @@ void cmdReportRegionMarkerItems(Command* command) {
 		return (isTrackSelected(track) && isPosInItem(pos, item));
 	}, multiLine);
 	if(multiLine) {
-		// Translators: The title of the review message for the action "OSARA: report regions, most recent marker and items at current position".
-		reviewMessage(translate("regions, most recent marker and items at current position"), s.str().c_str());
+		// Translators: The title of the review message for the action "OSARA: Report regions, last project marker and items on selected tracks at current position".
+		reviewMessage(translate("At Current Position"), s.str().c_str());
 	} else{
 		outputMessage(s);
 	}
@@ -3977,7 +3977,7 @@ Command COMMANDS[] = {
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: About"}, "OSARA_ABOUT", cmdAbout},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Report groups for current track"}, "OSARA_REPORTTRACKGROUPS", cmdReportTrackGroups},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Mute next message from OSARA"}, "OSARA_MUTENEXTMESSAGE", cmdMuteNextMessage},
-	{ MAIN_SECTION, {DEFACCEL, "OSARA: Report regions, most recent marker and items at current position"}, "OSARA_REPORTREGIONMARKERITEMS",cmdReportRegionMarkerItems},
+	{ MAIN_SECTION, {DEFACCEL, "OSARA: Report regions, last project marker and items on selected tracks at current position"}, "OSARA_REPORTREGIONMARKERITEMS",cmdReportRegionMarkerItems},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Go to first track"}, "OSARA_GOTOFIRSTTRACK", cmdGoToFirstTrack},
 	{ MAIN_SECTION, {DEFACCEL, "OSARA: Go to last track"}, "OSARA_GOTOLASTTRACK", cmdGoToLastTrack},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, "OSARA: Enable noncontiguous selection/toggle selection of current chord/note"}, "OSARA_MIDITOGGLESEL", cmdMidiToggleSelection},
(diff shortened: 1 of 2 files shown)


Lo-lo78:
@RDMurray I have inserted a region, a marker and an item in a track;
the region is in a separate row while the marker and item are in a single row without separation.
For instance:
Intro region 
Intro11.1 Tenax [chan 10]

The marker and the item are in the same line and for a speech synthesis the numbers can be misread.
Maybe this works best:
Intro region 
Intro marker
11.1 Tenax [chan 10] 

Greetings!

RDMurray:
@Lo-lo78  Thanks for letting me know.  I missed a new line.  I'll fix it tomorrow.

RDMurray:
@Lo-lo78  should be fixed now.  Let me know if you think the feedback is okay.  I'm trying to keep it as short as possible, similar to the report selection command.

RDMurray:
@Lo-lo78  Sorry, forgot to commit the fix.  Next build should be okay.

Lo-lo78:
@RDMurray 
YES! Works well.
Just a stupid thought:
When the OSARA window is opened with a double click of the shortcut, the system cursor is at the end even if the vocal synthesis reads the first line.
That's not a problem for me.
It also happens with action:
OSARA: About
It is the same problem.
It is okay for me!
👍

src/reaper_osara.cpp
```diff
@@ -3018,8 +3022,8 @@ void cmdIoMaster(Command* command) {
 void cmdReportRippleMode(Command* command) {
 	postCycleRippleMode(command->gaccel.accel.cmd);
 }
-
-string formatTracksWithState(const char* prefix, TrackStateCheck checkState,
+template <typename Func>
```
jcsteh, line 3025: Neat! :)

src/reaper_osara.cpp
```diff
@@ -3762,6 +3776,57 @@ void cmdToggleLoopSegScrub(Command* command) {
...
+		}
+	}
+	separate();
+	s << formatItemsWithState([pos](MediaItem* item) -> bool{
+		MediaTrack* track = GetMediaItem_Track(item);
+		return (isTrackSelected(track) && isPosInItem(pos, item));
+	}, multiLine);
+	if(multiLine) {
+		// Translators: The title of the review message for the action "OSARA: report regions, most recent marker and items at current position".
+		reviewMessage(translate("regions, most recent marker and items at current position"), s.str().c_str());
```
jcsteh, line 3824: 1. nit: GUI is the one case where we don't use all lower case. For controls, sentence case; i.e. only upper case the first letter. For dialog titles, title case.
2. This is a very verbose title. Could we shorten it to "At Current Position" or something?

src/reaper_osara.cpp
```diff
@@ -3762,6 +3776,57 @@ void cmdToggleLoopSegScrub(Command* command) {
...
+			s << name;
+		} else {
+			// Translators: used to report an unnamed marker. {} is replaced with the marker number.  
+			s << format(translate("marker {}"), number);
+		}
+	}
+	separate();
+	s << formatItemsWithState([pos](MediaItem* item) -> bool{
+		MediaTrack* track = GetMediaItem_Track(item);
+		return (isTrackSelected(track) && isPosInItem(pos, item));
```
jcsteh, line 3820: I understand why we only report items on selected tracks. Otherwise, we could report many, many items. However, I worry that this might not be obvious. I don't know if it's feasible to communicate this in the action name, but I wonder if we should at least note it in the readme.

jcsteh (changes requested):
Thanks!

jcsteh:
When the message review dialog opens, the cursor isn't at the end exactly, but it does select all the text in the control. That means pressing down arrow throws you to the end. I agree that's kinda annoying. Would you mind filing a separate issue for that?

src/reaper_osara.cpp
ScottChesworth: > I don't know if it's feasible to communicate this in the action name
> 
How's about OSARA: Report regions, closest project marker and items on selected tracks at current position

src/reaper_osara.cpp
RDMurray: I didn't know whether to go with a template or a function object parameter. Probably doesn't make much difference, but do you have a preference?

src/reaper_osara.cpp
RDMurray: @ScottChesworth  yes, the only thing is the marker is the nearest marker before the current position. What about "OSARA: Report regions, last project marker and items on selected tracks at current position"?

ScottChesworth: > What about "OSARA: Report regions, last project marker and items on selected tracks at current position"?
> 
That's got it :)

src/reaper_osara.cpp
jcsteh: I don't have a preference here. This is fine.

jcsteh (approved):

PR #673, opened 2022-02-07
Fix crash in Midi Event list caused by oversight in #659

leonardder:
In pull request #659, the MidiEventListData struct was changed to initialize several members to -1 by default, including the length member. However, to check whether a MidiEventListData is a note, we checked for `length == 0` before and that wasn't changed to reflect the new behaviour of default initialization to -1. This is now fixed.


leonardder:
Merging this instantly because it is a very small change/fix with a major positive outcome. I hope you don't mind @jcsteh but I've seen @ScottChesworth doing unapproved merges as well. May be we should set some rules on this matter at some point.

PR #674, opened 2022-02-07
Add feedback for commands that change item rate and pitch. fixes #307

### LINKED ISSUE #307: Requesting feedback for actions that change item pitch and rate
As a sound designer, I use these actions very frequently so it'd be really useful to have feedback for them, also because changing the project playrate provides speech so it'd be better UX consistency. Not all of these actions currently have keymap assignments but I'm including them here as well for consistency:

- Shift+1 Item properties: Decrease item rate by ~0.6% (10 cents); Command ID: 40520
- Shift+2 Item properties: Increase item rate by ~0.6% (10 cents); Command ID: 40519
- Shift+3 Item properties: Decrease item rate by ~0.6% (10 cents), clear 'preserve pitch'; Command ID: 40800
- Shift+4 Item properties: Increase item rate by ~0.6% (10 cents), clear 'preserve pitch'; Command ID: 40799
- Item properties: Decrease item rate by ~6% (one semitone); Command ID: 40518
- Item properties: Increase item rate by ~6% (one semitone); Command ID: 40517
- Shift+5 Item properties: Decrease item rate by ~6% (one semitone), clear 'preserve pitch'; Command ID: 40798
- Shift+6 Item properties: Increase item rate by ~6% (one semitone), clear 'preserve pitch'; Command ID: 40797
- Shift+7 Item properties: Pitch item down one cent; Command ID: 40207
- Shift+8 Item properties: Pitch item up one cent; Command ID: 40206
- Shift+9 Item properties: Pitch item down one semitone; Command ID: 40205
- Shift+0 Item properties: Pitch item up one semitone; Command ID: 40204
- Item properties: Pitch item down one octave; Command ID: 40516
- Item properties: Pitch item up one octave; Command ID: 40515
- Ctrl+Backspace Item properties: Reset item pitch; Command ID: 40653
- Alt+Backspace Item properties: Set item rate to 1.0; Command ID: 40652

The best feedback for these would probably be announcing either the new item rate or pitch depending on what the action is modifying, or just saying "x items rate/pitch increased/decreased) if multiple are selected, just like actions on multiple midi notes in the editor are currently handled.

RDMurray:
Add feedback for:
- Item properties: Decrease item rate by ~0.6% (10 cents)
- Item properties: Decrease item rate by ~0.6% (10 cents), clear 'preserve pitch'
- Item properties: Decrease item rate by ~6% (one semitone)
- Item properties: Decrease item rate by ~6% (one semitone), clear 'preserve pitch'
- Item properties: Increase item rate by ~0.6% (10 cents)
- Item properties: Increase item rate by ~0.6% (10 cents), clear 'preserve pitch'
- Item properties: Increase item rate by ~6% (one semitone)
- Item properties: Increase item rate by ~6% (one semitone), clear 'preserve pitch'
- Item properties: Set item rate from user-supplied source media tempo/bpm...
- Item properties: Set item rate to 1.0
- Item properties: Pitch item down one cent
- Item properties: Pitch item up one cent
- Item properties: Pitch item down one semitone
- Item properties: Pitch item up one semitone
- Item properties: Pitch item down one octave
- Item properties: Pitch item up one octave
- Item properties: Reset item pitch
- Item properties: Toggle take preserve pitch
- Item properties: Clear take preserve pitch
- Item properties: Set take preserve pitch

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1796,6 +1796,40 @@ void postToggleFreeItemPositioning(int command) {
 		translate("disabled free item positioning"));
 }
 
+void postChangeItemRate(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item) {
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
+	double rate = GetMediaItemTakeInfo_Value(take, "D_PLAYRATE");
+	// Translators: Used when changing item rate. {:g} is replaced by the new rate. E.G. "1.0 item rate"
+	outputMessage(format(translate("{:g} item rate"), rate));
+}
+
+void postChangeItemPitch(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item) {
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
+	double pitch = GetMediaItemTakeInfo_Value(take, "D_PITCH");
+	// Translators: Used when changing item PITCH. {:+g} is replaced by the new PITCH. E.G. "-1.0 SEMITONES"
+	outputMessage(format(translate("{:+g} semitones"), pitch));
+}
+
+void postToggleTakePreservePitch(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item) {
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
+	bool isPreserving = *(bool*)GetSetMediaItemTakeInfo(take, "B_PPITCH", nullptr);
+	outputMessage(isPreserving ?
+		translate("take preserve pitch enabled"):
+		translate("take preserve pitch disabled"));
+}
+
 typedef void (*PostCommandExecute)(int);
 typedef struct PostCommand {
 	int cmd;
@@ -1961,6 +1995,26 @@ PostCommand POST_COMMANDS[] = {
 	{40219, postChangeTransientDetectionThreshold}, // Transient detection threshold: Decrease
 	{40070, postToggleEnvelopePointsMoveWithMediaItems}, // Options: Envelope points move with media items
 	{40641, postToggleFreeItemPositioning}, // Track properties: Toggle free item positioning
+	{40520, postChangeItemRate}, // Item properties: Decrease item rate by ~0.6% (10 cents)
+	{40800, postChangeItemRate}, // Item properties: Decrease item rate by ~0.6% (10 cents), clear 'preserve pitch'
+	{40518, postChangeItemRate}, // Item properties: Decrease item rate by ~6% (one semitone)
+	{40798, postChangeItemRate}, // Item properties: Decrease item rate by ~6% (one semitone), clear 'preserve pitch'
+	{40519, postChangeItemRate}, // Item properties: Increase item rate by ~0.6% (10 cents)
+	{40799, postChangeItemRate}, // Item properties: Increase item rate by ~0.6% (10 cents), clear 'preserve pitch'
+	{40517, postChangeItemRate}, // Item properties: Increase item rate by ~6% (one semitone)
+	{40797, postChangeItemRate}, // Item properties: Increase item rate by ~6% (one semitone), clear 'preserve pitch'
+	{42374, postChangeItemRate}, // Item properties: Set item rate from user-supplied source media tempo/bpm...
+	{40652, postChangeItemRate}, // Item properties: Set item rate to 1.0
+	{40207, postChangeItemPitch}, // Item properties: Pitch item down one cent
+	{40206, postChangeItemPitch}, // Item properties: Pitch item up one cent
+	{40205, postChangeItemPitch}, // Item properties: Pitch item down one semitone
+	{40204, postChangeItemPitch}, // Item properties: Pitch item up one semitone
+	{40516, postChangeItemPitch}, // Item properties: Pitch item down one octave
+	{40515, postChangeItemPitch}, // Item properties: Pitch item up one octave
+	{40653, postChangeItemPitch}, // Item properties: Reset item pitch
+	{40566, postToggleTakePreservePitch}, // Item properties: Toggle take preserve pitch
+	{40796, postToggleTakePreservePitch}, //Item properties: Clear take preserve pitch
+	{40795, postToggleTakePreservePitch}, // Item properties: Set take preserve pitch
 	{0},
 };
 MidiPostCommand MIDI_POST_COMMANDS[] = {


readme.md
```diff
@@ -184,6 +184,27 @@ Most of these are actions built into REAPER, but a few are very useful actions f
...
+- Item properties: Pitch item up one cent: Shift+8
+- Item properties: Pitch item down one semitone: shift+9
+- Item properties: Pitch item up one semitone: Shift+0
+- Item properties: Pitch item down one octave
+- Item properties: Pitch item up one octave
+- Item properties: Reset item pitch: Control+Backspace
+- Item properties: Toggle take preserve pitch
+- Item properties: Clear take preserve pitch
+- Item properties: Set take preserve pitch
+
```
jcsteh, line 207: nit: Unnecessary blank line.
```suggestion
```

src/reaper_osara.cpp
```diff
@@ -1796,6 +1796,40 @@ void postToggleFreeItemPositioning(int command) {
 		translate("disabled free item positioning"));
 }
 
+void postChangeItemRate(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item) {
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
```
jcsteh, line 1804: Please null check take. It's possible for an item to have no takes.

src/reaper_osara.cpp
```diff
@@ -1796,6 +1796,40 @@ void postToggleFreeItemPositioning(int command) {
...
+	// Translators: Used when changing item rate. {:g} is replaced by the new rate. E.G. "1.0 item rate"
+	outputMessage(format(translate("{:g} item rate"), rate));
+}
+
+void postChangeItemPitch(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item) {
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
```
jcsteh, line 1815: Ditto.

src/reaper_osara.cpp
```diff
@@ -1796,6 +1796,40 @@ void postToggleFreeItemPositioning(int command) {
...
+void postToggleTakePreservePitch(int command) {
+	MediaItem* item = getItemWithFocus();
+	if(!item) {
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
+	bool isPreserving = *(bool*)GetSetMediaItemTakeInfo(take, "B_PPITCH", nullptr);
+	outputMessage(isPreserving ?
+		translate("take preserve pitch enabled"):
+		translate("take preserve pitch disabled"));
```
jcsteh, line 1830: Let's do "enabled take preserve pitch", etc. for efficiency.

jcsteh (changes requested):
Thanks!

RDMurray:
Using these commands to return to a pitch of 0 or a rate of 1 results in a tiny difference due to floating point arithmetic.  This isn't an OSARA problem, because as far as I can tell it is invisible to sighted users.  @jcsteh do you think we should round the number, like in the properties dialog, or somehow report that there is still a small change in pitch or rate?

Either way, Cockos should probably fix this as it affects sighted users as well.

jcsteh:
What do we end up reporting currently with this PR as is? I don't love the
idea of rounding, but presenting a number to the user in scientific
notation (if that is what we do) is probably worse.

RDMurray:
The pitch currently reports as a ridiculously small number in scientific notation.  Should we just round it to 0 if it is less than 0.000001 or something?  It hides the problem, but Cockos should probably fix it.

jcsteh:
Yuck. Let's not report scientific notation to users. So yeah, let's round if ridiculously small.

On another note, I wish there were a floating point format that never used scientific notation, rounded to the nearest n decimal places but didn't include trailing zeroes. Format g is the best we can do if we want to avoid trailing zeroes, but it does have this annoying issue of falling back to scientific notation.

ScottChesworth:
I have an UX proposal (one that I'm certain that @jcsteh will balk at, sorry in advance for making your brain itch dude).
When the value is within the boundaries that we'll round, would prepending wording like "close to" or "almost" when we've rounded be useful information?
That way, OSARA is doing its bit to make users aware that they're not quite settled at a perfect 1.0, and we can add a note in the ReadMe to explain why they'll need to reset rate or pitch.
I'm thinking temporary measures to get as close as we can to reliable contextual feedback until Cockos fix this, basically.

pitermach:
Just tried this out and it's working great here. Curiously though, I can always consistently return to an exact 1 rate/0 pitch here on the test file I have loaded, I get numbers with 6 digits when adjusting the rate but that's normal and happens with the master playrate as well. This is on the ARM64 Reaper 6.47 RC1 on Mac, haven't tried on Windows yet and I'm not sure which of these variables is the deciding factor here. ☺️

If you're seeing this big numbers on Windows I would either round it to the last 3 digits or just leave it for Cockos. Hearing something like "almost 0" sounds a little imprecise to me and could mean a lot of things.

RDMurray:
@pitermach I can make it do it every time here with pitch by doing shift+0 once followed by shift+7 100 times.  Interestingly it is stored in the RPP as 0 so maybe Cockos already fixed this.  I definitely remember someone sending me a project to look at that had exactly this problem in the rpp file. 

@jcsteh Maybe we should have a function that formats a double with the f format specifier and then strips trailing zeros.

Rounding prevents small numbers from being output as scientific notation, but that means that anything less than 0.0001 becomes 0, and large numbers still produce scientific notation.  The numbers in item properties are something like {:.6f}.

@ScottChesworth  I don't think near 0 or near 1 should be necessary.  If the pitch is shown in the properties dialog as 0.000000, it should have no affect regardless of what the API returns.  I think any other behaviour would be a Reaper bug.

src/reaper_osara.cpp
```diff
@@ -1796,6 +1802,55 @@ void postToggleFreeItemPositioning(int command) {
...
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
+	if(!take) {
+		return;
+	}
+	double rate = GetMediaItemTakeInfo_Value(take, "D_PLAYRATE");
+	// if(abs(rate) < 0.0001) { // round to 0 to avoid outputting scientific notation
+	// 	rate = 0.0;
+	// }
```
jcsteh, line 1817: Remove commented code.
```suggestion
```

src/reaper_osara.cpp
```diff
@@ -1796,6 +1802,55 @@ void postToggleFreeItemPositioning(int command) {
...
+		return;
+	}
+	MediaItem_Take* take = GetActiveTake(item);
+	if(!take) {
+		return;
+	}
+	double pitch = GetMediaItemTakeInfo_Value(take, "D_PITCH");
+	// if(abs(pitch) < 0.0001) { // round to 0 to avoid outputting scientific notation
+	// 	pitch = 0.0;
+	// }
```
jcsteh, line 1834:
```suggestion
```

jcsteh (changes requested):

src/reaper_osara.cpp
```diff
@@ -666,6 +666,12 @@ void shortenFxName(char* name, ostringstream& s) {
 	}
 }
 
+string formatDouble(double d, int precision, bool plus=false) {
```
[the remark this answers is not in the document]
RDMurray: I looked into it a bit, but didn't find much.  One [Stack overflow answer](https://stackoverflow.com/questions/58526802/how-to-add-custom-format-specifier-to-fmt) says "You cannot add a new format specifier to a built-in formatter ".  No idea if that is still true.  we could write a formatter for a custom type, but I don't see much advantage as we would have to call the constructor for our custom double type anyway.

src/reaper_osara.cpp
```diff
@@ -666,6 +666,12 @@ void shortenFxName(char* name, ostringstream& s) {
...
+	string s = format(plus ? "{:+.{}f}" : "{:.{}f}", d, precision);
+	size_t pos = s.find_last_not_of("-+0.");
```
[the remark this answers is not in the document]
RDMurray: It covers the cases where s is "+0" and "-0", resulting in an empty string. Then "0" is returned.

src/reaper_osara.cpp
RDMurray: oops sorry 🤦

src/reaper_osara.cpp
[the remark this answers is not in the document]
jcsteh: We could fork (or contribute to) fmt if we needed/wanted to in future.

jcsteh, line 669:
```suggestion
// Format a double d to precision decimal places, stripping trailing zeroes.
// If plus is true, a "+" prefix will be included for a positive number.
string formatDouble(double d, int precision, bool plus=false) {
```

jcsteh (approved):

PR #675, opened 2022-02-09
Script to sort the keymap files so that diffs are smaller

RDMurray:
Add tools/sortKeymap.py

usage: python tools/sortKeymap.py <Reaper keymap file>

To be used before committing a new keymap file.  It sorts the entries in the file so that diffs are smaller and more readable.

Also add "*.bak" to .gitignore; sortKeymap.py saves the original file with the .bak extension.

Include sorted versions of the latest Windows and Mac keymaps.


tools/sortKeymap.py
```diff
@@ -0,0 +1,57 @@
+import sys
```
leonardder, line 1: Copyright info:

```suggestion
# OSARA: Open Source Accessibility for the REAPER Application
# Keymap sorting utility
# Copyright 2022 Robbie Murray
# License: GNU General Public License version 2.0

import sys
```

tools/sortKeymap.py
```diff
@@ -0,0 +1,57 @@
...
+			elif key == 'SCR':
+				scrs.append(match)
+			elif key == 'KEY':
+				keys.append(match)
+		else:
+			raise Exception("Unable to parse line: "+line)
+
+	acts.sort(key = lambda m: (m.group(3), m.group(5)))
+	scrs.sort(key = lambda m: (m.group(3), m.group(5)))
+	#sort keys by section, action id, modifiers and key.
```
leonardder, line 37:
```suggestion
	# sort keys by section, action id, modifiers and key.
```

tools/sortKeymap.py
```diff
@@ -0,0 +1,57 @@
...
+	return [match.group(0) for match in (acts+scrs+keys)]
+
+def main():
+	fn = sys.argv[1]
+	orig = open(fn, encoding='utf-8').readlines()
+	os.replace(fn, fn+'.bak')
+	try:
+		sorted = sortKeyMap(orig)
+		with open(fn, 'x', encoding='utf-8') as outFile:
+			outFile.write('\n'.join(sorted))
```
leonardder, line 48: Are you sure this ought to be \n only, i.e. not \r\n? Something like this might be nicer.
You can also skip the line below if doing something like the following

```suggestion
			outFile.writeline(f"{l}{os.linesep}" for l in sorted)
```

leonardder (changes requested):
This is awesome, thank you! May be throing this through a linter will catch some additional things.

tools/sortKeymap.py
```diff
@@ -0,0 +1,57 @@
...
+	acts.sort(key = lambda m: (m.group(3), m.group(5)))
+	scrs.sort(key = lambda m: (m.group(3), m.group(5)))
+	#sort keys by section, action id, modifiers and key.
+	keys.sort(key = lambda m: (m.group(5), m.group(4), m.group(2),m.group(3)))
+	return [match.group(0) for match in (acts+scrs+keys)]
+
+def main():
+	fn = sys.argv[1]
+	orig = open(fn, encoding='utf-8').readlines()
+	os.replace(fn, fn+'.bak')
```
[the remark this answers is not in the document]
RDMurray: I'm happy to change this, but why do you prefer string interpolation over concatenation in this case?

tools/sortKeymap.py
RDMurray: I'm pretty sure \n always translates to the platform specific newline when writing to a file opened in text mode.  It certainly works in Windows, translating to CRLF.  I'm not attached to this way of doing it though.  Still curious about why to use string interpolation; afaik it should perform worse than concatenation, not that it matters here.

tools/sortKeymap.py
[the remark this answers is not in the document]
leonardder: I tend to find interpolation more readable, but I don't incist. However, if you stick to the current format, ensure that there are spaces around the + operator:

```suggestion
	os.replace(fn, fn + '.bak')
```

tools/sortKeymap.py
```diff
@@ -0,0 +1,57 @@
...
+				acts.append(match)
+			elif key == 'SCR':
+				scrs.append(match)
+			elif key == 'KEY':
+				keys.append(match)
+		else:
+			raise Exception("Unable to parse line: "+line)
+
+	acts.sort(key = lambda m: (m.group(3), m.group(5)))
+	scrs.sort(key = lambda m: (m.group(3), m.group(5)))
```
jcsteh, line 36: What is the significance of the sorting here? It'd be good to comment this at least. Alternatively, you could use named regexp groups `(?P<groupName>)` and then `match.group("groupName")` to make it clearer what the groups are.

jcsteh (changes requested):
Thanks for working on this.

tools/sortKeymap.py
RDMurray: I have added named groups to make it clearer. basically the aim is to keep actions at the same position in the file when the shortcuts are changed.

RDMurray:
@jcsteh  I added an option --test which will fail with exit code 1 if the keymap isn't already sorted.  I suppose that should be easy enough to run from Appvayer, but I haven't looked into exactly how to do it yet.

@leonardder  \n is translated to the correct platform line endings when writing the file.  I tried os.linesep but it translates to \r\r\n in the file.  

I haven't yet verified that utf-8 is the correct encoding.

PR #676, opened 2022-02-10
Improve feedback for item movement commands

### LINKED ISSUE #155: Improve action feedback for nudging items/envelope points, and growing/shrinking item edges
Currently, when nudging items/envelope points and growing/shrinking item edges, action feedback reports a summary of the action title - "shrink left edges of items" or "move items/envelope points right", for example. However, this seems like it duplicates functionality already made available via shortcut help, and it doesn't report positional information.

Can this feedback be tweaked to report changes to positional information as the item edge grows/shrinks or the envelope point moves? I'm imagining hearing spoken feedback in the same format as scrubbing.

If so, would it also be possible to either A) add check boxes to OSARA prefs to toggle action feedback for each of the relevant categories, or B) reword the existing check box that toggles feedback when scrubbing to be something like "Report position on timeline when scrubbing, nudging, growing and shrinking"? The jury is still out on whether a combined check box would be popular among users, but figured I'd include the idea here because I know you're keen to avoid clutter (and also, seeing as the existing check box is the default position when opening the prefs dialog, I'm sure it'd be something the troops could learn to live with).

One thing I haven't figured out is how positional information would be reported productively if multiple items or points were selected. It's a brain-breaker. Any thoughts/suggestions much appreciated.

FYI, the proposed change would apply to the following actions:
Item edit: Grow left edge of items (40225)
Item edit: Grow right edge of items (40228)
Item edit: Shrink left edge of items (40226)
Item edit: Shrink right edge of items (40227)
Item edit: Move items/envelope points left (40120)
Item edit: Move items/envelope points right (40119)
Item edit: Move items/envelope points left by grid size (40793)
Item edit: Move items/envelope points right by grid size (40794)
Item edit: Move items left, preserving timing of contents (40121)
Item edit: Move items right, preserving timing of contents (40122)
Time selection: Nudge left (40039)
Time selection: Nudge right (40040)
Time selection: Nudge left edge left (40320)
Time selection: Nudge left edge right (40321)
Time selection: Nudge right edge left (40322)
Time selection: Nudge right edge right (40323)

I realise that's a boat load of actions and that it might not be feasible with the time you currently have, but thanks in advance for anything you can do here.

RDMurray:
reports the new position for the commands:
- Item edit: Move items/envelope points right
- Item edit: Move items/envelope points left
- Item edit: Move items/envelope points right by grid size
- Item edit: Move items/envelope points left by grid size

fixes #155

### DIFF
--- a/src/envelopeCommands.cpp
+++ b/src/envelopeCommands.cpp
@@ -657,3 +657,40 @@ void postSelectMultipleEnvelopePoints(int command) {
 		translate_plural("{} point selected", "{} points selected", count),
 		count));
 }
+
+void cmdMoveSelEnvelopePoints(Command* command) {
+	TrackEnvelope* envelope;
+	double offset{0.0};
+	tie(envelope, offset) = getSelectedEnvelopeAndOffset();
+	if(!envelope || !currentEnvelopePoint || !shouldReportTimeMovement()) {
+		Main_OnCommand(command->gaccel.accel.cmd, 0);
+		return;
+	}
+	double oldPos {0.0};
+	GetEnvelopePointEx(envelope, currentAutomationItem, *currentEnvelopePoint, &oldPos, nullptr, nullptr, nullptr, nullptr);
+	oldPos += offset;
+	Main_OnCommand(command->gaccel.accel.cmd, 0);
+	double newPos{0.0};
+	GetEnvelopePointEx(envelope, currentAutomationItem, *currentEnvelopePoint, &newPos, nullptr, nullptr, nullptr, nullptr);
+	newPos += offset;
+	ostringstream s;
+	if(lastCommand != command->gaccel.accel.cmd) { 
+		const char* name = kbd_getTextFromCmd(command->gaccel.accel.cmd, nullptr);
+		const char* start;
+		// Skip the category before the colon (if any).
+		for (start = name; *start; ++start) {
+			if (*start == ':') {
+				name = start + 2;
+				break;
+			}
+		}
+		s<<name << " ";
+		resetTimeCache();
+	}
+	if(oldPos == newPos) {
+		s << translate("no change");
+	} else {
+		s << formatTime(newPos, TF_RULER, false, true);
+	}
+	outputMessage(s);
+}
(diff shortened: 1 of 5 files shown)


src/envelopeCommands.cpp
```diff
@@ -657,3 +657,40 @@ void postSelectMultipleEnvelopePoints(int command) {
 		translate_plural("{} point selected", "{} points selected", count),
 		count));
 }
+
+void cmdMoveSelEnvelopePoints(Command* command) {
+	TrackEnvelope* envelope;
+	double offset{0.0};
```
jcsteh, line 663: nit: Out of curiosity, why did you choose to 0 initialise offset here but not envelope (nullptr)? Neither of these needs to be initialised because they are both immediately initialised below, but I can understand playing it safe... if that were consistent.

Anyway, please initialise both or don't initialise both. I'm happy either way. :)

src/envelopeCommands.cpp
```diff
@@ -657,3 +657,40 @@ void postSelectMultipleEnvelopePoints(int command) {
...
+		const char* name = kbd_getTextFromCmd(command->gaccel.accel.cmd, nullptr);
+		const char* start;
+		// Skip the category before the colon (if any).
+		for (start = name; *start; ++start) {
+			if (*start == ':') {
+				name = start + 2;
+				break;
+			}
+		}
+		s<<name << " ";
```
jcsteh, line 687: We now have this action name + category stripping code in three places. I think it'd be good to unify this. Perhaps we could move the bulk of reportActionName into an appendActionName, outputActionName, formatActionName or similar which takes an ostringstream reference. Or we could have getActionName which returns a string, but that seems like wasteful copying given that we're always going to output to an ostringstream anyway.

jcsteh (changes requested):
Nice! Thanks.

src/envelopeCommands.cpp
RDMurray: I changed reportActionName to getActionName, returning const char*.  Is that okay?

jcsteh (approved):

PR #681, opened 2022-02-17
When watching gain reduction, don't assume TrackFX_GetNamedConfigParm will succeed.

### LINKED ISSUE #680: On Mac, removing an effect when monitoring Gain Reduction with Peak Watcher crashes Reaper
1. Create a track, instantiate ReaComp on it.
2. From the FX chain list, hit Option+W to load Peak Watcher, configure either 1st or 2nd watcher to monitor gain reduction, hit OK.
3. Still in the FX chain list, hit Delete to remove ReaComp. Reaper will crash.
4. Repeat the above without Peak Watcher in the equation. Reaper stays open.

Unfortunately this has been reproduced on two machines, different versions of Mac OS etc. I'm gonna label it P1.

jcsteh:
This hopefully fixes a crash on Mac when the effect being watched is removed. However, I really have no idea; it's completely untested.
Fixes #680.

### DIFF
--- a/src/peakWatcher.cpp
+++ b/src/peakWatcher.cpp
@@ -320,8 +320,10 @@ const LevelType LEVEL_TYPES[] = {
 			const TrackFx* tfx = get_if<TrackFx>(&watcher.target);
 			assert(tfx);
 			char text[10];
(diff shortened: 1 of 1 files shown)


ScottChesworth:
Got it in one. Thanks!

PR #684, opened 2022-02-17
Add feedback for commands that change preview volume in the media explorer

### LINKED ISSUE #430: bind shortcuts to Media explorer actions Increase/decrease preview volume and provide feedback
These actions were added in reaper 6.19 and could be useful.
They are:
- Preview: increase volume by 1 dB; 42177
- Preview: increase volume by 0.1 dB; 42179
- Preview: decrease volume by 1 dB; 42178
- Preview: decrease volume by 0.1 dB; 42180

Not sure which ones would be most useful, but you can decide that.

RDMurray:
There is no API for this, so it gets the text from the GUI.  Only tested on Windows, but the functions used all exist in swell; let's see what happens.

fixes #430

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1825,6 +1825,15 @@ void postToggleTakePreservePitch(int command) {
 		translate("disabled preserve pitch when changing item rate"));
 }
 
+void postMExplorerChangeVolume(int cmd, HWND hwnd) {
+	HWND w = GetDlgItem(hwnd, 1047);
+	char text[50];
+	if(!GetWindowText(w, text, ARRAYSIZE(text))) {
+		return;
+	}
+	outputMessage(text);
+}
+
 typedef void (*PostCommandExecute)(int);
 typedef struct PostCommand {
 	int cmd;
@@ -2068,6 +2077,13 @@ PostCustomCommand POST_CUSTOM_COMMANDS[] = {
 	{"_XENAKIOS_NUDGEITEMVOLUP", postChangeItemVolume}, // Xenakios/SWS: Nudge item volume up
 	{NULL},
 };
+
+using MExplorerPostExecute = void (*)(int, HWND);
+map<int, MExplorerPostExecute> MExplorerPostCommands{
+	{42178, postMExplorerChangeVolume }, // Preview: decrease volume by 1 dB
+	{42177, postMExplorerChangeVolume}, // Preview: increase volume by 1 dB
+};
+
 map<int, PostCommandExecute> postCommandsMap;
 map<int, string> POST_COMMAND_MESSAGES = {
 	// translate firstString begin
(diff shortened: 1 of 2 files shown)


Lo-lo78:
@RDMurray Yes, it works well.
There are other interesting actions:
Preview: adjust pitch by +01 semitones
Preview: adjust pitch by -01 semitones
Very important:
Preview: reset pitch
Very strange there is no volume reset, it is only a matter of time, in the future perhaps Reaper will also add other actions.
👍

RDMurray:
@Lo-lo78 We could probably support the pitch and rate actions in the same way.  If you really need to reset the volume to 0.0, you can double click the volume slider.

RDMurray:
This works on both Windows and Mac.

src/reaper_osara.cpp
```diff
@@ -4238,6 +4255,16 @@ bool handlePostCommand(int section, int command, int val=0, int valHw=0,
 			isHandlingCommand = false;
 			return true;
 		}
+	} else if(section == MEDIA_EXPLORER_SECTION) {
+		const auto it = MExplorerPostCommands.find(command);
+		if(it != MExplorerPostCommands.end()){
+			isHandlingCommand = true;
+			SendMessage(hwnd, WM_COMMAND, command, 0);
```
jcsteh, line 4262: Nice. I didn't realise you could trigger actions with WM_COMMAND, but it makes sense. I really thought there was an API function to run a command in an arbitrary section, but apparently there isn't!

src/reaper_osara.cpp
```diff
@@ -2068,6 +2078,13 @@ PostCustomCommand POST_CUSTOM_COMMANDS[] = {
 	{"_XENAKIOS_NUDGEITEMVOLUP", postChangeItemVolume}, // Xenakios/SWS: Nudge item volume up
 	{NULL},
 };
+
+using MExplorerPostExecute = void (*)(int, HWND);
+map<int, MExplorerPostExecute> MExplorerPostCommands{
```
jcsteh, line 2083: nit: The variable name should start with a lower case letter to differentiate variables from custom types:
```suggestion
map<int, MExplorerPostExecute> mExplorerPostCommands{
```

src/reaper_osara.cpp
```diff
@@ -4238,6 +4255,16 @@ bool handlePostCommand(int section, int command, int val=0, int valHw=0,
 			isHandlingCommand = false;
 			return true;
 		}
+	} else if(section == MEDIA_EXPLORER_SECTION) {
+		const auto it = MExplorerPostCommands.find(command);
+		if(it != MExplorerPostCommands.end()){
```
jcsteh, line 4260:
```suggestion
		const auto it = mExplorerPostCommands.find(command);
		if(it != mExplorerPostCommands.end()){
```

jcsteh (approved):

PR #694, opened 2022-02-25
Windows: Use the keyboard hook instead of key map bindings to bind shift+f10 to REAPER context menus.

### LINKED ISSUE #689: OSARA intercepting Shift+F10 when plugin controls have focus
STR:

- Add the latest version of Surge XT to a track
- In the plugin interface, tab to any slider and press Shift+F10

Expected: The plugin's internal context menu should open.
Actual: Reaper's track context menu is opened instead, though it can't be navigated with the arrow keys.

This issue only comes up on Windows since Mac uses different keys for the menus. This problem just came up on the Surge Discord where a user couldn't get SHift+F10 to work, but for some other people it worked fine. I did some testing and narrowed it down to OSARA after reverting my keymap to default. If you assign OSARA's "Show first context menu" action to something other than Shift+F10 which was the case for me, then this issue doesn't occur.

jcsteh:
This was interfering with context menus in FX plugin UIs, since REAPER was handling shift+f10 according to the key map.
Fixes #689.


jcsteh:
I'd appreciate additional testing here. Aside from just making sure this achieves its intent, getting the logic right for this was a bit mind boggling, so it'd be good to double check I didn't screw up.

jcsteh:
Zarqing hell. Looks like [the latest version of Visual Studio broke linking with lld-link](https://developercommunity.visualstudio.com/t/lots-of-lld-link-warnings-about-pdb-with-vs2022-17/1667797).

ScottChesworth:
Been running this PR for the past couple of days, using Shift+F10 instead of Apps key as often as I can remember to, and I haven't run into any issues yet. Are there any areas in particular that I should be poking at?

jcsteh:
Nothing in particular. I imagine any issues would be caught in daily usage. Thanks. The only other thing would be to try using the apps key again for a bit, as this does touch that code path as well.

ScottChesworth:
There have been plenty of times where I've forgotten to do Shift+F10, so I've kinda been testing both. I reckon merge whenever you're ready.

jcsteh:
Blah. I screwed up that merge. Going to have to force push.

Issue #697, opened 2022-03-01
Can OSARA unlock non-contiguous selection in the actions list using its key hook?


ScottChesworth:
@RDMurray discovered that it's semi-possible to make non-contiguous selections in the actions list dialog if users unbind Control+UpArrow, Control+DownArrow and Space. Semi-possible as in, Control+Space still can't be used, this seems to be handled by Reaper inside dialogs.

Given that we're talking about widely adopted actions, IMO it's not reasonable to have to unbind and rebind actions every time a user wants to backup bindings that they've customized. Could OSARA hack around this using its key hook when the actions list is open? I've already had some discussion with Cockos around this and Justin's hands seemed to be tightly tied, but would be happy to CC in anyone with deeper technical knowledge if that's a better way to go.

Thanks in advance for any advice/implementation.

PR #698, opened 2022-03-01
Make ctrl+up, ctrl+down and ctrl+space work in list views

RDMurray:
uses the keyboard hook to pass these keys directly to the control.  Allows multiple noncontiguous selection in the action list.

fixes: #697

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -2397,7 +2397,8 @@ LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
 		wParam != VK_APPS && wParam != VK_CONTROL &&
 				wParam != VK_F10 && wParam != VK_RETURN &&
 				wParam != VK_F6 && wParam != 'B' &&
-				wParam != VK_TAB && wParam != VK_DOWN)) {
+				wParam != VK_TAB && wParam != VK_DOWN &&
+				wParam != VK_UP && wParam != VK_SPACE)) {
 		// Return early if we're not interested in the key.
 		return CallNextHookEx(nullptr, code, wParam, lParam);
 	}
@@ -2494,6 +2495,12 @@ LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
 		if (maybeOpenFxPresetDialog()) {
 			return 1;
 		}
+	} else if(control && 
+		(wParam == VK_DOWN || wParam == VK_UP || wParam == VK_SPACE)
+		&& isListView(focus)
+	) {
+		SendMessage(focus, WM_KEYDOWN, wParam, lParam);
+		return 1;
 	}
 	return CallNextHookEx(nullptr, code, wParam, lParam);
 }


ScottChesworth:
All smooth sailing here so far. Control+Space, Control+UpArrow and Control+DownArrow all behave as per usual everywhere except in the list box, where they now switch to behaving as expected in a standard list box. This is wicked!
The only inconsistency I've spotted so far is that Control+Space is play/stop when focused in the search filter, but play/pause everywhere else in the actions list. That's probably a Reaper quirk though? Not worth sweating over IMO.

ptorpey:
I tested this but it didn’t seem to work.  Am I missing something?  Here is what I did:

 

1.	Get into the Actions list
2.	Hit control+DownArrow and then space to select an action.  The action is reported as “selected”
3.	While holding the control key down, I hit downArrow to get to another action.
4.	Hit space to select an action.
5.	Repeated holding control down while hitting the downArrow key to move through items and occasionally selecting one by hitting space.  Each time JAWS said “selected” after hitting space.
6.	Tabbed to the Hotkeys button and used the “Export Selected Hotkeys” option.
7.	Reaper said that the hotkey map was empty and, indeed, when editing the saved keymap it was blank.

Did I misunderstand and do something wrong?

 

--Pete

 

8.

ScottChesworth:
Sounds like you're pressing the right things. Can you try the same procedure with NVDA and let us know how that goes? I'm thinking JAWS' keyboard hook might be complicating things here, Jim might need to do a bit of extra scriptage.

jcsteh:
For what it's worth, JAWS being the problematic factor here seems very
unlikely to me. Still, it's worth a shot just to narrow the variables.

RDMurray:
@ptorpey is it possible that you only selected actions that hadn't been customized?  I just managed to do that here and it results in an empty file.  Try again, only selecting customized actions.  The other thing is that I didn't make it work with space, only with Control+Space.  Do people use space on its own for this?  I only learned today that space adds to the selection, I had always used Control+Space which toggles selection of the focused item.  Space could be added if it not working will confuse people.

ptorpey:
Good catch.  It is more than likely that I didn’t pick custom actions.  I’ll give this another shot with JAWS (and NVDA if JAWS doesn’t work) and report back.

 

--Pete

RDMurray:
I don't just mean custom actions, although those should work, but actions that have a key assigned that is different to the reaper default.  Default bindings are not saved in the keymap file.

ptorpey:
@RDMurray

Testing with JAWS, non-contiguous selection with control+space and arrowing up and down with up/down keys and control held down works as you originally described.  Seems like we can now create our own partial custom keymaps for importing into the more complete keymap without the extra work of saving one keymap at a time and editing them all together into one file.

 

Nice job!

 

BTW, I don’t have a strong opinion about space if selections can be made now with control+space.  Yes, it could be a bit non-intuitive since hitting space in most other contexts will select a single item in a list.  Currently hitting space in the Actions dialog starts and stops playback.  Perhaps a note in the help documentation would work since the only way I would have known about this is by following this thread.

 

Thanks again.

 

--Pete

ptorpey:
For the test I used a few of the Dry Redner options which I had assigned to keys that weren’t previously assigned by Osara or Reaper.

 

--Pete

jennykbrennan:
Any thoughts on how this should work on the Mac?  I can select with shift arrows and export successfully. But non contiguous is not happening here. I've tried what I've read here but the only thing I accomplish is adding selection with shift and up and down arrows. I'm not sure what I can try. Ctrl seems to clear selection. Space and control space does what they say they do, until I unbind them. Then they give me the little error ding. no matter how I try to use them.

jennykbrennan:
That was the first thing I tried. It appears to work but when I try to export, there was nothing selected. I will give it another try and see if maybe I tried to export non custom actions. Come to think of it, i do need to check on that.  I'll be back. :)

jennykbrennan:
yep. I actually had the thought several times   that it should work that way on mac. But  i never did get it to work. Voiceover says all the right things even when you don't actually interact with the table and that is just a touch misleading. Because why try a different way when it clearly works according to Voiceover. LOL. Should have known better, i know. Clunky as it is, it is very much doable. But with that said, i think  a guide  on how to actually do this on Mac is  probably needed. I'm fairly confident  on the mac and this one tripped me up. How will a newbie on the mac figure this out I wonder. I'll get one done, if there aren't anything readily available to find. Since I said it, I might as well follow up.  And with that, I'll leave you to it.  Cheers. :)

jcsteh:
This change only works on Windows, not Mac. That will need to be explicitly noted somewhere.

Lo-lo78:
@RDMurray 
OPTIMAL! It works perfectly.
👍

Lo-lo78:
Sorry, I use Windows 10 with NVDA, I know that on Mac the situation is a little different.
I forgot to mention the operating system.

Lo-lo78:
@RDMurray @jcsteh 
I think OSARA should announce "unselected" before Reaper's action description.
In many cases, Reaper's actions have long names and you have to wait to get the situation clear.
If OSARA announced the selected or the unselected right away it would be much faster to extract actions to share or to put in a backup.
I don't know if it's possible but doing so would have a lot of value and would be much easier to use.
My specific problem is that I use action IDs a lot and the atesa doubles up, waiting a long time to know if a certain action is selected or not.
Greetings!

jcsteh:
This is a standard list view control, so unfortunately, where selected gets
announced is controlled entirely by the screen reader. OSARA can't (and
shouldn't try to) affect that.

Having said that, here are some tips that might help:

1. If you haven't selected an action yet and you press control+downArrow,
it will always be unselected.
2. You can press control+space twice quickly to work out whether it's
selected or unselected.

Lo-lo78:
Yes, I knew this aspect of Windows, it doesn't depend on Reaper, NVDA or 
OSARA; it is Windows that has this behavior.

I was hoping you could get over it.

Greetings!Message ID: ***@***.***>

jcsteh:
It's something that could be tweaked in NVDA - it's probably even been
reported already - but it's not trivial for a bunch of reasons.

PR #701, opened 2022-03-04
Pass global shortcuts through to the main window in the params dialog…

RDMurray:
…, even when the combobox has focus.


RDMurray:
@ScottChesworth here it is if you want to test.  Just set any shortcuts to global in the action list and they should work in the params dialog.

jcsteh:
I've merged #702 now, which allows most keys to pass through when not focused on the combo box. This isn't necessarily mutually exclusive with this PR, though. I still disagree with global (non-text fields) bindings working in combo boxes, but if that's the behaviour REAPER has everywhere else, we probably should be consistent.

PR #702, opened 2022-03-05
Parameters dialog: Pass most keys through to REAPER's main window.

### LINKED ISSUE #784: Is it possible to move from one marker to another in the OSARA FX Parameters?
Hi everyone!
It would be very helpful if a solution could be found.
It is not important to have the numbers from 0 to 9 as it would be that with the tab you go to the progress box and with the two keys of the keymap you could move between the markers.
Obviously I was thinking about actions:

Markers: Go to next marker/project end
Markers: Go to previous marker/project start

I find it impractical to exit the FX Parameters and change markers from the tcp every time to move around the project.
Greetings!

jcsteh:
Previously, we only passed specific keys through to the main window.
Now, we do the inverse, capturing only specific keys for the Parameters dialog and passing the rest through to the main window.
The combo box and text boxes are still exceptions, where we only pass specific keys to the main window to ensure the user can interact with the control properly.
Practically speaking, this means users can access many more main window actions when focused on the slider or buttons.

This is yet another take on #700 and #701. It's less permissive than those in that it still doesn't allow marker navigation while in the combo box, which I think is a risky slippery slope. On the flip side, it doesn't require global bindings.

@ScottChesworth , @RDMurray , thoughts appreciated.

Fixes #784.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -533,19 +533,31 @@ class ParamsDialog {
 			// Let REAPER handle the space key so control+space works.
 			return 0; // Not interested.
 		}
-		if (
-			// A function key.
-			(VK_F1 <= msg->wParam && msg->wParam <= VK_F12) ||
-			// Anything with both alt and shift.
-			(GetAsyncKeyState(VK_MENU) & 0x8000 &&
-				GetAsyncKeyState(VK_SHIFT) & 0x8000) ||
-			// Anything with the control key, but only if not in a text box.
-			(!isClassName(GetFocus(), "Edit") &&
-				GetAsyncKeyState(VK_CONTROL) & 0x8000)
-		) {
-			return -666; // Force to main window.
+		if (msg->hwnd == dialog->paramCombo ||
+				isClassName(GetFocus(), "Edit")) {
+			// In text boxes and combo boxes, we only allow specific keys through to
+			// the main section.
+			if (
+				// A function key.
+				(VK_F1 <= msg->wParam && msg->wParam <= VK_F12) ||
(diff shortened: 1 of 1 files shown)


jcsteh:
@ScottChesworth, do you think this is still useful even with the change to global actions? It should certainly mean we need less global bindings.

ScottChesworth:
I think so yeah, but haven't had time to use it much in situ yet. Don't wanna keep it hanging around for long, but seems like a week of testing might be wise given that it's a fairly significant shift in logic.

jcsteh:
@ScottChesworth , any further thoughts on this?

jcsteh:
@ScottChesworth , another friendly nag for your thoughts here. :)

Lo-lo78:
Hague! I searched but the title of the issue did not make me realize that it was already open.

Lo-lo78:
It's perfect!
I don't know what the problem is for the dialog.
Maybe you want it to always work between parameters too?
It is very good!

Lo-lo78:
James I think you can merge. and take the snapshot.
(rofl)

ScottChesworth:
Argh, not sure how I missed the nags above, sorry @jcsteh. If you're confident that this is solid then of course go ahead and merge. If you'd still like more feedback, I'll start banging away on it and report back tomorrow.

jcsteh:
I'm reasonably confident this should be fine, so I'll just merge. If it causes problems, it's small enough that reverting it isn't a problem.

Lo-lo78:
I am trying it and it gives me no problems.
Great Scott for the idea

Issue #534, opened 2021-05-27
Enhancement suggestion: Change parameter list from a combo box to a list view


pitermach:
This is something I thought of while bouncing between the 2 operating systems more. The current setup works OK, but I think this would considerably improve efficiency, especially on Mac but also on Windows to an extent.

Pop-up buttons on Mac work quite differently to Windows combo boxes. The main difference is that, on Windows, pressing the arrow keys while on a combo box changes the selection even though the list doesn't open visually. On Mac, this isn't the case. Your first press of either arrow key first causes the menu of selections to appear, and only subsequent arrow presses move through the choices. Something else to note here is that while the menu is open, you can't tab, so you have to press enter on a parameter before you can tab to its value slider.

On Windows, the benefit though less pronounced, would be that unlike a combo box, a list view lets you do searches for more than the first letter of an item, which ironically can be done in a Mac popup menu even now. The filter box does kind of fill this role, but that ends up taking more keystrokes to focus it, do a search and then return to the list which isn't as practical if you want to adjust just one or two parameters.

ns-studios:
I think this would be very useful, and something I thought of suggesting myself seeing as the preset dialog uses the listview. +1

pitermach:
Superseeded by #1429 which is even better because it now supports categories.

PR #707, opened 2022-03-07
When the Parameter combo box in the Parameters dialog gets focus, rebuild the parameter list.

### LINKED ISSUE #706: Parameters list not refreshed when parameter changes affect other parameters
Don't know whether this is worth addressing or even possible to fix, but, as Scott says, maybe someone has a clever solution.

here is the scenario using ReaEq as an example:
1.  Add the ReaEQ plug-in to a track.
2. In the parameters list change band 1 type to something like "notch".  Note: The default is "Low Shelf".
3. Arrow down to the Frequency for band 1.  Osara announces "Frequency Low Shelf 1".  In actuality, since the band 1 filter has changed, this should say "Frequency Notch 1".

Summary: When one parameter in the list is changed that affects another parameter in the list, the list is not accordingly updated to speak the proper info.

I don't know how many other plug-ins or parameters this might affect.

Fortunately the problem is easily rectified by closing the parameters list and re-opening it again.  Then all parameters provide correct feedback.

In terms of a possible fix, it probably isn't a good idea to refresh the entire list each time a parameter is changed since one might be changing a parameter by hitting the up/down key quickly several times and we don't want to slow down this operation.  An alternative might be to update the list if (a)a parameter has changed and (b)one has tabbed or shift+tabbed from the slider to the parameter list itself.

--Pete

jcsteh:
This deals with cases where changing a parameter value also changes parameter names; e.g. changing the bad type in ReaEq.
Fixes #706.

This is a fairly naive solution in that it doesn't bother to keep track of whether a parameter value has actually changed. Thus, it rebuilds the parameter list more often than it strictly needs to. That said, unless this actually causes performance problems for plugins with large numbers of parameters, I'd prefer not to complicate the code here. It should be possible if this does cause problems, though.

Either way, this needs testing on plugins with large numbers of parameters.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -298,6 +298,7 @@ class ParamsDialog {
 	unique_ptr<ParamSource> source;
 	HWND dialog;
 	HWND paramCombo;
+	WNDPROC paramComboOrigProc;
 	HWND slider;
 	HWND valueEdit;
 	HWND valueLabel;
@@ -459,6 +460,20 @@ class ParamsDialog {
 				}
 		}
 		return FALSE;
+	}
+
+	static INT_PTR CALLBACK wndProc(HWND hwnd, UINT msg, WPARAM wParam,
+		LPARAM lParam
+	) {
+		ParamsDialog* dialog = (ParamsDialog*)GetWindowLongPtr(GetParent(hwnd),
+			GWLP_USERDATA);
+		if (msg == WM_SETFOCUS && dialog && hwnd == dialog->paramCombo) {
+			// Changing a param value might update param names; e.g. changing the
+			// band type in ReaEq.
+			dialog->updateParamList();
+			return 0;
+		}
+		return CallWindowProc(dialog->paramComboOrigProc, hwnd, msg, wParam, lParam);
 	}
 
 	accelerator_register_t accelReg;
@@ -634,6 +649,8 @@ class ParamsDialog {
 		SetWindowText(this->dialog, this->source->getTitle().c_str());
 		this->paramCombo = GetDlgItem(this->dialog, ID_PARAM);
 		WDL_UTF8_HookComboBox(this->paramCombo);
+		this->paramComboOrigProc = (WNDPROC)SetWindowLongPtr(this->paramCombo,
+			GWLP_WNDPROC, (LONG_PTR)this->wndProc);
 		this->slider = GetDlgItem(this->dialog, ID_PARAM_VAL_SLIDER);
 		// We need to do exotic stuff with this slider that we can't support on Mac:
 		// 1. Custom step values (TBM_SETLINESIZE, TBM_SETPAGESIZE).


jcsteh:
Oh look. It failed on Mac. What a Zarqing surprise. No idea what to do about that. SWELL doesn't support WM_SETFOCUS either, so I can't use that either.

leonardder:
> ... SWELL doesn't support WM_SETFOCUS either

Looks like that assumption is no longer correct. See https://github.com/justinfrankel/wdl/commit/20057d485a4ee345fb8db6c846f124672b300f7c
It will of course require a bump of swell/wdl.

jcsteh:
Oh. I forgot I need to change this to use WM_SETFOCUS instead of CBN_SETFOCUS. That's going to require implementing a WindowProc for the combo box, which is annoying, but should be possible.

jcsteh:
Would someone be able to test this on Mac? See https://github.com/jcsteh/osara/issues/706#issue-1160737802 for testing instructions.

jcsteh:
Oh, and it still needs testing with plugins with a large number of parameters, either on Windows or Mac.

pitermach:
Just did some testing on both Windows and Mac with SurgeXT. It's both a plugin that has a lot of parameters (about 2800 for the VST3) and dynamically changes their names. FOr the test I located the FX A1 FX type parameter and changed it from off to one of the effects, which causes the next parameters to change based on the selected effect.

- On Windows, the names change correctly but there's now some lag when the combo box gets focused. With NVDA and ZDSR it's about a second, with Narrator it's a bit less but still noticeable.
- On Mac, there is no lag when you tab into the popup, but the names in the list don't update. You do hear the new names if you use the CTRL+Tab shortcut however.

Would this be any easier to do if #534 were implemented? I suspect it might introduce even more headaches but I figured I'd throw it out there ☺️

jcsteh:
> On Windows, the names change correctly but there's now some lag when the combo box gets focused. With NVDA and ZDSR it's about a second, with Narrator it's a bit less but still noticeable.

Yeah, that's what I was afraid of. That's OSARA trying to fetch all the parameter names again.

> On Mac, there is no lag when you tab into the popup, but the names in the list don't update.

Ug. Terrific. That means the code isn't actually working at all, probably because the focus notification doesn't work properly on Mac even though it now should.

> Would this be any easier to do if #534 were implemented?

No; I suspect the lag and focus failure is not related to the type of control.

jcsteh:
Thanks for your testing regardless.

jcsteh:
Actually, I guess it's possible the lag is accessibility events being fired. Hmm.

ptorpey:
Sorry if this is a duplicate post, but I don’t think my last message got to the list.  Here it is again:

 

I don’t know if this is an Osara issue or related to how the JAWS scripts are handling this, but from the sound of this issue, I thought this fix would have addressed the curious behavior I am seeing.

 

I was using the TriLeveler plug-in and noticed that, as I changed parameter values in Osara’s parameter list, I was getting non-real world values.  When focus was first on the value of a parameter, however, the values were real world.

 

Specifically, here is what happens:

 

1.	Hit “p” to get into the Osara parameters list for the TriLeveler plug-in on the track.
2.	Tab to the first value, trim, in the list and see that it says something like 2.5 db.
3.	3.  Hit PageDown and the value changes to something like 0.57%.
4.	Shift+tab back to the list of parameters, arrow to another parameter, and then arrow back to the trim parameter.
5.	Tab to the v value and now I hear the real world value of 2.1 db instead of something like 0.57%.

 

Thus, as I change parameter values, they are changing but are not being reported correctly.

 

If you think it is a JAWS scripts issue I can pass it along to Jim to see what he can do.

 

Thanks.

 

--Pete

ptorpey:
I don’t know if this is Osara issue or related to how the JAWS scripts are handling this, but from the sound of this issue, I thought this fix would have addressed the curious behavior I am seeing.

 

I was using the TriLeveler plug-in and noticed that, as I changed parameter values in Osara’s parameter list, I was getting non-real world values.  When focus was first on the value of a parameter, however, the values were real world.

 

Specifically, here is what happens:

 

1.	Hit “p” to get into the Osara parameters list for the TriLeveler plug-in on the track.
2.	Tab to the first value, trim, in the list and see that it says something like 2.5 db.
3.	3.  Hit PageDown and the value changes to something like 0.57%.
4.	Shift+tab back to the list of parameters, arrow to another parameter, and then arrow back to the trim parameter.
5.	Tab to the v value and now I hear the real world value of 2.1 db instead of something like 0.57%.

 

Thus, as I change parameter values, they are changing but are not being reported correctly.

 

If you think it is a JAWS scripts issue I can pass it along to Jim to see what he can do.

 

Thanks.

 

--Pete

Issue #708, opened 2022-03-07
Add Feedback For View: Toggle snap to grid In The MIDI Editor


MatejGolian:
Please, add feedback for the MIDI editor action 'View: Toggle snap to grid'. The action ID is 1014.
Making sure this action is in the desired state is useful, when you want to split notes, but you want to make sure that they get split exactly where your cursor is and not right on the grid. There may be other scenarios where this info comes in handy, but this is my use case.

ScottChesworth:
This action doesn't show its state in the actions list. Can we report its state? If so, I don't know how.

MatejGolian:
What exactly does it mean that it doesn't show its state?
I ask because, when I open the actions list, I either see
'Alt+S; View: Toggle snap to grid; on'
or
'Alt+S; View: Toggle snap to grid; off'.
I'm not arguing, I'd just like to understand the issue here.

MatejGolian:
@RDMurray, if I get a vote, I'm in on your suggestion. Provided that it's not that hard to implement, of course. Two questions come to mind though:
1. Out of curiosity, I imagine that your solution would only really make sense in case of actions where reporting whether they are on or off is sufficient. How difficult is it to find out if a given action is just a simple on/off one? If it's possible to make all these actions talk, so to speak, without too much hassle, I think having it the way you propose is way better than doing it individually for each requested action, even if this could theoretically result in some messages being grammatically incorrect.
2. If we take my request as an example, what would the feedback be? "snap to grid on" and "snap to grid off"? If so, that would be great.

pitermach:
I like this idea, though if it gets implemented there should be a place in code to exclude actions if something gets too talkative. IE "Toggle loop segment scrub at edit cursor" is something you wouldn't want announced, but it's also the only one I can think of off the top of my head hence the suggestion.

jcsteh:
> Wouldn't it be a better idea to have feedback for every togglable action rather than implementing them one by one? It might result in slightly strange grammar, but it would be quite a big increase in accessibility without affecting existing functionality.

[...]

PR #712, opened 2022-03-11
Report the state of toggle actions not explicitly supported by OSARA.

jcsteh:
As suggested by @RDMurray in https://github.com/jcsteh/osara/issues/708#issuecomment-1060859048. I have pretty big reservations about this, but in the end, I thought it worth playing with.

For these actions, OSARA will report "enabled" or "disabled", followed by the action name.

Things worth noting:

1. Upon implementing this, I immediately noticed an annoyance: opening the Actions list reported "enabled Show actions list", interrupting the reporting of the dialog! Most of the actions to open/close windows in REAPER are toggle actions, which makes sense. However, because the window opens, this really isn't something we want: the window opening already provides sufficient feedback, plus the message interrupts important info from the window itself. I've added an exception for Show actions list, but I fear this will be a long road.
2. Stripping the category name (i.e. anything before a colon) doesn't work reliably. In the Media Explorer, we have "Autoplay: toggle on/off". Of course, with the category stripped, that resulted in a useless message like "enabled toggle on/off". So for this to be useful, we can't strip the category name, which probably increases verbosity to annoying levels in some cases.

Regardless, I'll leave this here for people to play with.

### DIFF
--- a/makePot.py
+++ b/makePot.py
@@ -73,29 +73,36 @@ def addMessage(data):
 		comments.extend(lastTranslatorsComment)
 		lastTranslatorsComment = []
 
-RE_CPP_TRANSLATE_FIRST_STRING_END = re.compile(r"^\s*// translate firstString end$")
-RE_CPP_TRANSLATE_FIRST_STRING = re.compile(r'^\s*[^/].*?"(?P<msgid>.*?)"')
-def addCppTranslateFirstString(input):
+RE_CPP_TRANSLATE_FIRST_N_STRINGS_END = re.compile(r"^\s*// translate first\d?Strings? end$")
+RE_CPP_COMMENT = re.compile(r'^\s*/')
+RE_CPP_STRING = re.compile(r'"(?P<msgid>.*?)"')
+def addCppTranslateFirstNStrings(input, maxStrings):
 	for line in input:
 		if handleTranslatorsComment(line):
 			continue
-		if RE_CPP_TRANSLATE_FIRST_STRING_END.match(line):
+		if RE_CPP_TRANSLATE_FIRST_N_STRINGS_END.match(line):
 			break
-		m = RE_CPP_TRANSLATE_FIRST_STRING.match(line)
-		if m:
+		if RE_CPP_COMMENT.match(line):
+			continue
+		for num, m in enumerate(RE_CPP_STRING.finditer(line)):
+			if num == maxStrings:
+				break
 			data = m.groupdict()
 			addMessage(data)
 
 RE_CPP_TRANSLATE = re.compile(r'\btranslate\("(?P<msgid>.*?)"\)')
 RE_CPP_TRANSLATE_CTXT = re.compile(r'\btranslate_ctxt\("(?P<context>.*?)",\s*"(?P<msgid>.*?)"\)')
 RE_CPP_TRANSLATE_PLURAL = re.compile(r'\btranslate_plural\("(?P<msgid>.*?)",\s*"(?P<plural>.*?)", .*?\)')
-RE_CPP_TRANSLATE_FIRST_STRING_BEGIN = re.compile(r"^\s*// translate firstString begin$")
+RE_CPP_TRANSLATE_FIRST_N_STRINGS_BEGIN = re.compile(r"^\s*// translate first(?P<maxStrings>\d)?Strings? begin$")
 def addCpp(input):
 	for line in input:
 		if handleTranslatorsComment(line):
 			continue
-		if RE_CPP_TRANSLATE_FIRST_STRING_BEGIN.match(line):
-			addCppTranslateFirstString(input)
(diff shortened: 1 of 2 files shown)


Lo-lo78:
@jcsteh 
It works well but is too verbose.
For example:
enabled Autoplay: Toggle on/off
Toggle on/off it should not be vocalized by the speech synthesis.
It is already spoken by the speech synthesis enabled Autoplay and disabled Autoplay.
You probably already know!
👍

ScottChesworth:
Hmmm, trouble is, Cockos aren't consistent about how they name actions, so we can't strip to improve verbosity with any consistency either.

MatejGolian:
I tested this build a little and as others have pointed out, this
results in more feedback being given than needed. Pity that it's not
safe to at least strip the category name from the output message. So I
have mixed feelings about this.
On the one hand it's a nice feature, but on the other hand, custom
tailored feedback is superior. That being said, couldn't it be useful
to keep this functionality as a fallback for those toggle actions that
Osara doesn't provide custom feedback for? Should someone find that a
particular action requires special addressing, they can still request
custom feedback for it. Or is this making things unnecessarily
complicated?
Of course, if this method is prone to produce unreliable results, it's
probably best to keep on doing things like before.

RDMurray:
It was my intention when I suggested this that it would just be a fallback
for toggles not supported by Osara yet; custom feedback would indeed be
preferable, especially for commands on the default key map.

MatejGolian:
@RDMurray, I see and sorry in case I misunderstood something before.

RDMurray:
Another related thought:  There could be a list of strings to provide custom feedback for toggle actions, like postCommandMessages, which would have "enabled" or "disabled" prepended to them.  It would make future implementation much easier.

ScottChesworth:
> There could be a list of strings to provide custom feedback for toggle actions, like postCommandMessages
> 
Ooh, I like that idea.

ScottChesworth:
So I've been using this more, and although I'm struggling to get used to the additional verbosity here, I reckon you've landed on something that'll be super useful to folks who are starting out. With that in mind, could this become a Config option? I think it makes sense to have it enabled by default, seeing as it exists to assist users who are still finding their feet.

jcsteh:
I can live with poor translatability for a fallback feature, but I don't think we should be prefixing explicit messages with "enabled" or "disabled" just to make it easier to add new messages. It really iso't good for translators.. Instead, we could create a map with two messages, one for on and one for off.

Why do we need an option for this? Can you give examples of actions you are using that this is providing extraneous feedback for? If this is bothering you, doesn't that suggest it is just poor UX across the board? I don't like the idea of creating options to compensate for crap UX.

RDMurray:
@jcsteh two messages per action would be fine. It would still reduce the friction for adding new feedback if they were in a map.

@ScottChesworth Where is it coming up in every day usage?  IMHO anything on the default keymap should have custom feedback implemented, and this should just be a fallback for people who want to customise their keymap.

ScottChesworth:
> Why do we need an option for this? Can you give examples of actions you are using that this is providing extraneous feedback for?
> 
To me, just enough verbosity to know what I need to know is preferable, so the answer to that is "er, practically all of them" lol.

> If this is bothering you, doesn't that suggest it is just poor UX across the board?
> 
No, because it's about the amount of reassurance that's useful. I'm using Reaper all day everyday in one way or another, have been doing that for long enough that I'm confident with most things, and perhaps more importantly, I have enough experience that I'm confident I'll get myself out of whatever sticky situations I end up in. The folks who I can see liking this extra feedback are newcomers and people who are using Reaper less frequently. It's difficult to find a middle ground between those needs, hence my suggestion that this would be optional. That said though...

> IMHO anything on the default keymap should have custom feedback implemented, and this should just be a fallback for people who want to customise their keymap.
> 
Sure, can see the logic in that. However, we're a fair way away from every mapped action talking at this point, so this being non-optional will introduce quite a bit of extra verbiage.

I'd like to see it be optional as above, but I think the benefits outweigh my reservations, so if that's gonna be a lot of work or we can't come to an agreement about it, then roll it out as is. I'll live. :)

jcsteh:
I've now added a map for messages. Specifying null messages means to report nothing. Adding support for a toggle action is now just one line of code:
`{{MAIN_SECTION, 40346}, {"full screen", "normal screen"}}, // Toggle fullscreen`

jcsteh:
I'm still not a massive fan of this due to the potential for user annoyance, but I think I've managed to mitigate at least some of those concerns. If nothing else, merging this will give us the ability to more easily add explicit support for toggle actions (1 line in a map instead of several lines of code scattered about). We can always remove the fallback if it becomes a real problem.

jcsteh:
That's the nature of this feedback, unfortunately. It's meant as a fallback
for when we don't have explicit support for actions, but it's never going
to be as nice as explicit support.

PR #713, opened 2022-03-11
Fix envelope related commands so that they work correctly on take envelopes where the play rate is not 1.0

RDMurray:
Take envelopes are scaled by take play rate.  

As of Reaper 6.51, there is a similar bug in the reaper built-in actions
- Envelope: Insert new point at current position (do not remove nearby points)
- Envelope: Insert new point at current position (remove nearby points)

"Envelope: Add/edit envelope point value at cursor" works correctly.

@Justinmac84 Please test the below build.

### DIFF
--- a/src/envelopeCommands.cpp
+++ b/src/envelopeCommands.cpp
@@ -23,45 +23,64 @@ using namespace fmt::literals;
 bool selectedEnvelopeIsTake = false;
 int currentAutomationItem = -1;
 
-// Returns the selected envelope and the offset for all envelope point times.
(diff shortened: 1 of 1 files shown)


jcsteh (approved):
Thanks for fixing this. Assuming the testing comes back positive, we're good to go here.

jcsteh:
NO feedback on testing, so I'm just going to merge this.

ScottChesworth:
I'll make sure Justin M knows a fix is available.

PR #726, opened 2022-03-20
On Windows, if the previous foreground window (e.g. dialog) closes, return focus to the foreground window before that.

### LINKED ISSUE #610: Can OSARA fix where focus lands after closing the Auto add FX dialog on Windows?
There's a disparity between where focus lands on Windows and Mac when users close the Auto add FX dialog. The Mac behaviour aligns with the visual UX.

1. Make sure that "Auto-open FX add dialog when opening empty track FX chain" is checked in the Plug-ins category of Reaper Prefs. Checked is its default state, so unfortunately the glitchy behaviour described below is what most Windows users will experience.
2. On Windows, hit F on a track. The Auto add dialog appears, all is well. 
3. Now hit Escape to dismiss the FX ad dialog. Visually, an empty FX window is still on screen, but keyboard focus goes back to the main window instead. Manual redirection of focus is required before users can paste plug-ins/load FX chains/access the FX menu bar, and users aren't being informed that the FX window is still open.
4. Comparing above steps on Mac, keyboard focus does land in the empty FX chain list properly on the first press of Escape when dismissing the Auto add dialog, then a second press of Escape closes the FX window proper. Ideally, this would also be true for Windows users.

I've already taken this to Cockos. Justin said:
> in Windows when you close a window the focus automatically goes to the owner window, which in this case is the main REAPER window. I'll look at if we can code up a solution around this, but it might be a big can of worms so maybe not too soon...
> 
Ergo, I'm raising it here to see whether we can rustle up a solution sooner with less technical headaches.

I've also noticed that reporting as I navigate through tracks becomes sporadic when that stray FX window is still on screen, so I'm gonna label this as a bug and P1 for now. Feel free to adjust accordingly if I'm over-egging it @jcsteh.

Any thoughts appreciated.

jcsteh:
This means that focus goes to the FX Chain when you open the FX Chain for a track with no FX and dismiss the Add FX dialog.
Similarly, in the Menu Editor, if you add an action and dismiss the Actions dialog, focus returns to the Menu Editor.
Fixes #610.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4393,8 +4393,28 @@ void annotateAccRole(HWND hwnd, long role) {
 		var);
 }
 
+HWND prevForegroundHwnd = nullptr;
+HWND prevPrevForegroundHwnd = nullptr;
+
 void CALLBACK handleWinEvent(HWINEVENTHOOK hook, DWORD event, HWND hwnd, LONG objId, long childId, DWORD thread, DWORD time) {
 	if (event == EVENT_OBJECT_FOCUS) {
+		HWND foreground = GetForegroundWindow();
+		if (foreground != prevForegroundHwnd) {
+			// The foreground window has changed.
+			if (IsWindowVisible(prevPrevForegroundHwnd) &&
+					!IsWindowVisible(prevForegroundHwnd)) {
+				// The previous foreground window has closed. For example, this happens
+				// when you open the FX chain for a track with no FX and dismiss the Add
+				// FX dialog. It also happens in the Menu Editor when you add an
+				// action and then dismiss the Actions dialog. REAPER returns focus to the
+				// track view in this case. We redirect focus to the foreground window
+				// before the last (the FX chain in the former example), which is more
+				// useful to the user.
+				SetForegroundWindow(prevPrevForegroundHwnd);
+			}
+			prevPrevForegroundHwnd = prevForegroundHwnd;
+			prevForegroundHwnd = foreground;
+		}
 		bool focusIsTrackView = isTrackViewWindow(hwnd);
 		if (focusIsTrackView) {
 			// Give these objects a non-generic role so NVDA doesn't fall back to


jcsteh:
This is a bit risky, so it should get some decent testing in real usage before we merge it.

Lo-lo78:
@jcsteh 
When I go to the options menu and select Customize menus / toolbars, OSARA ignores the window even if it opens.
On the other hand when I open the OSARA Cycle Action editor window it works much better.
It doesn't seem to work in all situations.
If I go to the action list and press the Menu editor button, the Customize menus / toolbars window opens perfectly. In this situation it works.
Greetings!

RDMurray:
This works great for those two cases.  I can't remember what the other one was, was it something to do with closing the parameters dialog?

What is the possible risk?  Is there anything in particular we should be testing?

@Lo-lo78 That is a separate problem that is not addressed by this PR.  For now the best way to get to the menu editor is through the action list.

jcsteh:
The other one was closing the OSARA Configuration dialog while in the MIDI
Editor, but I fixed that another way on both Windows and Mac. This fix need
not apply there, since we control the window ownership in that case.

The risk is that we apply this code when it's not appropriate. I can't
think of anything specific, but it's always possible that there's some
window where we don't want to do this for whatever reason. So, I don't have
anything specific in mind to test, except to say that areas of concern are
any possible cases where a dialog or foreground window change occurs.

Lo-lo78:
@jcsteh 
I made a shortcut with the action:
"Toolbars: Customize..."
Again the Customize menus / toolbars window is not focused.
With this build the only way to access this window is with the action list with the Menu editor button.
However, in many cases the build has greatly improved the focus of Reaper's windows.
Greetings!

jcsteh:
To clarify, are you saying this works correctly for you in normal snapshots but not with this build? I tested a normal snapshot earlier and couldn't get the menu editor to focus properly if I opened it from the Options menu there either. I didn't try binding an action, but I'm guessing I'd see the same behaviour. If you haven't tested this with a normal snapshot, please do and report back. Thanks.

Lo-lo78:
@jcsteh 
Yes, you are right with OSARA version 2022.1pre-896,0c8955c5 with a shortcut the Customize menus / toolbars window does not open.
Neither from the Options menu.
It opens only with the action list.
For me it's not a problem it's just a report.
Greetings!

jcsteh:
I'm going to merge this. We can always back it out if it causes problems in the wild.

PR #729, opened 2022-03-21
Report muted notes and CCs

### LINKED ISSUE #278: Can OSARA report if notes are muted in the piano roll?
Just had a befuddling 5 minutes in the MIDI Editor, and this feature would've avoided most of the head scratching. I reckon muted would be reported after note name and length, in keeping with the current format.

Thanks in advance if this is possible.

leonardder:
Closes #278

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -392,20 +392,23 @@ void previewNotes(MediaItem_Take* take, const vector<MidiNote>& notes) {
 	}
 	// Queue note on events for the new notes.
 	for (auto const& note: notes) {
+		if (note.muted) {
+			continue;
+		}
 		MIDI_event_t event = {0, 3, {
 			(unsigned char)(MIDI_NOTE_ON | note.channel),
 			(unsigned char)note.pitch, (unsigned char)note.velocity}};
 		previewSource.events.push_back(event);
+		// Save the note being previewed so we can turn it off later (previewNotesOff).
+		previewingNotes.push_back(note);
 	}
-	// Save the notes being previewed so we can turn them off later (previewNotesOff).
-	previewingNotes = notes;
 	// Send the events.
 	void* track = GetSetMediaItemTakeInfo(take, "P_TRACK", NULL);
 	previewReg.preview_track = track;
 	previewReg.curpos = 0.0;
 	PlayTrackPreview(&previewReg);
 	// Calculate the minimum note length.
-	double minLength = min_element(notes.cbegin(), notes.cend(), compareNotesByLength)->getLength();
+	double minLength = min_element(previewingNotes.cbegin(), previewingNotes.cend(), compareNotesByLength)->getLength();
 	// Schedule note off messages.
 	previewDoneTimer = SetTimer(nullptr, 0,
 		(UINT)(minLength ? minLength * 1000 : DEFAULT_PREVIEW_LENGTH), previewDone);
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -637,7 +637,9 @@ MidiNote findNoteInChord(MediaItem_Take* take, int direction) {
 		true,  // end
 		true,  // channel
 		true,  // pitch
-		true  // velocity
+		true,  // velocity
+		true,  // selected
+		true  // muted
 	});
 	if (chord.first == chord.second) {
 		return {-1};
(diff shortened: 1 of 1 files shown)


leonardder:
@ScottChesworth Could you please test drive this? I haven't had the time to test it myself yet. It should report muted when moving across muted notes, as well as the count of muted notes on chords.

MatejGolian:
@leonardder, I'm curious to know how the results of Scott's tests turn out, but my impression so far is that it totally works. I've tried it both on note and CC events and it behaved as expected.
Is my memory playing tricks on me, or was silencing note previews for muted notes part of the original request? If so, over here muted notes still have previews, but that will probably be addressed in a separate PR. Am I correct?

leonardder:
Thanks for pointing this out. Muted notes should no longer play now.

ScottChesworth:
It's working great here, thanks @leonardder! Only one tweak springs to mind, is it easy to get Edit: Mute events (toggle) ID 40055 reporting? Not a biggy if you don't have time, I can look into it at some point, this is already a huge improvement.

MatejGolian:
Don't mention it. :-) At least I believe that that's how the original request went. I apologize for being lazy to double-check it, but regardless, IMO it doesn't make much sense for muted notes to have previews anyway.
I confirm that the previews are gone, so all seems fine.

leonardder:
@ScottChesworth If you'd be able to do the toggle in a follow up, that'd be very helpful!

ScottChesworth:
Will a hearty dose of newb questions also be helpful? I hope so, 'cos here they are :)
- I've pulled the muted branch from your repo and think I know how to push back once I've got something that works.
- Tried doing a simple postToggleMute thing, but that always reports muted no matter what the state is. Seems like it's gonna need something more involved.
- Found this in the API docs which I think is relevant:
bool MIDI_GetNote(MediaItem_Take* take, int noteidx, bool* selectedOut, bool* mutedOut, double* startppqposOut, double* endppqposOut, int* chanOut, int* pitchOut, int* velOut)

So now the newb questions:
1. How do I use that?
2. Er, just that really.

A few hints would probably be enough if you've got time to jot something down. I'm code dopey but keen to learn.

leonardder:
it is probably best to start from a function like `void postMidiChangeVelocity(int command)` or `void postMidiChangePitch(int command)`. Something like:
```
void postMidiToggleMute(int command) {
	if (!shouldReportNotes) {
		return;
	}
	HWND editor = MIDIEditor_GetActive();
	MediaItem_Take* take = MIDIEditor_GetTake(editor);	
	// Get selected notes.
	vector<MidiNote> selectedNotes = getSelectedNotes(take);
	if (selectedNotes.size() == 0) {
		return;
	}
	bool generalize = false;
	if (selectedNotes.size() >= 8) {
		generalize = true;
	} else {
		// Get indexes for the current chord.
		auto chord = findChord(take, 0);
		if (chord.first == chord.second) {
			generalize = true;
		} else {
			generalize = !(all_of(
				selectedNotes.begin(), selectedNotes.end(),
				[chord](MidiNote n) { return chord.first.getIndex() <= n.index && n.index <= chord.second.getIndex(); }
			));
		}
	}
	ostringstream s;
	if (generalize) {
		int count = static_cast<int>(selectedNotes.size());
		// Report something like: "Toggled mute for {} notes"
	} else{ 
		for (auto note = selectedNotes.cbegin(); note != selectedNotes.cend(); ++note) {
			if (note.muted) {
				s << translate("muted");
			} else {  // unmuted
				s << translate("unmuted");
			}
			s << " " << getMidiNoteName(take, note->pitch, note->channel);
			if (note != selectedNotes.cend() - 1) {
				s << ", ";
			}
		}
	}
	outputMessage(s);
}
```

jcsteh:
One question is whether we need to be specific for this command or whether we can just always generalise. That is, do we need to report the mute state for each note (if there are less than 8) or do we just want to always generalise and say "5 notes muted"?

To answer that, we need to know: does alt+m reverse the note state on each selected note such that some are muted and some aren't or does it flip them all in one direction? For example, if you have 4 notes selected where 2 are muted and 2 are not, does alt+m flip each note individually (so you still have 2 muted and 2 unmuted) or does it flip them all in one direction (so you have 4 muted or 4 unmuted)? My guess is the latter, but that needs to be tested. If the forme3r, we need to support non-generalised output. If the latter, we can just always generalise.

ScottChesworth:
It's the former @jcsteh, just verified. Gulp.

jcsteh:
Well, the one positive here is that you can copy the velocity code with
less modification. You don't need that big switch statement, though; you'll
just be saying "mute toggled" or similar when generalising.

PR #738, opened 2022-04-03
Initialize refCount in the UiaProvider constructor and tidy up reference count handling.

glenrgordon:
Since refcount wasn't being explicitly initialized, it's initial value
could have been anything. If the initial value was somewhere near
0xffffffff and it eventually got increased to 1 by multiple AddRef
calls, the transition from 1 to 0 would have caused premature Release
and ultimately a crash. In fact, I observed this multiple times in my
use of Reaper.

### DIFF
--- a/src/uia.cpp
+++ b/src/uia.cpp
@@ -48,7 +48,7 @@ unique_ptr<UiaCore> uiaCore;
 // Provider code based on Microsoft's uiautomationSimpleProvider example.
 class UiaProvider : public IRawElementProviderSimple {
 	public:
-	UiaProvider(_In_ HWND hwnd): controlHWnd(hwnd) {}
+	UiaProvider(_In_ HWND hwnd): refCount(0), controlHWnd(hwnd) {}
 
 	// IUnknown methods
 	ULONG STDMETHODCALLTYPE AddRef() {
@@ -131,7 +131,7 @@ CComPtr<IRawElementProviderSimple> uiaProvider;
 LRESULT CALLBACK uiaWndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam) {
 	switch (msg) {
 		case WM_GETOBJECT:
-			if (static_cast<long>(lParam) == static_cast<long>(UiaRootObjectId)) {
+			if (static_cast<long>(lParam) == static_cast<long>(UiaRootObjectId) && uiaProvider) {
 				return UiaReturnRawElementProvider(hwnd, wParam, lParam, uiaProvider);
 			}
 			return 0;
@@ -184,6 +184,8 @@ bool initializeUia() {
 		return false;
 	}
 	ShowWindow(uiaWnd, SW_SHOWNA);
+	// Constructor  initializes refcount to 0, assignment to a CComPtr
+	// takes it to 1.
 	uiaProvider = new UiaProvider(uiaWnd);
 	return true;
 }


glenrgordon:
Not sure why this failed. Local build works fine with  clang 13.

jcsteh:
Ah, so that's what was causing that crash! Thank you. I've seen that occasionally, but it was rare enough that I never managed to catch it with a debugger.

The build failure is due to a bug in the Visual Studio version that AppVeyor is using. I worked around that by switching to the previous build image, but it looks like that has now rolled over and now we're stuck with a broken build again. Hopefully, that bug has been fixed in the latest version, so we'll have to try that and see. If not, I guess I'll have to figure out some other workaround.

I'm away from a computer at the moment, so I'm not going to be able to look at this until next week.

glenrgordon:
I didn’t see it with a debugger either. What I did was build a local copy of Osara so I had symbols. Then set up Windows to save local copies of crash dumps. Then waited for the crash to occur and looked at the dump.

src/uia.cpp
```diff
@@ -192,8 +194,8 @@ bool terminateUia() {
 	if (uiaProvider) {
 		// Null out uiaProvider so it can't be returned by WM_GETOBJECT during
 		// disconnection.
-		CComPtr<IRawElementProviderSimple> tmpProv = move(uiaProvider);
-		uiaProvider = nullptr;
+		CComPtr<IRawElementProviderSimple> tmpProv = uiaProvider;
+		uiaProvider.Release(); // contained pointer  is now null.
```
leonardder, line 198: I'm slightly worried about the ordering here. Should we first release, than call DisconnectProvider or vise versa?

glenrgordon:
When tmpProv is constructed reference count goes up by one and that reference persists till the end of the block. This means that even after releasing uiaProvider, the reference count is guaranteed to be at least 1 at the time DisconnectProvider is called. If UIAutomation core holds any references to this provider, the reference count will be at least 2. DisconnectProvider will never drop the reference count below 1 because tmpProv holds a reference that UIAutomationCore doesn’t know about.

 

I switched to this approach rather than std::move since the latter is not a common idiom when working with COM.

jcsteh:
> I switched to this approach rather than std::move since the latter is not a common idiom when working with COM.

Is there actually a functional difference? I'd argue using move makes the intent more obvious here, so I'd prefer to leave this bit alone unless it makes an actual difference which I'm missing.

glenrgordon:
I found myself not being able to  reason as clearly about the move semantics when applied to CComPTr because I’ve never seen  that idiom used with it before. Seems like it likely works as expected, so I find it hard to put up a strong argument against it’s use.

jcsteh:
If it makes you feel any better, Microsoft [documents that CComPtr supports move semantics](https://docs.microsoft.com/en-us/cpp/cpp/how-to-create-and-use-ccomptr-and-ccomqiptr-instances?view=msvc-170):

> Both CComPtr and ComQIPtr support move operations that have rvalue references.

glenrgordon:
After looking, I’m very surprised to notice that CComPtr has a move constructor that does the right thing when it comes to std::move, so any reservations I had are now completely gone.

Doesn’t seem like the “uiaProvider = nullptr” statement is actually needed since the CComPtr move constructor actually does a swap.

jcsteh:
I acknowledge that the CComPtr move constructor does a swap, but that's effectively an implementation detail. As far as I recall, std::move makes no guarantees about the state of the source except that it can be safely destroyed or assigned to. So, I think leaving the nullptr assignment is safer.

PR #739, opened 2022-04-04
Use latest Visual studio 2022 image on appveyor

RDMurray:
Let's see if this works.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -61,7 +61,9 @@ build_script:
  # We only build the pot on Windows. The choice of which OS is
  # arbitrary, but there's no point in building it on both.
  - ps: If ($IsWindows) { $targets = "installer", "pot" }
- - ps: '& $scons version=$env:APPVEYOR_BUILD_VERSION publisher="James Teh" $targets ;$Error.clear()'
+ - ps: Set-AppveyorBuildVariable "scons" $scons
+ - ps: Set-AppveyorBuildVariable "targets" $targets
+ - %scons% version=%APPVEYOR_BUILD_VERSION% publisher="James Teh" %targets%
 
 artifacts:
  - path: installer/*.exe
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -62,13 +62,12 @@ build_script:
      # We only build the pot on Windows. The choice of which OS is
      # arbitrary, but there's no point in building it on both.
      If ($IsWindows) { $targets = "installer", "pot" }
-     & {
-       $ErrorActionPreference = "SilentlyContinue"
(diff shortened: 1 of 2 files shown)


leonardder:
Ugh, I guess the crash is because of these warnings. We can probably silence them somehow, though question is whether we really want to do that.

leonardder:
It looks like it is on the radar, see https://github.com/llvm/llvm-project/issues/54487

@RDMurray If you don't mind, marking this as a draft. We'll  probably either need to wait until this has landed in Visual Studio or find a workaround that suppresses warnings or rather, doesn't let the build fail if there's a warning during linking. Please mark this ready for review when it builds correctly.

jcsteh:
It looks like the linker doesn't exit with a failure code, but PowerShell treats this as an exception because there was output to stderr. I vaguely recall there's some way to prevent that, but I don't recall how.

leonardder:
@RDMurray I hope you don't mind, but I just pushed a commit to your branch that will likely fix this. Let's see how the build does.

leonardder:
Never mind, I reverted it without luck. Tried placing 2>&1 after the scons command, but it still fails. May be something Like $ErrorActionPreference = 'SilentlyContinue' might work, but that's also a bit dubious.

leonardder:
note that continue is the default and still raises exceptions. You probably need SilentlyContinue here.

RDMurray:
> note that continue is the default and still raises exceptions. You probably need SilentlyContinue here.

Thanks, I'll try that.

RDMurray:
@ScottChesworth not quite yet :)  This only builds because it clears $Error which is probably not a good idea because it might hide real errors.

RDMurray:
thanks @jcsteh  I found the command from NVDA. it is Set-AppveyorBuildVariable

RDMurray:
That works for Windows, but cmd is no good for mac.  maybe using bash for the build script will work?

leonardder:
With ErrorActionPreference, we need to make absolutely sure that failed builds won't report as succeeded.

RDMurray:
It looks like appveyor sets ErrorActionPreference to Stop for each ps: command.

RDMurray:
Indeed. Interesting that the failed build succeeds on windows, but fails as it should on Mac. Time for me to learn about PowerShell rather than stabbing in the dark. :-)

RDMurray:
I think this should be okay now.

leonardder:
Ugh, such an ugly script this is, credits to PowerShell for the uglyness of course. Nice you've been able to come up with this.

jcsteh:
Thanks for your work on this. This is horrible... but that's a reflection on PowerShell and not your work. :)

PR #750, opened 2022-04-30
Value label in FX parameters dialog

AZSlow3:
The change allows sighted friend follow operations in FX parameters dialog. At the moment there is no visual feedback for switch parameters and hard to interpret numerical value for other parameters.


jcsteh:
Amazing. Thank you.

PR #751, opened 2022-04-30
Avoid unintended FX parameter change

AZSlow3:
In FX parameters dialog, edit control always set parameter of focus kill. Since precision of displayed number is fixed to 4, that can change actual parameter value even in case the text was not modified.

For example in ReqEQ if band frequency was 1000 Hz it is changed to 999.7 Hz.

This PR prevents parameter editing in case the text was not modified.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -390,7 +390,7 @@ class ParamsDialog {
 		char rawText[30];
 		if (GetDlgItemText(dialog, ID_PARAM_VAL_EDIT, rawText, sizeof(rawText)) == 0)
 			return;
-		if (this->param->getValueForEditing() == string(rawText))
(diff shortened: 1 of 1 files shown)


src/paramsUi.cpp
```diff
@@ -390,6 +390,8 @@ class ParamsDialog {
 		char rawText[30];
 		if (GetDlgItemText(dialog, ID_PARAM_VAL_EDIT, rawText, sizeof(rawText)) == 0)
 			return;
+		if (this->param->getValueForEditing() == string(rawText))
```
jcsteh, line 393: Avoid constructing string.
```suggestion
		if (this->param->getValueForEditing().compare(rawText) == 0)
```

jcsteh (approved):
Thank you.

PR #752, opened 2022-05-01
Add a "Report full time for time movement commands" configuration option.

### LINKED ISSUE #613: Ability to toggle complete speaking of position on/of
Thanks Jamie for reminding me to post this as a new issue.  I thought I already had but was apparently wrong!

Currently Osara speaks only the part of the position / time info that is unchanged.  I sometimes find that this incomplete feedback is misleading when moving to a new position and can lead to confusion.  I understand this saves speech feedback for some but would like the ability to have the entire time spoken after using such navigation commands.  Perhaps a toggle for this could be added to the Osara configuration settings.

--Pete

jcsteh:
Fixes #613.

@ptorpey, @Ambro86, could you test a build from here and let me know if this works for you?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -703,6 +703,10 @@ The dialog contains the following options:
 - Report time movement during playback/recording: When disabled, OSARA will not report actions during playback/recording which move the cursor or adjust positions or lengths.
  This includes item navigation.
  Although not strictly time movement, this also includes inserting markers or regions.
+- Report full time for time movement commands: When enabled, time movement commands such as moving by bar/beat will report the full time, even if a portion of that report has not changed since the last movement.
+ For example, if you are at bar 1 beat 1 0% and you move one beat forward, OSARA will report "bar 1 beat 2 0%", even though only the beat changed.
+ When this is disabled, OSARA will only report the part of the time that changed.
+ Using the same example, OSARA would report only "beat 2".
 - Move relative to the play cursor for time movement commands during playback: When enabled, time movement commands such as scrubbing or moving by bar/beat will move from where you are currently playing, rather than relative to the edit cursor.
 - Report markers during playback: When enabled, project markers and regions will be reported during playback as the cursor passes them.
 - Report transport state (play, record, etc.): When enabled, OSARA will report the transport state when you change it; e.g. if you begin playing or recording.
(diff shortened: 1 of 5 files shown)


Ambro86:
Thanks a lot! I've tried it on Windows and all works perfectly!

jcsteh:
This is going to need a complete refactor to handle item/envelope point navigation.

jcsteh:
Assuming this builds successfully, this is ready for another round of testing. This now reports full times for item navigation, envelope point movement, etc. if appropriate. I've also refactored the time formatting code so future code can easily benefit from this setting. There are still cases where we force caching or force no caching; e.g. reporting selection always forces caching, while second and subsequent presses of item edge movement commands always cache (since you're presumably pressing them rapidly at that point and have sufficient context from the previous press).

Ambro86:
Hi, thanks, but this exe file is seen as a virus by windows defender.
Why didn't this happen in previous builds?

2022-08-24 13:51 GMT+02:00, AppVeyorBot ***@***.***>:

jcsteh:
Unfortunately, that happens semi-regularly. OSARA installers aren't
Authenticode signed, so it's hit and miss as to whether Defender likes them
or not. I can tell you that these are built on a service which uses clean
virtual machines for each build, so there's pretty much 0 chance there is a
virus.

ptorpey:
Looks good on Windows.  Thanks for addressing this.  Very nice.

 

--Pete

Ambro86:
I've tried it on Windows and is all ok! Thanks!

2022-08-24 19:34 GMT+02:00, ptorpey ***@***.***>:

TimNoonanVoice:
The new time setting Seems to work correctly on Mac as well.

thanks



On 24 Aug 2022, at 9:49 pm, James Teh ***@***.***> wrote:


Assuming this builds successfully, this is ready for another round of testing. This now reports full times for item navigation, envelope point movement, etc. if appropriate. I've also refactored the time formatting code so future code can easily benefit from this setting. There are still cases where we force caching or force no caching; e.g. reporting selection always forces caching, while second and subsequent presses of item edge movement commands always cache (since you're presumably pressing them rapidly at that point and have sufficient context from the previous press).

ptorpey:
Very nice.  Working very well here in Windows also.  Implementing this in Osara itself is a much cleaner solution than having to import the keymap and LUA script.

 

Thanks.

 

--Pete

Ambro86:
Thanks a lot! I will try this last version. How to activate the full
report of time? Or is actived by default?
Thanks.

2022-05-01 12:39 GMT+02:00, AppVeyorBot ***@***.***>:

PR #761, opened 2022-05-19
When formatting numbers, only remove trailing zeros that are after the decimal point. fixes #760

### LINKED ISSUE #760: Non-correct pitch values zero trailing
OSARA trims zeros in a reported pitch value, and if there will be value like "-20.0", OSARA will report it as "-2 semitones". The same behavior watches an another integer values which ends with zeros (-10.0=-1, -30.0=-3 and etc).

RDMurray:
Fixes #760


jcsteh:
Thanks. This has the added advantage of being easier to read/follow.

PR #762, opened 2022-05-21
Fix reporting of the slider in the Parameters dialog with Windows Narrator.

### LINKED ISSUE #758: Parameters dialog: Narrator always reads value slider as 0%
The title pretty much explains everything. To reproduce, with Narrator running bring up the parameters dialog for a track or effect, then adjust the value using the slider. Narrator will keep reading the value as 0 (tested with osara_2022.1pre-961,04fb6f9c)

The reason I'm bringing this up is that Narrator seems to work much better in UI's built with the new Juce especially really large ones, so I found myself using it much more often when doing audio work on Windows.

jcsteh:
We can't annotate a real slider in the way we need for UIA. Since we don't really use the slider anyway, use a Static text control instead on Windows and annotate its role to fake a slider.
Fixes #758.

I've tested this with NVDA and Narrator, but it will also need testing with JAWS. I know the JAWS scripts have special handling for this slider, so it's possible things will break without tweaks to the scripts.


jcsteh:
Note that I can't use the existing static text control added recently because whenever the value changes, a name change event gets fired by the control and I can't override that, so Narrator was speaking the name of the control every time the value changed.

jcsteh:
CC @pitermach.

pitermach:
Thanks for looking into this. I tried out the new version, and can confirm that Narrator works great now. NVDA, ZDSR and Mac OS are also still working correctly.

You were also right with your hunch re: JAWS, which with the latest version of the scripts now sees the slider just as a text element whose value is always 0 and this is what also gets reported when moving down the parameters list.

Since the number of JAWS users is almost certainly way higher than narrator users I think the best thing to do is let Jim know about this and wait for a script update before merging this

ScottChesworth:
I've emailed the test build to Jim.

jcsteh:
JAWS ignores the accessibility info exposed for this control and fixing the JAWS scripts to deal with this is apparently going to be painful. So this is probably a no-go. This is why we can't have nice things.

pitermach:
THat's unfortunate. The last thing I can think of is contacting Microsoft since the main issue is on their end, I'll link the issue and maybe it'll eventually be adressed

leonardder:
Have you considered changing the slieder to something else, e.g. a spinner?

ScottChesworth:
How hard would it be to get a test build with @LeonarddeR's suggested changes? Narrator in the params dialog still comes up from time to time, I'd like to keep exploring solutions.

jcsteh:
It's not hard to throw a build together, but it will almost certainly break the JAWS scripts, which have very specific code to handle this slider.

ScottChesworth:
> it will almost certainly break the JAWS scripts,
> 
Is that also definitely the case with @LeonarddeR's question a couple of comments above? Reason I'm pushing on this is that there are several significant advantages to using JUCE GUIs with Narrator. Considering the adoption JUCE has in combination with how easy it is to start Narrator, I'd really like to arrive at some sort of middle ground if possible, where Narrator does work and JAWS gets... hmm... encouraged? Persuaded?

ScottChesworth:
Apologies in advance if the above is a daft question. We're well into implementation territory that I don't understand yet.

jcsteh:
> Is that also definitely the case with @LeonarddeR's question a couple of comments above?

Sorry, my answer wasn't clear. Yes, I'm saying I think that would also be the case with @leonardder's suggestion. His suggestion might have more of a chance of working without the scripts, but maybe less so with them.

> Reason I'm pushing on this is that there are several significant advantages to using JUCE GUIs with Narrator.

That's perhaps something we should discuss separately. I know there was a lot of lag with NVDA in the past, but I thought I fixed that a while ago. NVDA does have object navigation, which is pretty similar to the Narrator cursor, even in terms of commands. But anyway, that's out of scope here.

jcsteh:
Abandoning, since this approach won't work.

PR #764, opened 2022-05-30
Announce full Position for Paging

chigkim:
Prototype: I changed announce full position without zeros when pressing page up/down.
Do we need setting for this?

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -363,6 +363,10 @@ string formatCursorPosition(TimeFormat format, bool useCache) {
 	return formatTime(GetCursorPosition(), format, false, useCache);
 }
 
+string formatCursorPosition(TimeFormat format, bool useCache, bool includeZeros) {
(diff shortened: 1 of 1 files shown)


chigkim:
[reaper_osara.dylib.zip](https://github.com/jcsteh/osara/files/8794779/reaper_osara.dylib.zip)

jcsteh:
I think it does require a setting, yes. Most users seem to be happy with only the changed portion of the time being reported, but there are clearly some users (including you) that want the entire time to always be reported. Note that my draft PR #752 adds a setting for this. This needs to affect many more commands than just page up/down, though, so I think we need a solution which doesn't involve modifying every single one of those functions.

jcsteh:
The hope is to implement this all in formatTime, yes. The tricky part is
that there are some callers that always want the cache, some callers that
never want it and some callers that want it depending on the new user
setting. So useCache needs to support three possibilities.

jcsteh:
Closing in favour of #752, which is more broadly applicable and gives us a bit more flexibility. Thanks for your work, though.

PR #770, opened 2022-06-15
Fix the Windows installer to create a keymap backup when a previous backup exists.

RDMurray:


ScottChesworth:
Thanks for such a speedy fix @RDMurray! It's an important one and it worked in testing here, so I'm gonna merge this quicksharp. Hope you don't mind Jamie.

PR #777, opened 2022-08-08
When reporting multiple tracks (e.g. reporting muted tracks or selected tracks), summarise consecutive ranges of tracks (e.g. 1 drums through 5 piano).

### LINKED ISSUE #772: Track selection ranges
Hey guys. So just had a thought regarding the verbosity of track selection. No idea if this will be doable or how much work would be involved but here's my idea.
Currently when you have a bunch of tracks selected osara Will read out The track number and the name if there is one one at a time when you press control shift space. If I have 20 tracks selected it takes awhile for it to read out each track number and name one after another. Usually I just want to know if I've manage to select all of them. Instead could it say something like "tracks 1 through 20 selected"?
 if they're nonconsecutive tracks you could say something like "tracks 1 through five, 10 through 23 and 32 selected"  
If you wanted more detailed information you could still press control shift space twice to bring up the dialogue with a text field.
By  extension I imagine this could probably work with items too but haven't given that one a lot of thought yet.
Curious to hear your thoughts
Cheers 
Matt

jcsteh:
Fixes #772.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -875,3 +875,4 @@ The resulting installer can be found in the installer directory.
 - Leonard de Ruijter
 - Robbie Murray
 - Glen Gordon
+- James Scholes
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3135,31 +3135,80 @@ void cmdIoMaster(Command* command) {
 void cmdReportRippleMode(Command* command) {
 	postCycleRippleMode(command->gaccel.accel.cmd);
 }
+
+string formatTrackRange(int start, const char* startName,
+	int end, const char* endName, const char* separator
+) {
+	ostringstream s;
+	s << start;
+	if (startName && startName[0]) {
+		s << " " << startName;
+	}
+	const string startText = s.str();
+	const int diff = end - start;
+	if (diff == 0) {
+		// Single track. Just report that track.
+		return startText;
+	}
+	s.str("");
+	s << end;
+	if (endName && endName[0]) {
+		s << " " << endName;
+	}
+	const string endText = s.str();
+	if (diff == 1) {
+		// Two consecutive tracks, so report them separately.
+		s.str("");
+		s << startText << separator << endText;
+		return s.str();
+	}
+	// Translators: Used when reporting a range of tracks. {start} will be
+	// replaced with the first track. {end} will be replaced with the last track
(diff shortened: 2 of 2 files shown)


jcsteh:
Thanks to @jscholes for the original code. I've made a few tweaks, including:

1. Fixed some extraneous comma separators and other little nits, plus some code refactor.
2. Made the "x through y" message properly translatable.
3. Removed the "x and y" message, as we can just report these as two separate tracks separated by a comma (x, y), so I didn't feel it was useful to differentiate this case. Let me know if there's a reason the "x and y" message is preferred.

UX questions:

1. Like the original code, this also impacts the reviewable message which appears in a dialog when you press the command twice. Should it instead only affect the non-dialog version, leaving the dialog version to display the full list in case the user wants to review all of it?
2. For ranges to make sense, the track numbers need to be reported even if "Report track numbers" is disabled. For individual tracks (not in a consecutive range), should the track number be reported in this case? For now, it is reported because doing otherwise seemed inconsistent with ranges, but I can change that if anyone has a strong argument to the contrary. This does mean that this command ignores the "Report track numbers" setting, though.

jcsteh:
CC @mattymc2 in case he has thoughts, ideally useful ones.

mattymc2:
I always have thoughts. No promises how useful they are though. I will check this out this evening. Legendary that you could get this implemented though.

jscholes:
@jcsteh 

> this also impacts the reviewable message which appears in a dialog when you press the command twice. Should it instead only affect the non-dialog version, leaving the dialog version to display the full list in case the user wants to review all of it?

Now you mention it, this was one of the outstanding parts that stopped me from submitting a PR at the time.  I was of the opinion, and still am, that allowing the dialog to show the expanded list was a convenient way to give users the best of both worlds, but I never got around to it.

I also considered adding an Osara setting to enable/disable ranges, but that seemed/seems like overkill.

jscholes:
Thank you for cleaning up the code to get it in the hands of users!

jcsteh:
@mattymc2's original request suggested that the dialog should show the verbose form too, so that's three of us leaning in that direction now.

jcsteh:
I've implemented this now; the dialog shows individual tracks, not summarised ranges.

ScottChesworth:
I hereby proclaim that this is flippin' marvelous. Thanks @jcsteh!

PR #782, opened 2022-08-20
Fix media explorer volume feedback in Reaper 6.65 and above, fixes #781

### LINKED ISSUE #781: Media Explorer preview volume announcements are broken
First reported on the RWP list, with the latest version of OSARA and Reaper if you use the Alt+Shift+Up/Down hotkeys to adjust the media explorer preview volume, the value is no longer spoken and OSARA just says "DB". The report was on Windows but I can also reproduce this on Mac.

Reaper 6.65 changed the media explorer window by adding an edit box to type in a preview volume in addition to the slider, perhaps these changes are what caused this to break.

RDMurray:
fix #781  by checking for the control id of the new edit control for media explorer volume in Reaper 6.65.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1840,7 +1840,10 @@ void postToggleTakePreservePitch(int command) {
 }
 
 void postMExplorerChangeVolume(int cmd, HWND hwnd) {
-	HWND w = GetDlgItem(hwnd, 1047);
+	HWND w = GetDlgItem(hwnd, 997);
+	if(!w) {// support Reaper versions before 6.65
+		w = GetDlgItem(hwnd, 1047);
+	}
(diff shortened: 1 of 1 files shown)


RDMurray:
Tested in Reaper 6.66 and 6.64. The only difference in feedback is that just the number is reported in 6.65 and above, whereas previous versions also reported dB. I think this is okay as it is consistent with the reporting of volume for tracks.

RDMurray:
I'd appreciate it if someone could test this on Mac with Reaper 6.65 or above.

pitermach:
Just tested under Mac OS Ventura Beta and Reaper 6.65. It's working great now, thanks!

ScottChesworth:
Thanks for the quick fix Robbie. Merged it.

PR #783, opened 2022-08-21
Create a toggle action for each OSARA setting.

### LINKED ISSUE #766: Actions to toggle individual prefs-settings via shortcuts
I would like to ask for some toggle-actions, that allow turning on/off the various options you can currently only set in the Osara Preferences.

Some of these are helpful but I can imagine, that they might not be needed all the time, so an easy way to turn them off/on via shortcut would be very helpful.
It is possible to set it in the preferences, but to me, it feels a little tedious to have to always go into the preferences to toggle them. 
For instance "Report markers during playback" is something, you might not always be interested in, so if I could toggle this easily via shortcut, this would be awesome.

I have tried to turn them on/off in Lua by setting the extstates Osara uses, but this didn't work.

Having the settings toggleable via action would be awesome, especially if you could see the toggle-state in the actionlist as well.

jcsteh:
This reports the changed state when run. It also reports the toggle state to REAPER so that it can show up in the Actions dialog, be queried from scripts, etc.

This required a significant refactor of the configuration code which makes it easier to maintain and extend.

Fixes #766.


jcsteh:
This attempt is unfixably broken on Mac. This is why we can't have nice things.

jcsteh:
Okay. We now have a completely refactored configuration framework which should make adding new check boxes a lot easier (three lines of code). If I've done this properly, everything should look the same. This particularly needs testing on Mac, since I'm only testing on Windows here and this is a massive rewrite.

In addition, every setting gets a toggle action which reports when you change it and also reports the toggle state to REAPER so it shows on/off in the Actions dialog.

These commits need cleaning up, but I'd rather do that once we're sure on the direction. I think this is ready for some testing.

jcsteh:
In case anyone's curious, here's what it looks like to define a setting:

```
BoolSetting(reportScrub,
	"Report position when &scrubbing",
	true)
```

where true is the default. This defines a variable for it (settings::reportScrub), makes it get loaded/saved, adds it to the GUI (automatically figuring out coordinates) and creates a toggle action for it.

ScottChesworth:
I believe modern folk would describe this as God tier. :D
So just to be clear, is it the OSARA Config GUI and how new additions to that show up on Mac that needs testing?

ptorpey:
Just FYI, I successfully downloaded the earlier build today, but Windows Defender blocked the download of this version.  I haven’t had that happen in a long time with these builds.  Don’t know why this might be different.

 

--Pete

jcsteh:
You don't need to test additions to the dialog on Mac, just test that the dialog works as it always did.

ScottChesworth:
Paging @jennykbrennan and @pitermach, if either or preferably both of you have time, we need a few seasoned Mac bashers to install the latest AppVeyor build in this thread, look over the OSARA Config dialog with a microscope and report any oddeties.

soundwarrior20:
Hi do you still need help with this?  how do I install the bill please?

On 22 Aug 2022, at 00:34, ScottChesworth ***@***.***> wrote:

﻿
Paging @jennykbrennan and @pitermach, if either or preferably both of you have time, we need a few seasoned Mac bashers to install the latest AppVeyor build in this thread, look over the OSARA Config dialog with a microscope and report any oddeties.

GarthHum:
Well, I’ve had a brief play and all looks well, certainly no microscopes used though. Should I have seen toggles in the action list for the existing configuration check boxes? If so not found.

ScottChesworth:
> Should I have seen toggles in the action list for the existing configuration check boxes? If so not found.
> 
Uh oh, those should be visible yeah. Just to double check everyone's on the same page, typing "OSARA: toggle" into the filter with section set to Main should find a bunch of 'em.

jcsteh:
Yeah, the toggle actions should work... and are totally not the thing I expected to fail on Mac.

GarthHum:
Correct.
There are 62 actions in the list when I filter to OSARA.

TimNoonanVoice:
The Check boxes seem fine in the dialog, and in my Mac set-up here, I find the OSARA toggle options in the actions list, and they appear to change their state fine.

Tim


On 22 Aug 2022, at 11:51 am, James Teh ***@***.***> wrote:


Yeah, the toggle actions should work... and are totally not the thing I expected to fail on Mac.

jcsteh:
62 sounds about right. As an example, you should now have "OSARA: Toggle Report changes made via control surfaces", which didn't exist before this pull request.

mespotine:
Awesome and thank you for the features.

TimNoonanVoice:
All seems good here too on Mac.

Tim


On 22 Aug 2022, at 10:46 am, Garth Humphreys ***@***.***> wrote:


Well, I’ve had a brief play and all looks well, certainly no microscopes used though. Should I have seen toggles in the action list for the existing configuration check boxes? If so not found.

jcsteh:
<sigh> This attempt is unfixably broken on Mac. This is why we can't have
nice things.

ptorpey:
Amazingly, this build downloaded successfully without anything blocking it.  Go figure!

 

--Pete

PR #785, opened 2022-08-26
When OSARA's Report time selection command is pressed twice, include all selections, not just the focus.

### LINKED ISSUE #456: Use control+shift+spacebar twice quickly to provide additional feedback
Scott recommended that I float this idea on GitHub to see what people think and assess if the effort to implement the concept is worth the potential benefits.

The control+Shift+Spacebar hotkey provides speech feedback on what is selected only for the element that has focus.  This gives accurate but limited feedback to those using a screen reader compared to what is available to a sighted user.  This feedback is succint and directly related to what the user's current task and its function should remain.

However there is an opportunity to provide more detailed and extensive feedback to the user about the state of the project and what is selected across the project when the user hits the control+shift+spacebar hotkey twice quickly.  This could include telling the user what items may be selected on tracks other than the track that has focus, other information of interest that is readily available to a sighted user, etc.

The idea is to retain the functionality of the control+shift+spacebar hotkey while expanding on the function of hitting this hotkey twice to provide extended information to the user.

The expectation is that the the additional information can avoid confusion for new users on why they might be getting unexpected results from their actions and also reduce potential errors on the part of more experienced users.

--Pete

jcsteh:
Fixes #456.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -744,7 +744,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Report tracks with phase inverted: Alt+Shift+F9
  - Pressing this twice will display the information in a dialog with a text box for easy review.
 - OSARA: Report track/item/time selection (depending on focus): Control+Shift+Space
- - Pressing this twice will display the information in a dialog with a text box for easy review.
+ - Pressing this twice will display the information for all selections (not just the focus) in a dialog with a text box for easy review.
 - OSARA: Remove items/tracks/contents of time selection/markers/envelope points (depending on focus): Delete
 - OSARA: Report edit/play cursor position and transport state: Control+Shift+J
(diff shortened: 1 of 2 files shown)


jcsteh:
@ptorpey, @tbdalgaard, @RichCaloggero, feedback appreciated.

Lo-lo78:
Nice idea!

👍

Il 26/08/2022 12:50, James Teh ha scritto:

ptorpey:
Looks very nice.  Great that one can arrow down through the list of selected items, tracks, and time.  Should give people a good idea of everything that is selected.

 

I need to find a refresher on how to do non-contiguous selection of items since I couldn’t get that to work.  But time selections and non-contiguous selection of tracks worked great.

 

Thanks for this.

 

--Pete

PR #789, opened 2022-09-08
Fix "OSARA: Focus event nearest edit cursor"

RDMurray:
Reported on RWP, Control+f focused the wrong event in the list because the event positions in the event list are based on the start of the midi source, and Osara currently always bases positions on project time.

This assumes that PPQ is set to the default of 960. is it common for people to change it? I suspect not.

On a related subject, Reaper now seems to cause the edit cursor to follow focus in the event list by default. I think this makes the option in Osara preferences redundant. The only thing it does now is to automatically focus the nearest event when switching to the event list. Should we remove it?

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -203,15 +203,17 @@ struct MidiEventListData {
 		auto setting = format("list_{}", index);
 		char eventData[255] = "\0";
 		if (MIDIEditor_GetSetting_str(editor, setting.c_str(), eventData, sizeof(eventData))) {
+			MediaItem_Take* take = MIDIEditor_GetTake (editor);
 			string key, val;
 			istringstream s(eventData);
+			double eventPosPpq = -1.0;
+			double lengthPpq = -1.0;
 			while(getline(getline(s, key, '='), val, ' ')) {
 				if (key == "pos") {
-					auto eventPosQn = stof(val);
-					data.position = TimeMap2_QNToTime(nullptr, eventPosQn);
+					eventPosPpq = stof(val) * 960.0; // fixme: only works with the default ppq of 960.
+					data.position = MIDI_GetProjTimeFromPPQPos(take, eventPosPpq);
 				} else if (key == "len") {
-					auto lengthQn = stof(val);
-					data.length = TimeMap2_QNToTime(nullptr, lengthQn);
+					lengthPpq = stof(val) * 960.0;
 				} else if (key == "msg") {
 					data.message = val;
 				} else if (key == "offvel") {
@@ -219,6 +221,10 @@ struct MidiEventListData {
 				} else if (key == "sel") {
 					data.selected = val == "1" ? true: false;
 				}
+			}
+			if (lengthPpq>=0.0) {
+				double endPos = MIDI_GetProjTimeFromPPQPos (take, lengthPpq + eventPosPpq);
+				data.length = endPos - data.position;
 			}
 		} else {
 			data.position = DBL_MAX;
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -15,6 +15,7 @@
 #include <functional>
 #include <float.h>
 #include <compare>
+#include<regex>
 #include "midiEditorCommands.h"
 #include "osara.h"
 #include "config.h"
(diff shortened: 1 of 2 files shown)


jcsteh:
I'm confused by this. So REAPER gives us event times in QN (which is a project time format), but they're relative to the MIDI take, not the project? That seems odd to me, especially since you can't get QN times from a MIDI take.

The REAPER docs say pos_pq/len_pq may also be available in the event data. Should we be using those instead?

I'm pretty sure I've seen at least one mention on RWP of someone changing the PPQ to something other than the default of 960. @ScottChesworth , do you recall anything like that?

@leonardder, what do you think?

RDMurray:
I started by assuming that pos_pq and len_pq would work, but they didn't.  I logged the string returned by MIDIEditor_GetSetting_str to the console to see what was going on.  pos_pq only appeared for the first event in the list and len_pq wasn't there at all.   pos and len match the event times in the list which are midi source based.

I don't know what's going on with pos_pq. @justinfrankel could you take a look at this please? There doesn't seem to be a way to convert the positions from MIDIEditor_GetSetting_str to project time without knowing the ppq setting of the midi take. Would it be possible to add the ability to retrieve the PPQ setting to GetMediaItemTakeInfo_Value     ?

RDMurray:
> pos and len match the event times in the list which are midi source based.

Sorry I just looked at it again and the event list does show project time unless view/Piano roll time base is set to source beats.

Lo-lo78:
@RDMurray Works well.
The only problem is that if there are no midi notes in the event list in the 1 beat 1 bar but there are CC midi events they are not selected well. It always takes the first event.
After the first midi note yes. The midi event is perfectly selected.

RDMurray:
SWS gets ticks per quarter note by parsing the item state chunk.  I guess we could do the same, but would probably need to cache it for performance.

ScottChesworth:
> I'm pretty sure I've seen at least one mention on RWP of someone changing the PPQ to something other than the default of 960. @ScottChesworth , do you recall anything like that?
> 
Got a vague memory of someone mentioning reducing PPQ to make quantize behave more how they thought it should yeah, but we're talking a very vague memory and I wasn't able to find it in the archive, so don't hang your hat on that. FWIW, 960 has been the standard setting in every DAW I've used here over the years, so if I were in your shoes, I wouldn't break my back to support anything else. The vast majority of OSARA users that I'm aware of tend to think in percentages and I'm pretty sure I've never been asked how to modify PPQ here.

src/midiEditorCommands.cpp
```diff
@@ -203,15 +203,17 @@ struct MidiEventListData {
...
+			MediaItem_Take* take = MIDIEditor_GetTake (editor);
 			string key, val;
 			istringstream s(eventData);
+			double eventPosPpq = -1.0;
+			double lengthPpq = -1.0;
 			while(getline(getline(s, key, '='), val, ' ')) {
 				if (key == "pos") {
-					auto eventPosQn = stof(val);
-					data.position = TimeMap2_QNToTime(nullptr, eventPosQn);
+					eventPosPpq = stof(val) * 960.0; // fixme: only works with the default ppq of 960.
```
jcsteh, line 213: We use 960.0 twice, so please make it a const. It doesn't have to be at module level, just in a common scope for both uses is fine.

jcsteh (approved):
I really don't love this, but if we don't have another choice, such is life. I wish the event data included pq.

Let's just leave 960 for now.

Approving with my comment below addressed. I'd also really like an opinion from @leonardder , but if he can't get to this in the next few days, feel free to merge.

RDMurray:
@leonardder  It's very inconsistent.  I saw pos_pq a couple of times, but I just created a new midi item, inserted some notes and pos_pq wasn't reported for any of them. I also didn't know what the number was; It certainly wasn't the ppq position, it was something like 1.0xxxxx. 

Getting the ticks per quarternote setting from the item in the same way as SWS shouldn't be a problem, as long as we cache it. That way parsing the item state would only be necessary once each time the item changed.

leonardder:
> Getting the ticks per quarternote setting from the item in the same way as SWS shouldn't be a problem, as long as we cache it. That way parsing the item state would only be necessary once each time the item changed.

If we can do that, I think I'd prefer that.

RDMurray:
I have implemented the above. It didn't take much effort to find a MIDI file that uses a different PPQ, so this is definitely necessary. for testing a file included with windows "C:\Windows\Media\flourish.mid" has a PPQ of 384. You can check in Reaper with control+f2.

src/midiEditorCommands.cpp
```diff
@@ -26,6 +27,29 @@
 using namespace std;
 using namespace fmt::literals;
 
+// returns the pulses per quarter note of the first midi source in the item.
+int getItemPPQ(MediaItem* item) {
+	const int defaultPPQ = 960;
+	static MediaItem* cachedItem;
```
leonardder, line 33: This caches the media item for as long as REAPER is running. Would it be possible to refresh the cached item for as long as the event list is open?
Not sure what `kbd_OnMidiList` does exactly, but may be that can be used?

RDMurray: It caches PPQ for the item until it is called with a different item. Do you know if it is possible to change PPQ after an item is created? It isn't possible to change it  in the item source properties.

leonardder: I think you're right that the ppq will stay always the same then. The option that allows you to change ppq says new items explicitly.

leonardder (approved):

leonardder:
@RDMurray I believe you can safely merge this, or do you want me to do it?

PR #790, opened 2022-09-10
Revert let edit cursor follow focus in MIDI event list

leonardder:
Not sure when, but this became obsolete since REAPER does this natively. See https://github.com/jcsteh/osara/pull/599#issuecomment-1241954806

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -714,8 +714,6 @@ The dialog contains the following options:
  When disabled, track numbers will not be reported except for unnamed tracks.
 - Report FX when moving to tracks/takes: When enabled, OSARA will report the names of any effects on a track or take when you move to it.
 - Report MIDI notes in MIDI editor: When enabled, OSARA will report the names of individual MIDI notes and the number of notes in a chord.
-- Edit cursor follows focus in MIDI event list: When enabled, the edit cursor in the MIDI  editor event list wil follow along with the focused event.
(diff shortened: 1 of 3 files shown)


jcsteh:
I vaguely recall someone saying this worked fine on Mac even though we couldn't support it there, suggesting that REAPER must have supported it there. I wonder if there's some setting or other condition that affects whether REAPER does this?

Anyway, since multiple people have verified that this isn't necessary, I'll merge this. Note that i haven't tested it myself, though.

Thanks.

jcsteh:
Blah. I messed up the PR number in the merge commit, so I forced push d66dee0a3c65c4421fba1499b3cc7f06f18dd9bb with a corrected commit message.

jcsteh:
Meanwhile, now that I've merged this, I spotted this in @RDMurray's https://github.com/jcsteh/osara/pull/789#issue-1367030169:

> The only thing it does now is to automatically focus the nearest event when switching to the event list.

That does still seem like useful functionality, since otherwise, pressing down arrow will throw you right back to the top of your item. Hmm.

leonardder:
Oops, oh well. I think instead of a revert of a revert, I should rewrite it in such a way that it focuses the nearest event again, but I guess it is ok to make that default behaviour without an explicit toggle?

PR #791, opened 2022-09-12
Re-introduce focusing the nearest event when entering event list from outside

leonardder:


RDMurray:
Looks good, I agree an option isn't necessary for this. Merging, but we still have the problem I am attempting to fix in #789

PR #798, opened 2022-10-04
Fix F6 in the FX window to work with non-english translations

### LINKED ISSUE #792: I can't get to vst ui by pressing f6 in fx window
> 

Hello.
I recently found a problem when I was using an old version of osara, about before the pre990 version.
When I loaded vst on the track, I could open the fx window and press f6 to enter the vst plugin interface.
For example, I loaded Ozone 9, after I pressed f6, then pressed tab, I could see a button like this →
Preset Readout Frame button Enter
Options button Enter
Left Arrow button Enter
Right Arrow button Enter
History button Enter
Help button Enter
Then I can press enter or the arrow keys to control these functions.
But recently I updated osara, using the pre996 version, also have tried to download the latest version of osara, press f6 are not working, will only switch the solo state of the track.
I used the installed version of reaper and the portable version of reaper, both versions are 6.67, only the difference of osara version, and found that the old version of osara is normal, but the new version of osara can not enter the plug-in ui by pressing f6.
Is there a new way to do it? Or is this a problem with osara?
Thanks!

Translated with www.DeepL.com/Translator (free version)

RDMurray:
Using F6 to focus the plugin UI didn't work with some Reaper langpacks due to relying on the window title. Fixes #792

### DIFF
--- a/src/fxChain.cpp
+++ b/src/fxChain.cpp
@@ -53,17 +53,9 @@ void shortenFxName(const char* name, ostringstream& s) {
 
 bool maybeSwitchToFxPluginWindow() {
 	HWND window = GetForegroundWindow();
-	char name[8];
-	if (GetWindowText(window, name, sizeof(name)) == 0)
-		return false;
-	if (strncmp(name, "FX: ", 4) != 0 && // FX chain window
-		// floating FX window, for different plug-in types
-		strncmp(name, "DX: ", 4) != 0 &&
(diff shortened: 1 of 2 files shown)


leonardder (approved):
Does this also cover the bypass/active issue you mentioned in #792
If yes, as this has been tested, feel free to give it a spin and merge it.

RDMurray:
I was wrong about bypass/active being affected. It doesn't rely on the window title.

PR #804, opened 2022-10-27
When Report changes made via control surfaces is enabled, don't report automation that is being played back, nor parameter changes caused by an automation mode change.

### LINKED ISSUE #802: The problem that osara will read aloud 0.0 when switching between Automation mode again
Hi friends.
When I press
ctrl+f12 and check this option
Report changes made via control surfaces
Once I switch to Automation mode, it keeps reading 0.0 aloud.
Whether I press
shift+l to switch modes in the options, or press
ctrl+shift+\
The same thing happens when I switch to Automation mode.
The content that is supposed to be read aloud will be interrupted and become read aloud 0.0.
Can this problem be fixed?
Thank you!

jcsteh:
Fixes #802.


jcsteh:
Testing this requires the latest dev build of REAPER. @ScottChesworth, could you take this for a spin and see if it does what you want? I don't have a surface here, so I can't really test this properly. You should text with volume, pan and FX parameters with no automation, automation playing back, writing automation, touch/latch, etc.

ScottChesworth:
I'm slammed here at the moment, not likely to be able to spend a ton of time with it until early next week, but I can circulate the build and instructions to a handful of heavy surface users. Excited to see where this goes!

leejul:
Test for OSARA and CSI performed on:
OS - Windows 10 pro 64 bit.
Reaper version - 64-bit Reaper installed on a desktop
Reaper version : reaper669+dev1026_x64-install.exe
OSARA version - osara_pr804-1023,7332f373.exe

1. Pans state OSC values. It's always been like this, is it possible to have the right pan values ​​as a track with surface.

2. I select write mode and NVDA no longer tells me the fader volume value of the selected track only, this is great. This allows us to do an automation without having to disable the action of OSARA.
If not wanted, it would be very cool to add that.
However, here is the problem:
If I activate the write mode, I do a volume automation and I let the playback of the project advance, the track names are spoken correctly when I touch a fader except the selected one, until then, everything is fine.
But if I stop playback and go back to playback mode, no more faders speak when I touch them.
Note that I have enabled return to read mode automatically after writing an automation in Reaper's preferences.
In addition, OSARA loses track order. I press up arrow or down arrow and NVDA tells me nonsense.
It tells me either any track number or a volume value.
But, the track selection runs fine, I asked a seer.
In addition, the panoramic works but he no longer speaks.
The Select, Mute, Solo and Arm buttons work perfectly on all tracks.
The Jog Wheel tells me a volume value when I turn the jog wheel.
The transport buttons also work perfectly.
And if I load an FX on the track, the faders are placed on loading but nothing is announced.
Solution, I have to delete all tracks and start over.
That's it, I can't go further in the tests for the moment.
Thank you Jammie for your excellent work and for giving time to the CSI project.

jcsteh:
>     1. Pans state OSC values. It's always been like this, is it possible to have the right pan values ​​as a track with surface.

This sounds like a separate issue. Could you please file this separately, providing an example of what is being reported and what you expect instead? Thanks.

>        But if I stop playback and go back to playback mode, no more faders speak when I touch them.
>        Note that I have enabled return to read mode automatically after writing an automation in Reaper's preferences.

When in read mode, your faders should have no effect. Read mode means that you're only playing back automation; you can't make any changes to the parameters. Can you explain why you expect OSARA to report the faders in this case?

leejul:
ctn: When in read mode, your faders should have no effect. Read mode 
means that you're only playing back automation; you can't make any 
changes to the parameters. Can you explain why you expect OSARA to 
report the faders in this case?


Yes I know but when I touch the faders, NVDA is supposed to say the 
track number or name because my faders are touch sensitive.
While there, NVDA says nothing more.
Same problem with the rotary knob for the pan.

jcsteh:
OSARA reports the track number/name when a parameter changes. If we're ignoring the parameter change, there's no reason to report the track number/name. I could have it report the track number/name for all surface notifications, but then we're back where we were at the start: playing back automation will keep reporting track numbers/names.

davidkreynolds:
Hi,

 

I can’t get any of this to work with Jaws or NVDA.

Windows 11 X64, Reaper 6.69.

1.Arrows no longer speak selected track.

2.Faders no longer speak.

I’m also copying an email I received from someone who doesn’t have a github account.

<quoted text begins>

Portable Reaper installed and the new Osara into that. Existing projects playing fine.

 

Although hadn't installed CSI as yet, still getting the original problem after turning on speak control surface changes in Osara preferences. As before on my installed version, I am hearing 'Atmosphere volume -13dB' as I arrow from track to track. When I press F4 for the Actions list, I hear '-13dB'. There's no point conducting any further tests with CSI. This just shouldn't be happening.

 

For clarification, what keeps getting spoken (Atmosphere volume -13dB) is the name of a track (Atmosphere) to which there was continuous volume automation recorded from my X-Touch. If I bypass the volume envelope on the Atmosphere track, then the same problematic info is announced constantly for the only other track to which volume automation was recorded in the same way. Bypassing that track's volume envelope stops the problem entirely.

 

Note that there is no problem with Osara speaking this kind of automation data where envelope points are dropped in using Reaper actions as opposed to fader movements from the surface.

<Quoted text ends>

David.

jcsteh:
The required REAPER change isn't in 6.69. You will need the latest REAPER 6.69+dev pre-release build. Per the pre-release instructions, I'm not supposed to link to that here, but if you Google reaper pre-release, you will find it pretty quickly. :)

ScottChesworth:
I've already linked a bunch of testers to it and followed up with advice to manually refresh the builds page to make sure the latest uploads are displayed.

leejul:
Hi Jamie,
Here is an audio demonstration of the behavior I have here.
https://1drv.ms/u/s!AldObfqUjgIthqgaIh-7-0GaRyVp3g?e=i3rj89

MatejGolian:
I don't have any control surface, but for me Osara 2022.1pre-1024 seems to have broken something, so my guess is that it could be attributed to changes related to this control surface stuff.
After launching Reaper for the first time with Osara 2022.1pre-1024 all seems fine, but when I close Reaper and reopen it again and open any kind of Reaper menu (be it a general menu or context menu), NVDA goes crazy and starts reading all menu items on each key press. When I go back to 2022.1pre-1021, all is back to normal.
I don't understand why this problem only shows up from launching Reaper with 2022.1pre-1024 for the second time and onward, but that's how it is.
It would be nice if someone could confirm this.

MatejGolian:
Update: Now it seems fine after all, but I'll investigate further.

jcsteh:
@MatejGolian, if there is indeed a problem with pre1024, it's not related to this surface stuff. This surface stuff hasn't been merged into main snapshots yet; you can only get builds from this pull request. Thus, if the problem reappears, please file a separate issue. Thanks.

MatejGolian:
OK, thank you.

leejul:
Hi Jamie,
I downloaded the latest version of OSARA and the problem is not fixed.
Are there any developments for this issue?
Don't hesitate if I can help.

Thanks.

jcsteh:
I don't have any updates. I really don't understand enough about what's going on nor what is expected to do anything further here. As an example, see https://github.com/jcsteh/osara/pull/804#issuecomment-1294382628. It seems there's an expectation that we report track numbers/names for every touch, even when playing back automation, but without reporting the automation changes. As I understand it, those goals are just incompatible.

leejul:
Hi Jamie,
I don't understand what he is asking either. They can easily disable 
control surface reporting when writing an automation and re-enable 
control surface reporting when in playback mode.
Even better, they can even do a custom action that would be triggered on 
the control surface.
You just need to add the following 2 actions:
OSARA: Enable Report changes made via control surfaces
OSARA: Desable Report changes made via control surfaces
In this case, he could do the following 2 custom actions:

Custom action 1:
Automation: Set track automation mode to write
OSARA: Desable Report changes made via control surfaces

Custom action 2:
Automation: Set track automation mode to read
OSARA: Enable Report changes made via control surfaces
We could switch between the 2 modes with 2 dedicated buttons on the 
surface so as not to hear NVDA while writing automation.
But I would like to find the behavior of before with OSARA and my 
control surface.
If you listened to the audio demo that I put in my last message, I did 
not have this behavior with the old version of OSARA.
Now, after writing an automation, OSARA does not report anything on the 
X-Touch. I have to save the project and restart Reaper to find the 
control surface report.
Do you think you can fix this behavior or do I have to find another 
solution?
Thank you bro.

jcsteh:
@leejul, you wrote:

> Yes I know but when I touch the faders, NVDA is supposed to say the
track number or name because my faders are touch sensitive.

But the whole point of this request is to stop reporting parameter changes caused by automation changes or changes caused other than directly by the user. The only reason your faders report anything when you touch them is that this causes a parameter change in REAPER. It's impossible to separate this from other parameter changes when automation is being played back. The goals you want for this change are incompatible.

leejul:
Ok, I'll try to be as clear as possible.
Personally, I'm not asking for anything special. I would simply like to 
find the initial behavior of OSARA. As we had in the old version.
Currently, if I'm doing an automation, as soon as I change the 
automation mode to playback mode, I no longer get NVDA feedback when I 
turn pan knobs or faders. I absolutely have to restart Reaper to find 
the report via the control surface. This is not normal. This is a bug 
that was not there in the old version of OSARA.
If users want to disable NVDA's feedback when writing automations, it's 
very simple. With CSI and the X-Touch, they simply have to press a 
button to deactivate the report via the control surfaces and NVDA no 
longer speaks while writing. The procedure is very simple. For the 
moment, I can no longer use the X-Touch to make automations because as 
soon as I make one, I have no feedback as soon as I switch to reading 
mode. So I don't know which track I'm on anymore.
It must be understood that the initial reason for using a control 
surface is to make automations. Now I can't use it anymore.
I make a demonstration in the mp3 that I put in my posts.
So, my question is the following, is it possible to find the same 
behavior for the control surfaces that we had with the old version of OSARA?
Thanks again for your time Jammie.

jcsteh:
This is a pull request build. It hasn't been merged. The old behaviour of OSARA is still in the official snapshots.

I created this build because there were complaints about OSARA reporting parameter changes during automation playback, automation mode changes, etc. See #802. When automation is playing back, you can't actually change parameters anyway; you only get the automated parameter values. I don't understand why you want reporting in this case, since you're not able to actually change the parameters anyway.

Regardless, it seems there are conflicting requirements here and I can't determine what people actually want. So I think I'm giving up on this one.

leejul:
Ok Jamie, I found the problem.
In Read and Touch automation mode, we no longer have feedback from faders and pans.
On the other hand, everything works perfectly in Latch, Write, Group and Trim/Read mode.
We have this problem only in Read and Touch mode.

jcsteh:
As I keep saying, when you're in read mode, you're playing back automation. You can't make parameter changes in read mode; they should do nothing. Any parameter changes that occur in read mode are only automation. Given that, why do you expect parameter feedback in read mode?

leejul:
This bug only occurs in Read automation mode and Touch mode.
The active automation mode is not valid only during playback. When the 
project is stopped, the automation mode does not change. Unless you tell 
Reaper to switch automation mode after writing an automation, when the 
project is stopped.
So I explain to you.
To do an automation, I change the mode to Write.
I launch the playback of the project and I move the fader or the pan.
I press stop.
Now, if I want to listen to the result, I absolutely have to change the 
mode otherwise I will overwrite my automation as soon as I start playback.
So I change the automation mode to Read or Trim/Read. Or if I want to 
apply a correction, in Touch or Latch mode.
If I choose Read or Touch mode, the X-Touch no longer speaks, even if 
the project is stopped while all the other automation modes work 
perfectly, NVDA correctly reports the values.
This bug comes from OSARA with the control surface report, I think.
I tested on a machine without CSI. I used the Mackie Control Universal 
protocol integrated in Reaper with my X-Touch and it's exactly the same bug.
The X-Touch should speak the track number/name and volume/pan value at 
all times when the project is stopped. Otherwise, we cannot know which 
track we are selected. It should not be forgotten that the faders are 
sensitive to the touch and that as soon as you touch a fader, you select 
the track, during the reading of a project or when stopped.
So if we don't get feedback when we touch a fader, we can't be sure 
we've selected the right track.
For users who don't want NVDA feedback regardless of automation mode, 
they already have a dedicated button on the X-Touch in CSI that toggles 
the Report changes made via control surfaces feature.
It doesn't matter the type of control surface, it's very easy to add in CSI.
So they have the power to disable NVDA feedback under their fingers, 
whenever they want.
All that needs to be done is to have the X-Touch vocalized at all times, 
regardless of automation mode.
That's why I added the Report changes made via control surfaces function 
easily accessible on a dedicated button on the X-Touch.
Before doing an automation, in Write mode, you press the button to 
deactivate feedback.
Same thing in Touch or Latch mode.
As I told you in another post, to have full control over NVDA feedback 
with control surfaces, we would need the following 2 actions:
1. Enable Report changes made via control surfaces
2. Disable Report changes made via control surfaces
In this case, we could make custom actions or scripts that would be 
triggered directly on the control surface.
So the user could press the Write button, and a script would be assigned 
to that button. The script disables NVDA feedback and enables Write 
automation mode.
Same with Read mode.
A script re-enables NVDA's return and changes the mode to Read.
We can trigger anything we want with the buttons on the control surface.
I read the bug report Paul made, it's exactly the same bug I'm talking 
about here.
I hope I was able to help solve this purpose.
Don't hesitate if you have any questions.
Thanks Jammie.

jcsteh:
When playback is stopped and you're in read mode, there are still parameter changes that can happen due to automation; e.g. if you navigate through the project. Also, #802 and your own #813 cannot be fixed without ignoring certain parameter changes. Furthermore, many users (Paul is probably the most vocal) don't seem to be happy with continually toggling on and off surface feedback. They want it to just not report changes caused by automation.

Anyway, as I said, this wasn't merged. You can just install an official snapshot to get the old behaviour back. I'm not pursuing this any further, as I can't figure out how to balance people's various requirements here and I'm kinda done trying to work through this. At this point, I'm very seriously considering just removing surface feedback altogether, as I just can't see a way to make it work the way people want.

jcsteh:
The other thing that is puzzling me is why you want to select a track by touching a fader when in read mode. When in read mode, you can't change the parameter anyway. So what's the use case for touching the fader to select the track? And why report the parameter value when you can't adjust it?

leejul:
In reading mode, the surface is used in the same way as the computer 
keyboard. We can change the volume or pan value, arm or disarm a track, 
same thing with solo and mute. In short, we control Reaper in the same 
way as a computer keyboard. We can even do audio item or MIDI editing. 
This is why we must always know which track is selected when we touch a 
fader and what is the pan/volume value.
I have a French-speaking user of the X-Touch who hardly uses the 
computer keyboard in Reaper anymore.
It does everything with the X-Touch.
Like on a computer keyboard, the X-Touch offers modifiers like Control, 
Alt, Shift and Option. You can make the combinations you want with any 
button, pan or fader and assign an Reaper action. In addition, you also 
have the up and down arrows, which allow you to select the next or 
previous track, left and right to select the next or previous item.
You can also add up to 2 X-Touch extension device which adds 8 strips 
per extension which ultimately gives 24 strips.
Since you integrated the Report changes made functionality via control 
surfaces, sky is the limit with the X-Touch.
We can never thank you enough for that.
Do not hesitate if it lacks clarity.
Thanks Jammie.

leejul:
I can understand users who don't want to have feedback from NVDA while 
writing an automation. But certainly not in Read mode. It doesn't make 
sense. In addition, to enter Write mode, users must press a button to 
switch modes in any case. So you would have to explain to me what is the 
relationship with the Read mode. Read mode must absolutely be vocalized. 
Users who disagree with this, I think they have not understood the 
usefulness of a control surface with Reaper. Again, if you do the 2 
actions I mentioned to you, I can modify CSI to make NVDA no longer 
speak when we switch to Write mode. So everyone will be happy. I do not 
see the problem.

jcsteh:
My understanding from some users is that the reverse is true. They want feedback in write mode so they know what they're changing. If automation is playing, though, as it does in read mode, you cannot change parameter values. They will just snap back to whatever value was recorded in automation.

Anyway, there's clearly a difference in use cases here and a gap in communication we clearly can't resolve. I think discussing this any further is just going to lead to frustration. In short, I give up.

jcsteh:
Okay, I can give you a very simple example to show what I mean about read mode:

1. Set automation to read.
2. Arm the volume envelope.
3. Press alt+upArrow (Track: Nudge track volume up).

Observe that you cannot change the value. It stays where it is. That makes sense because the level is now being read entirely from the envelope. Any changes that occur in this mode are thus changes from automation or changes caused by automation mode switches, which just aren't useful to the user.

I really don't know how I can explain this any better.

leejul:
I really don't understand where the problem is.
Simply reactivate NVDA's feedback to Read and Touch mode, add 2 new 
OSARA actions and users can configure CSI as they want according to 
their personalized configuration.
I will be able to modify CSI so that users can get the behavior they 
want with their control surface.
So this will answer all user requests.
The problem will then be solved.
I think the user should be given full freedom to choose how NVDA 
vocalizes with the control surface.
If you don't agree with this proposal, I can't do anything at my level.
But know that I can easily make changes in CSI that would customize the 
NVDA feedback of the control surfaces.
I just need you to add 2 actions for us.
Thanks again Jammie.

jcsteh:
The concern you raised in #813 would seem to disagree with this. The fix here is the same fix we would need for that. Again, it's just a parameter change. By your logic, you should just turn off surface feedback to avoid the issue in #813 and then turn it back on afterwards. That approach just doesn't scale.

There is just no good solution here.

jcsteh:
Also, you keep saying you want this feedback reactivated. I stress again: it was never changed in official snapshots. This change was never merged. Go download an official snapshot and you'll have things back the way they were.

leejul:
Yes I know, sorry, I meant enable feedback, not re-enable, sorry that's 
my mistake.
I'm talking about the official version of OSARA. Not the test version.

PR #810, opened 2022-11-17
Add feedback for "Transport: Toggle repeat" in midi editor. fixes #809

### LINKED ISSUE #809: Repeat on/off toggle isn't announced in midi editor
In the TCP the repeat state is announced when pressing control R.  In the midi editor, pressing control R does toggle the repeat state but it isn't announced.  It would be good if it was announced.

RDMurray:
fixes #809

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1227,7 +1227,7 @@ void reportRepeat(bool repeat) {
 }
 
 void postToggleRepeat(int command) {
-	reportRepeat(GetToggleCommandState(command));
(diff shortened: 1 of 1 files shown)


RDMurray:
No idea what's going on. The AppVeyor log doesn't help, there are no errors until 
`scons: building terminated because of errors.`

RDMurray:
AppVeyor seems to be suppressing stderr output for some reason. The build fails here with the latest visual studio build tools (17.4.1): several errors like 
```
error: unqualified call to 'std::move' [-Werror,-Wunqualified-std-cast-call]
```

PR #812, opened 2022-11-17
Remove trailing zeros and scientific notation when reporting volume and zoom level. fixes #21

### LINKED ISSUE #21: Truncate excess zeros when announcing volume, pan and pixel values
In OSARA currently, when adjusting the volume of a track to -4 DB for example, OSARA reports "-4.00 db". It would be ideal to have any extra zeros trimmed to cut down on speech chatter, so instead it reports "-4 db". 
This also occurs when zooming in/out of a project. OSARA might say "150.000 pixels" instead of "150 pixels".

However, if a value has non-zeros, report them. For instance, "-4.05 db" or "-4.11 db"

RDMurray:
Please test and give me your opinion. 
##questions:
- Should it report "+" for positive gain?
- are the number of decimal places okay?

fixes #21

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -625,7 +625,7 @@ class ParamsDialog {
 
 	public:
 
-	ParamsDialog(unique_ptr<ParamSource> source): source(move(source)) {
+	ParamsDialog(unique_ptr<ParamSource> source): source(std::move(source)) {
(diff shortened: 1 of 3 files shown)


Lo-lo78:
Yes. It works well.
👍

jcsteh:
Looks great to me. Thanks.

PR #818, opened 2022-12-13
In the MIDI editor, report times using ticks if the MIDI editor is displaying times thus.

### LINKED ISSUE #817: Report position in bars, beats and ticks when moving in the MIDI editor
Think I've found a MIDI Editor preference that OSARA reports aren't respecting yet. It would be useful if they did.
1. Open the MIDI Editor, go to the Options top menu, then Time format for ruler, transport, event properties submenu, choose Measures.Beats.MIDI_ticks.
2. Move to a note, notice how OSARA is still reporting percentages instead of Ticks. That's the problem.
3. Move by grid, same dealio.
4. Now hit Control+F2 on a note for the Event Properties dialog, observe that measures.beats.ticks is reported as expected in the Position and Length fields.

Can our reporting respect the preference? Its name includes transport, so this seems logical. It would be super useful during fine editing, and also a welcome addition for old-timers and switchers from other DAWs where ticks are still displayed by default.

Thanks in advance.

jcsteh:
Fixes #817.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -700,7 +700,7 @@ void cmdMidiMoveCursor(Command* command) {
 	HWND editor = MIDIEditor_GetActive();
 	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
 	ostringstream s;
-	s << formatCursorPosition(TF_MEASURE);
+	s << formatCursorPosition(TF_MIDI);
 	MediaItem_Take* take = MIDIEditor_GetTake(editor);
 	auto chord = findChord(take, 0, {
 		true,  // start
@@ -987,7 +987,7 @@ void moveToChord(int direction, bool clearSelection=true, bool select=true) {
 	fakeFocus = FOCUS_NOTE;
 	ostringstream s;
 	if (settings::reportScrub) {
-		s << formatCursorPosition(TF_MEASURE) << " ";
+		s << formatCursorPosition(TF_MIDI) << " ";
 	}
 	if (!select && !isNoteSelected(take, chord.first.getIndex())) {
 		s << translate("unselected") << " ";
@@ -1129,7 +1129,7 @@ void cmdMidiInsertNote(Command* command) {
 		}
 	}
 	if (reportNewPos) {
-		s << formatCursorPosition(TF_MEASURE);
+		s << formatCursorPosition(TF_MIDI);
 	}
 	outputMessage(s);
 }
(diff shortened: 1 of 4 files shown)


jcsteh:
I thought this would take me 20 minutes. Instead, it took me over an hour and now I'm up too late. Damn it.

@ScottChesworth, it'd be great if you could take this one for a spin before I merge.

ScottChesworth:
Argh, sorry about the gruntwork man. I did have a look at whether I'd be able to do it before opening the issue, couldn't figure it out.
Movement through MIDI data, by grid, plus movement that happens as a result of note insertion all reporting as expected.
The only slight disparity I've found is that "OSARA: Report edit/play cursor position and transport state" still reports percentages. Seems our action doesn't have a specific MIDI Editor context? I've verified that the ruler is reflecting ticks visually, but up to you to decide how much that matters, I don't know how much extra work would be involved.

jcsteh:
I was wondering whether you'd notice that. In the original version of my code, that did work. However, it also created a bug where if a MIDI editor was open but sitting around in the background (as can happen if you screw around with focus enough), all times (even in the main section) would report with ticks. Given that users might end up in this situation and could be pretty confused by it, I figured this slight disparity was better than the bug. It seems REAPER doesn't have a way to ask for the focused MIDI editor, not just the last active editor.

ScottChesworth:
> It seems REAPER doesn't have a way to ask for the focused MIDI editor, not just the last active editor.
> 
Do we need an API from Cockos? Happy to prod Justin.

ScottChesworth:
> I don't know whether it will work on Mac. That will need testing.
> 
Any tips on agrivating the situation? I'm not having much luck getting a rogue MIDI editor to hang around here.

jcsteh:
Yeah, it's much harder now that OSARA corrects REAPER screwing with the focus whenever you open a dialog or the like. The simplest way is to use NVDA obj nav to force focus to the REAPER main window without closing the MIDI editor.

On Mac, the question is whether this code reports ticks at all in the MIDI editor or whether it always reports the ruler unit from the main section.

pitermach:
Here's what I'm seeing here on Mac:

- In an empty project I create an empty midi item. My time ruler defaults to HMS because I work much more often with sound not music.
- I focus the item and bring up the midi editor with Command-Option-E, VOiceOver confirms that a new window was focused so I also go ahead and change the ruler to show ticks.
- If I now press Command-SHift-J to hear the position or move through chords with left/right, Osara still speaks the position in hours/minutes/seconds, though interestingly if I move through the notes in a chord with up/down I hear tick values.
- If I switch focus away from the midi editor and back by pressing command+Accent twice, then OSARA starts reporting the position in bars/beats/ticks as it should and if I focus the main window again it goes back to speaking hours/minutes/seconds.

TLDR: There seems to be a glitch where OSARA doesn't realize the midi editor got focused after you open it, but switching the focus away and back manually fixes it. Not sure if this was the case before this change - it's not something I payed attention to before. Hope that helps.

jcsteh:
Thanks for testing. As usual, supporting Mac is why we can't have nice things.

jcsteh:
Maybe the next build will help on Mac? If not, I give up.

pitermach:
Yep, that did it. Now the report switches to bars right away when the editor is opened with the edit action.

PR #820, opened 2022-12-14
When moving between chords in the MIDI editor, don't report note counts if report scrubbing is disabled.

jcsteh:
This is regarding https://github.com/jcsteh/osara/commit/973b57fa50dd19f4be72707ba73c0eb2ffeac587#commitcomment-92955201.
I don't entirely understand what you're saying, @ScottChesworth , mostly because i can't be bothered fully getting my head around it. If I've understood you correctly, the change you want is a single line of code. The easiest way to be sure is for you to test this PR. :)

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -992,7 +992,7 @@ void moveToChord(int direction, bool clearSelection=true, bool select=true) {
 	if (!select && !isNoteSelected(take, chord.first.getIndex())) {
 		s << translate("unselected") << " ";
 	}
-	if (settings::reportNotes) {
+	if (settings::reportNotes && settings::reportScrub) {
 		int count = chord.second - chord.first;
 		// Translators: used when reporting the number of notes in a chord.
 		// {} will be replaced by the number of notes. E.g. "3 notes"


ScottChesworth:
This is exactly what I was trying to describe. Hubba hubba, that's some head-turning UX right there! I reckon this is good to go man. Can link people who were in the RWP thread to it if you want more feedback before merging though?

jcsteh:
If it makes you feel any better, I went and read the RWP thread in question and nearly reverted the change just out of spite.

ScottChesworth:
It was indeed a, er, difficult day for diplomacy. Yeah, let's go with that.

ElCrackLoko:
Guys please, I don't feel good about this change. Where can I download a version of Osara that will read me the location time of the midi editor notes when I hover over them with the right or left arrow, even when the speak time option is unchecked? It was fine that he just didn't mention Pixels, but the notes, geez, I have to press Control Shift+J all the time in the midi editor, either that or turn on the Report position when moving cursor option. Or maybe it's my Osara's mistake? Since I have marked the option to announce notes in the Midi editor, but I don't think it's for that, but only for the notes as such when we give the up or down arrow. A hug for everyone.

ScottChesworth:
So you want to keep Report position when scrubbing un-checked in OSARA Config, but you still want to hear full reporting in the piano roll, is that right?
Hmm, tricky because we went with tying chord movement to scrubbing.
Two custom actions would do it.
In main section, combine "Item: Open in built-in MIDI editor (set default behavior in preferences)" and "OSARA: Toggle Report position when scrubbing"
In MIDI Editor section, combine "File: Close window" with "OSARA: Toggle Report position when scrubbing".
If that causes any extra speech chatter around opening/closing, place "OSARA: Mute next message from OSARA" second in each custom action.

Shout if you're not comfortable working with custom actions there, I can make and upload them for you.

ElCrackLoko:
Hi, friend. The truth today I can tell you that the language is a 
barrier; I don't quite understand what I have to modify because the 
Spanish translation that Google translate gives me is not the same as it 
should appear in Reaper. Like everything mixes up. Wouldn't it be too 
much to ask to create an option for it to appear in Osara's options to 
check and uncheck, as well as the others? They were asking me the same 
thing in the community that I manage, PMA, but I haven't been able to 
answer them, that's why I also made the query/request here.

ScottChesworth:
Just realised that my suggestion wouldn't work anyway because components of a custom action can't span multiple sections. At least, there's no way to do that with REAPER's GUI.
So yeah, I guess this will need to become a separate config option. @jcsteh, do you have any concerns around that?

ElCrackLoko:
Please, I would appreciate it as you have no idea if you incorporate this option

jcsteh:
This was discussed in #821. As per that discussion, I'd be curious to understand why you feel that MIDI editing is fundamentally different to scrubbing audio. That is, why do you want no time reports when scrubbing audio, but you do want time reports when effectively scrubbing MIDI?

ScottChesworth:
I've got a working build with it as a separate config option. That work you did on defining settings made it super easy! Gonna send a PR in a sec as this isn't the first time I've heard the request.

ScottChesworth:
Summarizing the rationale I'm hearing most often for a separate option: once you've got your horizontal zoom set the way you like it, scrub moves in consistent increments. Additionally when scrubbing audio, there are often extra positional audible clues specific to the material being scrubbed. For example a user might think "aha, there's a sharp transient, I'm recognizing that as the snare drum because it doesn't have the low end presence I'd expect from scrubbing a kick, so I'm probably near beate 2 or 4 of the bar right now". Whereas moving through chords, there can be huge amounts of movement involved, and if you're a consistent player who cuts long passages at a time, there's often nothing audibly different between one instance of a chord that you played in bar 25 versus the next time you played it at I dunno, bar 50.

ElCrackLoko:
Well, my point of view is the following:
When you record an audio, you generally have a single take, a single piece which is the item, even for editing you have to move a lot with the right and left arrows, being that the current pixel announcement is somewhat unnecessary. But not so with Midi, because, even if you record a take along the way, as Midi is, you will have several pieces in the same Midi Item, which we can move at will in the Midi editor or more thoroughly, in the list of Midi events . So where do I want to go? What for the Midi editor if it is necessary that it informs you the time of each note, so you can quantize it to your liking, or move it to where you want without having to keep pressing Control Shift J all the time. But not so with the audio, if it is dequantized or out of time, it is better to record a new take and thus do not modify or damage the transients. On a personal level since I work with Reaper, I like to know in what time, in what location I have each Midi note. Even if you are like me and work LoFi, ChillHop, RapJazz, where there is a lot of sampling, a lot of use of the Midi editor and precise timing with ReaSamplomatic, greetings.

jcsteh:
Fair enough. I'm happy with a separate option. As always, I just wanted to make sure we're being intentional about a change, not just doing it "because". The latter is the path to bad UX.

ElCrackLoko:
I have always been in favor of change, in fact, the world of music production is complicated from the beginning. And, I think, the more configuration options you have to choose from, the more the scope will expand to the point where you will have diverse configurations and everyone will be happy. Up the changes. I really appreciate that you take this request into account, the truth is that I am not used to working activating the option to announce the movement of the cursor with Control F12 when I enter the Midi editor or deactivating it when I leave the piano roll to go to edit audio. If a user has an unfavorable experience, he can always leave his point of view here on Github, like I have.

jcsteh:
I don't agree that more configuration options necessarily leads to a better experience. Sometimes, it just makes things more confusing without necessarily providing a real benefit, especially for new users. This case appears to be an exception and I'm okay with that. Nevertheless, I'm always going to ask people to justify why they think a particular use case warrants a new option. I realise this seems painful, but this is about making the best possible experience for all users, including new ones.

ScottChesworth:
@ElCrackLoko, update your OSARA and you'll find a new option: "Report position when navigating chords in MIDI editor". Happy new year from London to the PMA people! :)

ElCrackLoko:
Thank you very much, I tried it and it works as expected. I know that several members of PMA will be fully grateful to them, it was something that was really missed; What a good service! Happy New Year to you too, greetings to London

PR #828, opened 2023-01-03
Only eat WM_KEYDOWN VK_CONTROL messages when used to cancel a note preview

cfillion:
OSARA currently unconditionally prevent the system from passing any `WM_KEYDOWN` messages for the Control key to the rest of the hook chain or the target window procedure.

This affects ReaImGui which expects to receive these messages. (The issue can be duplicated via the ReaImGui_Demo.lua script under Inputs, Navigation & Focus -> Keyboard, Gamepad & Navigation State: after installing OSARA, the LeftCtrl and RightCtrl keys are not triggered.)


jcsteh (approved):
Makes sense. Thanks.

PR #829, opened 2023-01-06
Report named MIDI notes when received via MIDI input.

### LINKED ISSUE #826: Wishlist: a plug-in, script or feature to automatically report key switches on the fly during composition
I'd like to explore whether there's any potential around getting key switches to be reported automatically, most likely in conjunction with REAPER's template and named notes features. Think of this as a place to collect/discuss thoughts/ideas. I'm definitely not smart enough to implement this, I'd just love to use it. Tagging my man @TimTam at the top, maybe he'll be able to help with the technical.

## Scenario:
Key switches are a widely adopted method for controlling articulations when using sample-based instruments. With orchestral libraries in particular, judicious use of key switches is often an essential part of emulating realism. But here's the rub, there's no standardization around where/how key switches are assigned from instrument to instrument, right now, the standards I've observed are that 1) modifying the assignments is typically inaccessible and 2) remembering switches per instrument is incompatible with the small portion of my brain that stays active during composition.

## Current workarounds:
1. Some folks take notes. That's a good strategy in terms of pooling resources, but there's no standardization around that either and checking documentation mid creative flow can be disruptive.
2. Real mates share tem-plates. More work on the front end of course, but by using REAPER's named notes feature (usually on dedicated key switch tracks), reporting can be integrated into the project. However, that still involves the user taking their hands off the controller, opening the MIDI editor and scrolling the pitch cursor. For a scatter-brained part-timer prone to imposter syndrome such as myself, anything that interrupts notes being played gives me time to think about what I'm making, which is inherently risky.

## Exploration of how auto-reporting could work:
1. I'm not seeing this being a core OSARA feature, mainly because I came up empty when trying to imagine how it would work in conjunction with what we currently provide. Decided to float it here though as this is a good place to reach composers who also have coding/scripting chops. If I've missed an idea here, let me know.
2. My first thought was a ReaScript that could be fed a note range and use OSARA's OutputMessage function to report note names within that range. This would go hand in hand with the description of templating above. But then, a scripted approach isn't sitting right because I'm not sure how that would work for multiple instruments/libraries within a project, plus preservation of imported scripts around key map updates is a bit shakey. Seems like there'd be very real chances for less technically inclined composers to get caught short.
3. A js plug-in reporting through OutputMessage instead, maybe? My thought is that this could have a minimal accessible GUI for configuration and be incorporated into shared track templates wherever relevant, thus tailoring the functionality to each instrument/library. This could be provided both as a standalone download and through ReaPack, our snapshots page and Toni's ReaPack repo would be good spots.

## Who can do what?
I'd test the heck out of anything in this area, would help with updating existing templates, could potentially sponsor the work, and would gladly put time into making a free training resource available. Can also contribute childlike enthusiasm during ungodly hours of the night (if that doesn't get you gunning for a finish line, nothing would).  :)

Any and all thoughts/ideas appreciated. Let's bat it around in the comments, I'll update this top post if something starts to take shape.

### LINKED ISSUE #930: Would it be possible for OSARA to report MIDI notes when pressing keys in the Virtual MIDI Keyboard?
This has just been requested on RWP by someone who's struggling to learn the layout. I think there's a valid use case because we don't have any other form of landmarking available, said I'd ask if there was any way it could be done.

jcsteh:
This should be considered a proof of concept at this stage. The UX hasn't really been discussed yet, but I thought it'd be useful to get a prototype out there so people can play with it and see where it can go.

The implementation is currently very simple. When a MIDI note is pressed via a MIDI input, if the current track has an explicitly defined name for that note, report that name.

Questions/considerations:

1. I guess reporting any named note isn't always ideal. For example, this could be really annoying for a drum synth which names its notes. But how do we determine which names to report?
    - I can't really think of a place where we could specify a range of notes to report for the track.
    - One idea I had was to use some character or sequence of characters in the note name to specify a name which should be auto reported. This could be something like "key switch", but that's not great for translation. It could be something as simple as an asterisk "*", but that's kinda cryptic. We could strip out these characters when auto reporting to avoid annoying verbosity.
2. Which track should we use to get the note names? The current track is easy, but perhaps it isn't ideal in some cases. I guess we could walk through all tracks to find the first armed track which is getting input from the triggering device and channel (or from all devices/channels). That could really hurt performance, though, since we'd have to walk all tracks whenever a note is played.
3. Should this be behind a config setting? I guess it probably should.

Fixes #826. Fixes #930.


ScottChesworth:
Wowzers, didn't expect you to start prototyping so soon, thanks! Thoughts coming up in line:

> 2. Which track should we use to get the note names? The current track is easy, but perhaps it isn't ideal in some cases. 
> 
IMO current track would do fine. Leave the choices of which track is to be reported and when with the user. Trying to anticipate this across multiple tracks for all of the different templating styles out there will be such a slippery slope man, I can't imagine that juice being worth the squeeze.

> I guess we could walk through all tracks to find the first armed track which is getting input from the triggering device and channel (or from all devices/channels). That could really hurt performance, though, since we'd have to walk all tracks whenever a note is played.
> 
Nah, I'm pretty sure minimizing the potential performance hit will be key in this case even if that means the feature requires a little extra legwork from the user. Most folk who are gonna use this reporting will be running sample-based instruments and those projects can get real heavy on resources real quick, we don't want OSARA causing any negative impact in those environments.

> 3. Should this be behind a config setting? I guess it probably should.
> 
Seems sensible yeah. Perhaps something like "Automatically report named MIDI notes for currently selected track"? Unchecked by default I think.

ScottChesworth:
Would it be easy for this to check whether the currently selected track is armed? During initial testing, it seemed a bit unintuitive that I was hearing keyswitches being reported even though the track was unarmed, IE, pressing the keyswitches didn't have any effect.

jcsteh:
> Perhaps something like "Automatically report named MIDI notes for currently selected track"? Unchecked by default I think.

Sounds reasonable, though we should figure out question 1 (which notes to report) first, as that might impact the name of the setting.

> Would it be easy for this to check whether the currently selected track is armed?

Piece of cake. Done in 8a0f0f147e729adcba96574ead98d3327eb5a1a8.

mattymc2:
You’re a wizard James. Can’t wait to try this Sent from my iPhoneOn 7 Jan 2023, at 9:49 am, AppVeyorBot ***@***.***> wrote:﻿
Build succeeded! Build osara pr829-1049,d72dc78a completed (commit d72dc78a0d by @jcsteh)
Downloads:

Windows

installer/osara_pr829-1049,d72dc78a.exe
locale/osara_pr829-1049,d72dc78a.pot



Mac

installer/osara_pr829-1049,d72dc78a.dmg

ScottChesworth:
> we should figure out question 1 (which notes to report) first, as that might impact the name of the setting.
> 
The extra control that your suggestion of auto-reporting based on a character in the note name would provide is appealing to me. A couple of hesitations though:
1. It puts an unusually high knowledge requirement in the way of the feature starting to work. Is that a worthy trade for more granular control?
2. It'll decrease the amount of templates that are compatible with the feature, at least initially. Also, users would need to re-customize templates from any sources that don't adopt our choice of prefixed character/characters any time their templates are updated.

> Piece of cake. Done in [8a0f0f1](https://github.com/jcsteh/osara/commit/8a0f0f147e729adcba96574ead98d3327eb5a1a8).
> 
Notice how quickly @mattymc2 appeared after you said cake? Hold on, gonna try a trick that'll get rid of him.
So anyway, lately I've been binging Australia's Got Talent on YouTube... :)

mattymc2:
Aaaaaaand I’m gone Sent from my iPhoneOn 7 Jan 2023, at 4:12 pm, ScottChesworth ***@***.***> wrote:﻿

we should figure out question 1 (which notes to report) first, as that might impact the name of the setting.

The extra control that your suggestion of auto-reporting based on a character in the note name would provide is appealing to me. A couple of hesitations though:

It puts an unusually high knowledge requirement in the way of the feature starting to work. Is that a worthy trade for more granular control?
It'll decrease the amount of templates that are compatible with the feature, at least initially. Also, users would need to re-customize templates from any sources that don't adopt our choice of prefixed character/characters any time their templates are updated.


Piece of cake. Done in 8a0f0f1.

Notice how quickly @mattymc2 appeared after you said cake? Hold on, gonna try a trick that'll get rid of him.
So anyway, lately I've been binging Australia's Got Talent on YouTube... :)

jcsteh:
Yeah, I agree the special character thing is not ideal. I don't have any other ideas though. Maybe we should just leave it as a setting and people can just flip it on and off as they need it? Now that all settings are exposed as toggle actions, that isn't so difficult for people who need to toggle it frequently.

ScottChesworth:
Fine with me man. Thanks.

jcsteh:
I've added a setting now. Let's see how this works in situ.

I do have a clarifying question about the UX here. I understand that looking at documentation or some other source for key switches interrupts flow. On the other hand, stabbing various keys to see if OSARA reports anything also feels pretty ugly and time consuming. Wouldn't the change of the sound caused by the key switch be a more natural way of understanding whether you found the right key switch? I almost wonder whether it'd be better to have an action to report all named notes on the current track, since then you're not just stabbing arbitrary keys until you find the right one, although hearing C2, D2, etc. isn't always easy to map to physical keys.

ScottChesworth:
Now that they can speak, I'm locating keyswitches here by laying four fingers across white keys from C to F and banging on those while I scroll through octaves. Utterly graceless, but it locates stuff quicly. This would be less necessary with a controller that has more keys, but I'm cheap. :)
However, an action to report first and last named note on the current track would be super useful if it reported the named note then standard MIDI note. Reporting the range would tell you whether the keyswitches were confined to an octave at one end of the board or whether it's worth exploring both ends outside of what's playable.

ScottChesworth:
Example reports:
"Legatto C0, some other fancy musical term that I'm too drunk to spell C1". This tells you to scroll octave left.
"legatto C0, some other thing G5". That one tells you there's stuff to explore by jumping octaves in both directions.

jcsteh:
Why is having the note names reported on input preferable to just having an action to report the notes, though? Is it purely about not being able to locate the right octave physically or is there some other advantage to reporting them on input (keeping in mind that a key switch will alter the sound that plays inside the playable range)?

ScottChesworth:
It's about finding those keyswitches being a PITA on small controllers, because more often than not, the change isn't verifiable without scrolling octaves to try a note, scrolling to play, scrolling to try etc. Hmm, is your idea of an action to report named note range a more ellegant solution to the problem then?

jcsteh:
I was originally suggesting an action that reported all named notes rather than just a range, but I guess that could get nasty if an instrument had a tonne of them.

Lo-lo78:
Sorry if I didn't understand the thread well.
😉
What I understand is that we want there to be an indication of the midi input stream to make sure that the midi is active.
Is that it?
Unfortunately midi is a protocol that has not only 127 notes but many events.
If I understood correctly we would need a signal to understand if the midi connection is online.
A word saying that Reaper is receiving the midi would suffice.
For me this would already be great!
Very useful when using CSI hardware and other hellish toys.
Greetings!

ScottChesworth:
> I was originally suggesting an action that reported all named notes rather than just a range, but I guess that could get nasty if an instrument had a tonne of them.
> 
Some orchestral libraries do have a lot yeah. Hmm, I don't know how much value there is in OSARA providing a potentially long report on this when users can already scroll the pitch cursor to hear everything that's been named. With speech interupt already being available as they scroll through choices, it seems like that would be a more productive way to collect a full rundown of what's available. For me, the use case around hearing the first and last named notes is just to jog my memory without both hands needing to leave the controller.

ScottChesworth:
Maybe @timtam can weigh in on this? I don't know whether he struggles to remember which keyswitch does what and where they're assigned like I do, but I do know he uses them a lot.

ScottChesworth:
> Can’t wait to try this
> 
Have you had a go with it yet Matt? Any thoughts on what you'd find useful if we add a report action?

ptorpey:
I’m wondering how sighted people manage what you are trying to accomplish.  Would that give any insight as to how a UX for folks using screen readers should be designed?  If I am understanding this correctly, it sounds like one would have to customize this for each instrument?

I’ve started using the Notes capabilities of Reaper to keep notes about a project with a particular project.  That probably wouldn’t help you with what you want to do, but I did see an action:

SWS/S&M: Open/close Notes window (global notes)

Which seems to be associated with global notes, presumably accessible across projects.  

 

I set a hotkey to quickly open my project notes and one could set a hotkey to quickly open the global notes to quickly see what keys were assigned for each instrument by doing a search.

 

I’m guessing that one could put all of the special keys for each instrument into that notes file and have it available anywhere.

 

Anyway, just an idea.  Don’t know if that helps or gives you any other ideas.

 

--Pete

ScottChesworth:
> I’m wondering how sighted people manage what you are trying to accomplish.  Would that give any insight as to how a UX for folks using screen readers should be designed?
> 
Keyswitch assignments are shown as part of the GUI of each instrument, most often illustrated with highlighted notes on a keyboard. So nope, nothing we can take cues from.

jcsteh:
Honestly, I don't know. I implemented a thing for #826, but I don't really know whether it's that useful. Since i worked on this, a related issue #930 was filed which would effectively need all MIDI notes to be reported, not just named ones. So now I'm wondering whether we should just have a setting to report all MIDI notes that people flip on and off as desired.

LeonarddeR:
This now reports when armed. Just a suggestion, might it make sense to bind this to record monitoring instead? That requires arming anyway.

ranetto:
Hi everyone, I think this is very useful both for reporting the notes and any CC controllers that come out of an unused midi controller like CSI, or simply also to check if a synthesizer sends something out when for example it executes a sequence or a pat of battery. I would do a dare action that chooses whether to turn on this setting when arming the track or turn it off, this would avoid having another software such as midi monitor. I have often noticed that both master and other synth keyboards send continuous controllers which can annoy reaper or in any case fill the midi buffers when we use them to record and this, without an external editor, is the only way to verify this. By being able to listen to the midi notes or in any case a controller with its value, we could also check what a komplete sends or what a nano kontrol 2 or other used ones sends.
It would also be used to check what a midi trig outputs while we trigger the drums.
So I would propose a simple dare action that decides whether to say what comes only from the armed and selected track or not. Thanks always to everyone for the work done.

LeonarddeR:
How about a toggle that allows you report either nothing, notes, or everything that's coming into a track?

ranetto:
I'd say it's perfect thanks

LeonarddeR:
@jcsteh Do you have time to work on this further, or do you want me to have a try?

ScottChesworth:
> This now reports when armed. Just a suggestion, might it make sense to bind this to record monitoring instead? That requires arming anyway.
> 
It makes more sense to report when armed IMO, because 1) there are still a bunch of people who prefer the pre REAPER 6 behaviour where monitoring wasn't automatic, and 2) some people use external hardware that they monitor direct, IE they probably wouldn't think to turn on REAPER's record monitoring.

ScottChesworth:
> How about a toggle that allows you report either nothing, notes, or everything that's coming into a track?
> 
So a cycle rather than a toggle? I know we can write the action to cycle those choices, but what would the GUI look like in OSARA Config dialog? So far as I know, we can only easily put check boxes in there.

Timtam:
Don't we have combo boxes in the peak watcher config dialog as well already?

ScottChesworth:
Oh yeah, fair point.

ScottChesworth:
@LeonarddeR , how much work is it gonna be to get this rebased against master? I'm thinking now discussion has picked up again, it would be great to have a workable build people can experiment with.

LeonarddeR:
> It makes more sense to report when armed IMO

Makes sense

> So a cycle rather than a toggle? 

Yes, correct.

> I know we can write the action to cycle those choices, but what would the GUI look like in OSARA Config dialog? So far as I know, we can only easily put check boxes in there.

I'd say either a combo box or radio buttons. We need to write a macro for it though, just like the BoolSetting one. I'm very uncomfortable with that, but it's never too late to learn.

> @LeonarddeR , how much work is it gonna be to get this rebased against master?

Not that much actually 😉

jcsteh:
I can work on this, but it's still unclear to me what exactly we want. The proposed multi choice doesn't separate notes from named notes, for example. Also, reporting CCs is going to get very verbose very quickly, possibly uselessly verbose. Remember that turning a knob or pressing a pedal usually fires many changes in quick succession, not just one.

Adding multi choice controls to the config dialog is going to be tedious, but not impossible. The config framework is set up to allow toggles to be added very easily. There's no way to add multi choice currently. All of that said, I'm not entirely convinced that mutually exclusive multi choice is what we want here, given the unresolved question about notes vs named notes and putting CCs on top of that.

jcsteh:
For reference, adding a new type of setting to the config dialog requires 6 different definitions of a new macro, plus support in site_scons/makeConfigRc.py. I also have no idea how we can express both enumerated names and translated names for each value of the setting within a macro.

ptorpey:
Also, reporting CCs is going to get very verbose very quickly, possibly uselessly verbose. Remember that turning a knob or pressing a pedal usually fires many changes in quick succession, not just one.

[PT] This isn’t a feature I would necessarily use…Just following the discussion.  But one idea for the many CC events being spoken is to do what some screen readers do with a line of 80 “asterisks”, or “x’s”, in a row.  Maybe just speak a few (like 2 or 3) if there is a long contiguous string of such events.

jcsteh:
We are going to try having all MIDI notes reported but only while shortcut help is enabled. We can't block the notes from reaching REAPER, but most of the time, this won't matter. We will probably need to disable "MIDI inputs as step input" while shortcut help is enabled and re-enable it afterwards.

jcsteh:
This is now done, so all notes will now report if the current track is armed, but only while shortcut help is enabled. I'm not touching CCs here; we can handle those separately if desired.

ranetto:
thanks for the work done so far, I think it would be very useful to be able to implement a code that also shows us the controllers received, especially today when many external surfaces are used, many synths capable of giving us automation controls in their sequences, many cards audio that can be driven via plufins such as uad which can provide controller output or more simply to see when a keyboard sends those annoying continuous controllers which needlessly invade the midi lanes being recorded in reaper. Thanks if you can do that it would be very helpful. I had to make myself a little software as a midi monitor and I can't wait to be able to no longer use it but only use Osara. Thank you!

jcsteh:
As I said, I'm not working on CCs in this pull request. Aside from anything else, I'm not convinced this is a need that is specific to screen reader users. For notes, the screen reader specific use case is a lot clearer: notes are often marked with colour or other indication in software or hardware, and the virtual MIDI keyboard has a visual interface which shows the user which keys on the keyboard map to which MIDI notes. For CCs, everyone, sighted or blind, is pretty much in the same boat: they don't necessarily know which knobs map to which CCs. If you can provide a satisfactory counter-argument, you're welcome to file another issue to request this. For now, I'm not going to block this any further on implementing something which is questionable and a great deal harder to get right.

jcsteh:
@ScottChesworth, have you had a chance to try this? It's done as far as I'm concerned. I was just leaving it open to gather any feedback before merging. As noted above, I'm treating CCs as out of scope for now.

ScottChesworth:
Tried on Windows here and Jen gave it a wangle on Mac, very excellent! One last thought, seeing as we have conditionals toggling step input now, could that be reported when it happens? Just thinking it's a change to the REAPER native workflow isn't it. Maybe OSARA says something like:
- "paused step input, shortcut help enabled"
- "shortcut help disabled, resumed step input"?

jcsteh:
I mean, shortcut help in general is a change to the REAPER workflow. Any action executed while shortcut help is enabled gets swallowed. I know step input feels different because we haven't swallowed that before, but I'd argue it's the same from a user perspective.

ScottChesworth:
Seems unlike OSARA to adjust a setting and not report it is all. I don't feel strongly enough about it to hold the merge up though.

jcsteh:
Explicit is usually better than implicit, so I'll concede on this one.

PR #831, opened 2023-01-09
Add new "Report time movement in MIDI editor" option in OSARA Config

ScottChesworth:
unlinked functionality from "Report position when scrubbing" and updated ReadMe.


jcsteh:
I worry that "time movement" might be a big vague. Technically, moving by bar/beat is also "time movement". What about "Report position when moving to chords in the MIDI editor" or similar?

ScottChesworth:
Good point. I'll tweak the wording.

PR #833, opened 2023-01-14
Save Peak Watcher settings with the project.

### LINKED ISSUE #100: Add the ability to save peak watcher configuration
I often find myself re-configuring the peak watcher every time I restart Reaper, and it's usually configured to watch the last touched and master tracks. It would be nice if peak watcher configuration settings could be saved.

I would recommend we add two radio buttons in the peak watcher dialog to select whether the settings are global or per project. If global, only master and last touched tracks are available in the combo boxes. If per-project is set, then the combo boxes behave as they do today. When OK is pressed, peak watcher settings are either saved to the global configuration (just as Osara main preferences are), or saved to the currently open project.

After a quick glance at the Reaper API documentation, it wasn't immediately obvious to me how to be notified when a new project is loaded and restore its per-project peak watcher settings, but it seems as though it should be possible.

This raises the issue as to whether it would be generally useful to be able to save Osara main configuration per project, but I'll leave that to another issue.

jcsteh:
This is not super well tested (and not tested at all in real usage), so I'd appreciate help with that.

Those that want default settings for new projects can configure a new project template.

Fixes #100.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -584,6 +584,9 @@ You can also quickly pause Peak Watcher using OSARA: Pause/resume Peak Watcher.
 While paused, Peak Watcher won't notify you of any level changes.
 You can later use the same action again to resume automatic reporting.
 
+Peak Watcher settings are saved with the project.
+To configure default settings to use for new projects, you can save them in a project template and configure this template to be used as the template for new projects in the Project section of REAPER Preferences.
+
 ### Shortcut Help
 It is possible to have REAPER list all shortcuts and to search for individual shortcuts in the Action List.
 However, it is sometimes convenient to be able to simply press a shortcut and immediately learn what action it will perform.
--- a/src/osara.h
+++ b/src/osara.h
@@ -201,6 +201,7 @@
 #define REAPERAPI_WANT_GetTrackGUID
 #define REAPERAPI_WANT_guidToString
 #define REAPERAPI_WANT_stringToGuid
+#define REAPERAPI_WANT_GetCurrentProjectInLoadSave
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
 
--- a/src/osara.h
+++ b/src/osara.h
@@ -198,6 +198,9 @@
 #define REAPERAPI_WANT_TrackFX_GetParamFromIdent
 #define REAPERAPI_WANT_TrackFX_GetNamedConfigParm
 #define REAPERAPI_WANT_GetItemStateChunk
+#define REAPERAPI_WANT_GetTrackGUID
+#define REAPERAPI_WANT_guidToString
+#define REAPERAPI_WANT_stringToGuid
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
 
--- a/src/peakWatcher.cpp
+++ b/src/peakWatcher.cpp
@@ -2,7 +2,7 @@
  * OSARA: Open Source Accessibility for the REAPER Application
  * Peak Watcher code
  * Author: James Teh <jamie@jantrid.net>
- * Copyright 2015-2022 NV Access Limited, James Teh
+ * Copyright 2015-2023 NV Access Limited, James Teh
  * License: GNU General Public License version 2.0
  */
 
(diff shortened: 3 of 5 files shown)


jcsteh:
Stupid... Mac... gah!

jcsteh:
In case anyone is interested, this took me 2 hours and added 150 lines of code.

ScottChesworth:
> this took me 2 hours and added 150 lines of code.
> 
There's no such linear measurements for gratitude, so for now I'll just say O-M-Actual-G thaaaaaaankyooooouuuu.

mattymc2:
Plus 1 to this.  What can I even say!! Will test at work on Monday. Sent from my iPhoneOn 14 Jan 2023, at 11:27 pm, ScottChesworth ***@***.***> wrote:﻿

this took me 2 hours and added 150 lines of code.

There's no such linear measurements for gratitude, so for now I'll just say O-M-Actual-G thaaaaaaankyooooouuuu.

Lo-lo78:
> In case anyone is interested, this took me 2 hours and added 150 lines of code.

Congratulations James!
It works perfectly!
👍👍👍

ScottChesworth:
Not having any luck saving a default template with Peak Watcher incorporated here yet. Could be user error, here are the steps I've taken:
1. Open fresh project, configure second watcher to report short-term LUFS at -12 on master, holding for 5 seconds.
2. Go to File, Project templates, Save project as template.
3. Set as default in Project category of REAPER Prefs.

Now when I reload the template (either from File/Project templates or by hitting the Default option in the Project Templates section of REAPER's optional startup prompt screen), the Cockos js loudness meter is instantiated on my master with Short term LUFS enabled and all other units set to off in its parameter, all as expected. However, second watcher reports "not configured" in the context menu, no settings are retained when I load its configuration dialog.

I also tried the steps above saving settings for peak dB with first and second watcher just to see whether it was the LUFS metering throwing a spanner in the spokes, but no dice.

ScottChesworth:
Yay, that got it, thanks! One last question, how much work would it be to have Peak Watcher configurations included if the user sets stuff up, then goes to Project Settings and hits "Save as default project settings"? Just asking because that's a widely adopted approach for choosing defaults. If it's too much work though, we can put out a tutorial explaining this approach.

jcsteh:
That'd be nice, but as far as I know, there's no notification from REAPER for that save, which is why I designed it this way instead. If there were such a notification, OSARA would need entirely separate code to save and load for that case, since it'd have to save somewhere outside a project.

ScottChesworth:
Gotcha. I'll rustle up a demo as things are then.

RDMurray:
Impressive. I estimate it would have taken me at least twice as long!

Works great here as long as there is only one project tab. It would be good  to store settings for each open project. Currently it breaks when switching tabs.

Maybe replace watchers with a map from `map<ReaProject*, array<Watcher> >`. 

the function EnumProjects can retrieve the currently focused project.

jcsteh:
How do you feel about handling multiple projects in a separate pull request? If I understand you correctly, this would already be broken even without storing settings.

I guess this has the potential to screw up saved settings in any project you open with multiple tabs, though. So maybe support for multiple projects needs to come first before we try to merge this. I feel like that's going to be a bit painful - there are a bunch of places which assume a single project - so I'm probably not going to have time to deal with that for a while.

jcsteh:
I also don't know how to detect when a project is closed, which would mean we'd keep data for dead projects hanging around indefinitely. That seems bad.

RDMurray:
I've got time tomorrow, so I'll try to implement multiple project support in a separate PR.
as for the data for dead projects, I was thinking we could garbage collect them on project load.

jcsteh:
Gc on project load could work. It doesn't handle the case that someone had 10 projects open and closed 9 of them without opening any others, but that seems like a pretty obscure use case.

Doing this on project load will make that dependent on the work done in this PR, since we didn't have the projectconfig registration before that.

RDMurray:
I just checked SWS. They do it in `project_config_extension_t::BeginLoadProjectState`.

Okay, should I base it on this PR and then they can both be merged together?

RDMurray:
What I came up with is in #837 . 

[Changes from this pr](https://github.com/jcsteh/osara/compare/peakSave...RDMurray:peakTabs)

PR #835, opened 2023-01-15
When moving through chords in the MIDI editor, only play note preview  and report note count if the cursor position changes.

### LINKED ISSUE #830: When moving through chords in piano roll, only play note preview if cursor position changes.
@ptorpey asked on RWP whether we can clarify the UX around reaching the end of navigable notes in the piano roll. This would also apply to the start when navigating to previous chord.
Our reporting style is consistent already, if the cursor position doesn't change, OSARA doesn't report. However, note previews do play irrespective of whether the cursor position has actually changed. Navigating chords with our new config options to slim down the amount of reporting exasipates this inconsistency.

Do we have enough control over when note previews play to bring those in line with our reporting style? Ideally we'd only hear a preview when navigating to previous or next chord results in the cursor position changing.

jcsteh:
@@ptorpey, @ScottChesworth, please take this for a spin and let me know if it's what you're after.

Fixes #830.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -971,6 +971,7 @@ void moveToChord(int direction, bool clearSelection=true, bool select=true) {
 		MIDIEditor_OnCommand(editor, 40214); // Edit: Unselect all
 		isSelectionContiguous = true;
 	}
+	const double oldCursor = GetCursorPosition();
 	// Move the edit cursor to this chord, select it and play it.
 	bool cursorSet = false;
 	vector<MidiNote> notes(chord.first, chord.second);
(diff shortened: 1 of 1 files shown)


ScottChesworth:
That's exactly what I was After, thanks! Tested it with Report full time checked as well, it's useful, tells you instantly that there's not gonna be anything new at the end of the next time report.

ptorpey:
Works nicely here also with and without the option to report full time movement commands and/or report position in midi editor when moving by chords.

 

This is very helpful for knowing when you’ve reached the beginning or end of an item in the midi editor regardless of Osara options.

 

Thanks.

 

P.S.  Microsoft Defender did not like downloading or running the last two builds for some reason.  Most times it works, but occasionally there is a build that MS doesn’t like.  Who knows why some builds are different.

 

--Pete

PR #837, opened 2023-01-18
Add support for multiple project tabs to peak watcher

RDMurray:
Currently draft awaiting merge of #833 

testing appreciated.

Currently this only reports for the active project. Is it desirable to have the possibility to still hear automatic reports from other project tabs?

I also left the pause functionality global. Should it be remembered per tab?

### DIFF
--- a/src/peakWatcher.cpp
+++ b/src/peakWatcher.cpp
@@ -13,6 +13,8 @@
 #include <cassert>
 #include <variant>
 #include <vector>
+#include<map>
+#include<array>
 #include <utility>
 // osara.h includes windows.h, which must be included before other Windows
 // headers.
@@ -50,6 +52,7 @@ T& varGet(V& var) {
 class Watcher;
 bool isWatchingAnything();
 void stop();
+bool isPaused{false};
 
 // Peak Watcher can watch various types of levels; e.g. peak dB, momentary LUFS.
 struct LevelType {
@@ -76,6 +79,10 @@ struct LevelType {
 		return a > b;
 	}
 };
+
+ReaProject* currentProject(){
+	return EnumProjects(-1,nullptr, 0);
+}
 
 void describeTrack(MediaTrack* track, ostringstream& s) {
 	int trackNum = (int)(size_t)GetSetMediaTrackInfo(track, "IP_TRACKNUMBER",
@@ -167,7 +174,7 @@ class Watcher {
 };
 
 const int NUM_WATCHERS = 2;
-Watcher watchers[NUM_WATCHERS];
+map<const ReaProject*, array<Watcher, NUM_WATCHERS> > watchers;
 
 const char* WATCHER_NAMES[NUM_WATCHERS] = {
 	// translate firstString begin
@@ -366,7 +373,7 @@ void Watcher::description(ostringstream& s) {
 }
 
 void resetWatcher(int watcherIndex, bool report=false) {
-	Watcher& watcher = watchers[watcherIndex];
+	Watcher& watcher = watchers[currentProject()][watcherIndex];
 	watcher.reset();
 	if (report) {
 		if (watcher.isDisabled()) {
@@ -382,7 +389,7 @@ void resetWatcher(int watcherIndex, bool report=false) {
 
 bool isWatchingMultipleValues() {
 	int count = 0;
-	for (Watcher& watcher : watchers) {
+	for (Watcher& watcher : watchers[currentProject()]) {
 		if (!watcher.isDisabled()) {
 			++count;
 		}
@@ -397,8 +404,9 @@ void CALLBACK tick(HWND hwnd, UINT msg, UINT_PTR event, DWORD time) {
 	ostringstream s;
 	s << fixed << setprecision(1);
 	const bool multiple = isWatchingMultipleValues();
+	const ReaProject* proj = currentProject();
 	for (int w = 0; w < NUM_WATCHERS; ++w) {
-		Watcher& watcher = watchers[w];
+		Watcher& watcher = watchers[proj][w];
 		if (watcher.isDisabled()) {
 			continue;
 		}
@@ -474,7 +482,7 @@ void stop() {
 }
 
 bool isWatchingAnything() {
-	for (Watcher& watcher : watchers) {
+	for (Watcher& watcher : watchers[currentProject()]) {
 		if (!watcher.isDisabled()) {
 			return true;
 		}
@@ -698,7 +706,7 @@ class Dialog {
 void report(int watcherIndex, int channel) {
 	assert(watcherIndex < NUM_WATCHERS);
 	assert(channel < NUM_CHANNELS);
-	Watcher& watcher = watchers[watcherIndex];
+	Watcher& watcher = watchers[currentProject()][watcherIndex];
 	if (watcher.isDisabled()) {
 		// Translators: Reported when the user tries to report a Peak Watcher
 		// channel, but the Peak Watcher value is disabled.
@@ -777,7 +785,7 @@ bool processExtensionLine(const char* line, ProjectStateContext* ctx,
 		if (word != "WATCHER") {
 			continue;
 		}
-		Watcher& watcher = watchers[w];
+		Watcher& watcher = watchers[project][w];
 		watcher.reset();
 		input >> word;
 		if (word == "NONE") {
@@ -834,7 +842,7 @@ void saveExtensionConfig(ProjectStateContext* ctx, bool isUndo,
 	}
 	ctx->AddLine(CONFIG_HEADER);
 	ReaProject* project = GetCurrentProjectInLoadSave();
-	for (Watcher& watcher : watchers) {
+	for (Watcher& watcher : watchers[currentProject()]) {
 		ostringstream out;
 		out << "WATCHER";
 		if (holds_alternative<NoTarget>(watcher.target)) {
@@ -887,6 +895,7 @@ void cmdPeakWatcher(Command* command) {
 	// MIIM_TYPE is deprecated, but win32_utf8 still relies on it.
 	itemInfo.fMask = MIIM_TYPE | MIIM_ID;
 	itemInfo.fType = MFT_STRING;
+	const ReaProject* project = peakWatcher::currentProject();
 	for (int w = 0; w < peakWatcher::NUM_WATCHERS; ++w) {
 		itemInfo.wID = w + 1;
 		ostringstream s;
@@ -895,7 +904,7 @@ void cmdPeakWatcher(Command* command) {
 		// After this, information about the existing configuration for the value
 		// will be appended.
 		s << translate(peakWatcher::WATCHER_NAMES[w]) << ", ";
-		peakWatcher::watchers[w].description(s);
+		peakWatcher::watchers[project][w].description(s);
 		// Make sure this stays around until the InsertMenuItem call.
 		string str = s.str();
 		itemInfo.dwTypeData = (char*)str.c_str();
@@ -908,7 +917,7 @@ void cmdPeakWatcher(Command* command) {
 	if (w == -1) {
 		return;
 	}
-	peakWatcher::Watcher& watcher = peakWatcher::watchers[w];
+	peakWatcher::Watcher& watcher = peakWatcher::watchers[project][w];
 
 	new peakWatcher::Dialog(target, watcher, types);
 }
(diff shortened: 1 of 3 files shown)


MatejGolian:
My personal opinion is that it's sufficient to just hear reports for
the active project.

2023-01-18 16:09 GMT+01:00, AppVeyorBot ***@***.***>:

ScottChesworth:
> Is it desirable to have the possibility to still hear automatic reports from other project tabs?
> 
That runs the risk of hearing reports and not being able to figure out where they're coming from, so we'd need to think about how that was reported. Seems more intuitive to only hear reports for the active tab to me, IMO it's not worth putting time into unless someone comes forward with a really clear use case for it.

> I also left the pause functionality global. Should it be remembered per tab?
> 
If we can customize all other settings per tab now, then yeah, seems logical for pausing to behave the same. Not gonna be a showstopper though if that adds a lot of work.

MatejGolian:
To elaborate a little on my previous comment, I know that we don't
necessarily have to follow the sighted UX in every case, but as far as
peak monitoring goes, I think that sighted users only see what's going
on on the active tab at a given time and therefore I think that it
makes sense to keep this part simple.

RDMurray:
It's certainly easier to keep it the way it is. I just wanted to make sure it was the expected behaviour.  Making pause work per-tab should be easy enough. should it also be saved with the project? Btw the build in this PR also contains Jamie's work on saving peak watcher settings. 

When testing: opening, saving and switching between multiple tabs should all work. Please let me know if there is any weirdness.

ScottChesworth:
Yeah, saving pause state with the project seems logical here.

ptorpey:
Just FYI, after installing these latest builds, I’ve been getting the following error/notification from Reaper when loading a project:

 

There were 1 elements in the project that were saved by extensions. If you save this project without those extensions installed, you may lose those extensions' data.

 

Project tokens not recognized:

 

OSARA_PEAKWATCHER

 

 

--Pete

ptorpey:
As an added note, this isn’t just speech coming back but an actual dialog that pops up that one has to dismiss each time a project is loaded.  I’m guessing this is related to the updates recently made to the Peak Watcher.  Is there any way that the user can get rid of it permanently for all projects?

 

--Pete

jcsteh:
Are you running a build with the Peak Watcher changes when you see these warnings or another build (e.g. an official snapshot build)? Note that official snapshots don't have these changes yet. I don't think you should be seeing these warnings with a new Peak Watcher build, and if you are, that's a bit concerning.

jcsteh:
To put this another way, I would expect such warnings if you're running an official snapshot build, but I'm surprised if you're seeing them with a build from this pull request.

ptorpey:
I just installed:

osara_pr840-1066,072a1bf3

 

I must say, I don’t know how to interpret the version numbers, but I generally release the latest test builds from these threads.  Does the “PR” at the beginning mean public release?

 

--Pete

jcsteh:
pr840 means pull request #840, which is the pull request for nested folders. That does not include the Peak Watcher changes. Until changes are merged into the official snapshots (2023.1pre), builds from a pull request only include changes specific to that pull request. They all separately "branch off" the official snapshots, if that makes sense.

ptorpey:
Okay, I just downloaded and installed:

osara_2023.1pre-1061,c0a32720

from  GitHub, and am getting the same error dialog popping up each time I open a project.  The “2023.1pre) should be the released version, right?

 

--Pete

jcsteh:
Let's try this again. Release builds (2023.1pre) and any pr builds other than this pr (837) do not support the Peak Watcher changes. If you saved a project with a new Peak Watcher build (e.g. pr837), they will contain Peak Watcher data. If you try to open any of those projects with a build that doesn't support Peak Watcher data (2023.1pre or a pr other than 837), REAPER will complain. Once all of this is merged into official builds, none of this will be an issue unless you're using an outdated OSARA with a newer project.

src/peakWatcher.cpp
```diff
@@ -397,8 +404,9 @@ void CALLBACK tick(HWND hwnd, UINT msg, UINT_PTR event, DWORD time) {
 	ostringstream s;
 	s << fixed << setprecision(1);
 	const bool multiple = isWatchingMultipleValues();
+	const ReaProject* proj = currentProject();
 	for (int w = 0; w < NUM_WATCHERS; ++w) {
-		Watcher& watcher = watchers[w];
+		Watcher& watcher = watchers[proj][w];
```
jcsteh, line 409: This isn't a big issue since we only have 2 watchers, but this means we're doing a map lookup for project twice when we don't need to. Perhaps we could instead get the array of watchers for the project:

```
const auto& projWatchers = watchers[currentProject()];
for (int w = 0; w < NUM_WATCHERS; ++w) {
	Watcher& watcher = projWatchers[w];
```

src/peakWatcher.cpp
```diff
@@ -717,6 +730,172 @@ void report(int watcherIndex, int channel) {
...
+		if (w >= NUM_WATCHERS) {
+			continue;
+		}
+		istringstream input(data);
+		string word;
+		input >> word;
+		if (word != "WATCHER") {
+			continue;
+		}
+		Watcher& watcher = watchers[project][w];
```
jcsteh, line 793: Same as above re map lookup for the project.

src/peakWatcher.cpp
```diff
@@ -717,6 +730,172 @@ void report(int watcherIndex, int channel) {
...
+		}
+		input >> word;
+		if (word == "NOTIFY") {
+			for (auto& channel : watcher.channels) {
+				input >> word;
+				channel.notify = word == "1";
+			}
+		}
+	}
+	if (isWatchingAnything()) {
```
jcsteh, line 836: I was originally worried this was going to be a problem if the project being loaded isn't the active project. I'm not certain this can happen, but based on the fact that there is a separate GetCurrentProjectInLoadSave function, I'm guessing it can. Perhaps this happens if you load multiple projects at once; e.g. when starting REAPER with multiple tabs left open.

However, I'm now realising that this will probably be handled by the code which handles tab switching. Is that correct?

If so, we should still fix this to only assume we're watching something if GetCurrentProjectInLoadSave is actually the active project.

src/peakWatcher.cpp
```diff
@@ -717,6 +730,172 @@ void report(int watcherIndex, int channel) {
...
+		for (auto& channel : watcher.channels) {
+			out << " " << (int)channel.notify;
+		}
+		ctx->AddLine("%s", out.str().c_str());
+	}
+	ctx->AddLine(CONFIG_FOOTER);
+}
+
+void BeginLoadProjectState (bool isUndo, struct project_config_extension_t* reg){
+	//clean up configuration data for dead projects
```
jcsteh, line 875: Can we use ValidatePtr instead? That should simplify this a lot, since we could just walk through watchers erasing any associated with invalid projects.

src/peakWatcher.cpp
```diff
@@ -747,7 +927,7 @@ void cmdPeakWatcher(Command* command) {
 		// After this, information about the existing configuration for the value
 		// will be appended.
 		s << translate(peakWatcher::WATCHER_NAMES[w]) << ", ";
-		peakWatcher::watchers[w].description(s);
+		peakWatcher::watchers[project][w].description(s);
```
jcsteh, line 930: As above re multiple map lookups.

src/peakWatcher.cpp
```diff
@@ -793,13 +973,23 @@ void cmdPausePeakWatcher(Command* command) {
 	if (peakWatcher::timer) {
 		// Running.
 		peakWatcher::stop();
+		peakWatcher::isPaused=true;
 		outputMessage(translate("paused Peak Watcher"));
 	} else if (peakWatcher::isWatchingAnything()) {
 		// Paused.
 		peakWatcher::start();
+		peakWatcher::isPaused=true;
```
jcsteh, line 981: I'm guessing this should be false?

src/peakWatcher.cpp
```diff
@@ -50,6 +52,7 @@ T& varGet(V& var) {
 class Watcher;
 bool isWatchingAnything();
 void stop();
+bool isPaused{false};
```
jcsteh, line 55: I'm sure I'm missing something, but I'm not quite following why we need isPaused when we didn't before.

src/peakWatcher.h
```diff
@@ -16,3 +20,4 @@ void cmdReportPeakWatcherW2C2(Command* command);
 void cmdResetPeakWatcherW1(Command* command);
 void cmdResetPeakWatcherW2(Command* command);
 void cmdPausePeakWatcher(Command* command);
+void peakWatcherOnSwitchProjectTab();
```
jcsteh, line 23: Please move this to onSwitchTab in the peakWatcher namespace (alongside initialize). The cmd functions aren't in the namespace, but that's because no cmd function has a namespace.

jcsteh (commented):
Thanks for working on this.

ptorpey:
that makes it clear now what is going on.  I had previously opened the project when running one of the PR builds.  The fact that I’m still seeing the error with the release build is because Osara previously loaded some data about Peak Watcher with the project when running the PR build.

 

Thanks for the explanation.

 

--Pete

src/peakWatcher.cpp
RDMurray: Yes it is handled when switching tabs. I added a condition to only call start if loading the active project as you suggested.

src/peakWatcher.cpp
RDMurray: Of course. I forgot about that function.

RDMurray: I also realised the code was broken because after map::erase the iterator is invalid.

jcsteh: Yeah, that map erase behaviour is confusing. In case you aren't aware (sorry if you already are), when you erase, don't increment the iterator, but when you don't erase, do increment the iterator:
```
if (...) {
  map.erase(it);
} else {
  ++it;
}
```
It'd be nice if there were an erase_if method that takes a lambda, but that doesn't exist yet. :(

jcsteh: Oh wait, there is [std::erase_if for maps in C++ 20](https://en.cppreference.com/w/cpp/container/map/erase_if). Not sure if our Windows and Mac C++ libraries have it or not.

src/peakWatcher.cpp
RDMurray: It's purely to keep track of paused state when switching tabs. Before adding isPaused, switching from a tab with peak watcher disabled to one with peak watcher enabled would report peak watcher paused because the timer was disabled.

jcsteh: Ah. I was thinking you'd just start the timer if Peak Watcher is enabled in the tab you're switching to, but that would potentially cause you to lose the pause state since you can't tell whether it was disabled or paused in the previous tab.

src/peakWatcher.cpp
RDMurray: I didn't know about that one, even though it was mentioned further down the stackoverflow page I was reading. :)

jcsteh: Heh, I saw an answer on StackOverflow too, but it suggested erase_if didn't exist yet. I guess that StackOverflow post was before C++ 20.

src/peakWatcher.cpp
```diff
@@ -873,19 +874,10 @@ void saveExtensionConfig(ProjectStateContext* ctx, bool isUndo,
...
-		}
-		openProjects.push_back(project);
-	}
-	for (auto& [project, _]: watchers){
-		if (find(openProjects.begin(), openProjects.end(), project) == openProjects.end()){
-			watchers.erase(project);
-		}
-	}
+	erase_if(watchers, [](const auto& item) {
+auto const& [project, _] = item;
```
jcsteh, line 878: Ug. I wish we could use std::ignore in decomposition/structured bindings. I can't decide whether this or `project = item->first` looks cleaner. I'm surprised the compiler doesn't complain about an unused variable here.

jcsteh (approved):
This looks pretty good to me. I guess I'll merge the original PR and then this one if you don't have any further changes here.

I know there's a request to have pause state per tab and potentially save it. I think we can deal with those in follow-ups.

src/peakWatcher.cpp
RDMurray: This was left over from the loop, I think `project = item.first` is much better. BTW, it looks like `_` will be an official placeholder / wild card in the future. [A nice placeholder with no name](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2169r3.pdf) Compilers apparently don't complain about unused variables in structured binding because it is common to ignore one or more of them.

Lo-lo78:
Seems to work fine...
👍

PR #838, opened 2023-01-18
Added action "OSARA: Go to master track if visible"

ScottChesworth:


RDMurray:
Is there a reason for making it only work when the master is visible? It could show the master track and select it if it was invisible.

ScottChesworth:
REAPER crashed if this was used when the master wasn't visible. I've got another commit coming that clarifies the action name and protects against that. Hmm, suppose I could automatically make it visible, but that feels kinda heavy handed. Do we do that sort of thing anywhere else?

RDMurray:
Automatically making the master visible seems reasonable to me. If you don't want the master visible, why would you try to select it?

ScottChesworth:
Fair point. Will do that then. I just had eyes confirm that scrolling isn't working, definitely wanna fix that before this gets merged but not sure what I've done wrong yet.

RDMurray:
Interesting. The scrolling should be handled by the call to cmdMoveToTrack. Do you know if it works for the normal upArrow/downArrow commands?

ScottChesworth:
It does work for those yeah. Think I see what I did wrong now, I'll fix and get another check this afternoon.

ScottChesworth:
Scroll happens as soon as I DownArrow from master to the first track, which makes me think perhaps master needs to be included in the cmdMoveToTrack function, but I don't understand everything that function is doing.

RDMurray:
Most of the contents of cmdMoveToTrack are unneeded. I would just copy the lines from my earlier comment and then call `postGoToTrack(0, track);`. 

I suspect that the action "Track: Vertical scroll selected tracks into view" doesn't work for the master. Maybe it would work to select only track 1, then scroll, then select the master?

jcsteh:
Does the master automatically scroll into view if it's invisible and you make it visible? In that case, can we just hide it and show it again to force scroll?

ScottChesworth:
Master doesn't scroll into view automatically as a result of toggling its visibility.

jcsteh:
Do cmdGoToFirstTrack and cmdGoToLastTrack scroll the track into view visually? They don't explicitly do this, so I'm guessing not... but perhaps SetOnlyTrackSelected does this?

ScottChesworth:
Hmm that's a thought. Will try to rebase this and get a sight check.

jcsteh:
No need to rebase; cmdGoToFirst/LastTrack are already in OSARA snapshots, so you would just need to verify those.

ranetto:
I checked and it behaves like this:
1: when I select the tracks with the up or down arrows the track is visually selected, you have to have an appropriate zoom to see its name but it is selected.
2: When I go with the dare action to the beginning of the tracks, therefore on the first or last one the selected one is not displayed, that is, it is not even highlighted, to do this you have to move with the arrows once down and then up up to select it and highlight its name.
3: The master track is never highlighted even when it is made visible.
4: Even when opening the master track routing, the window is clearly highlighted, focused and visible, but when exiting the routing the master track is still not selected.
Maybe it's not really selectable from the program?
I hope I was helpful!
Thank you!

jcsteh:
That's super helpful, thank you.

jcsteh:
@ranetto Two more questions if I may:

1. If you have quite a few tracks and you select the last track (so that the first track scrolls off the screen), does showing the master track (View menu -> Master track) show it visually or do you have to scroll up to it?
2. You noted that the master track is never highlighted. To clarify, when you move up to it with the arrow keys, does it ever even get scrolled onto the screen? Or is track 1 always the first track you can see?

Thanks.

ranetto:
I'll check now and write to you

ranetto:
1: When I scroll the entire screen, the master track also disappears, for example if I have 80 tracks, I start from 1, I see from 1 to 15, according to the zoom set, if I scroll the first it perhaps becomes 3 and then, if I scroll again the first becomes the 15th, going down the tracks, the master track also disappears and is no longer visible with any zoom level, not even the smallest one. Even if you select all tracks with control A, the master track is never selected. When I get closer, let's say in the first 15 tracks, the master track becomes visible again, but never selected, so I would say that the screen scrolls and the master track is similar to the others, but it is fixed above all the tracks, therefore before the first place of let's say the grid, but it never moves from there. We are the ones who by selecting the tracks, therefore selecting them also means scrolling the screen to make it visible, the master track disappears as the first location is no longer visible. I also want to add that in the reaper settings, there are adjustments that allow you to decide what you have centered on the screen, now I have never looked into them in depth but if you want I will, that is, there are settings that decide what you want to have at the center of the screen screen, or in the center of the mouse position. I don't know if the theme you use affects this, but I don't think so. I add but only for experimentation and completeness that: if I select all the tracks with control a, then activate our multiselection with scift plus space, and then holding the scift I move to the master track where I press space, osara tells me selected, but It is not true. So with osara I can select and deselect the master track but it is not selected on the screen. When I move on it with space pressed, then with all the other tracks selected and with osara which with space would make me select and deselect the master track, the master itself disappears from view. But this is just for experiment, I think it means that to select all the tracks and let me see them, reaper scrolls the screen. I hope it is useful available for any other test thanks to you.

ranetto:
I must add that after activating the envelope on the master, then with our selection or by going over it it makes me select it: clearly it has a different color from the other tracks, but with scifg space osara tells me it is selected and clearly it is also selected on the video, this doesn't happen if I wanted to add it to the selection of other tracks, so if I select all the other tracks and want to add the master track to the selection it doesn't work, probably because it's of a different type. Once I have activated the envelope on the master, I can easily select and deselect it but always alone and not together with others. I don't know when I selected it what I could do with it, apparently nothing changes, the things I can do, the contexts etc, don't change, even by setting a time selection, nothing changes. I thought it was important to add this because I still checked after my previous write up.

jcsteh:
Superseded by #991.

PR #840, opened 2023-01-20
report nested folders, fixes #763

### LINKED ISSUE #763: When folders are multi-level
Hello,
I wanted to report that I have had some problems with the folders.
I often do projects that have folders and sub folders and I don't understand when I close the first level or second or third level folder.
I give an example:

traccia 1 1 open folder 
traccia 2
traccia 3
traccia quattro open folder 
traccia 5 end of folder
traccia 6
traccia 7
traccia 8 end of folder

The problem is that if the second folder contained in the main folder is updated with a new track and closes the folder, I don't know if it does so with the first or second level one.
I always have to rehearse.
Is it possible to number the levels of the folders?
For instance:

traccia 1 1 open folder 1
traccia 2
traccia 3
traccia quattro open folder 2 
traccia 5 end of folder 2
traccia 6
traccia 7
traccia 8 end of folder 1

It is very useful when you have very complex projects; if for some reason you make a mistake you risk that the routing is all wrong.
If you find a system with OSARA to distinguish folders and sub folders that would be great!

Greetings!

RDMurray:
Report when a folder is nested inside another folder. For tracks that close a folder, reports the name (and number if report track numbers is enabled) of the folder being closed. E.G.:
* 1 open folder instruments
* 2 open nested folder strings
* ...
* 10 violin end of folder 2 strings
* ...
* 20 end of folder 1 instruments

### DIFF
--- a/src/osara.h
+++ b/src/osara.h
@@ -207,6 +207,7 @@
 #define REAPERAPI_WANT_FreeHeapPtr
 #define REAPERAPI_WANT_TimeMap2_timeToQN
 #define REAPERAPI_WANT_MIDI_GetGrid
+#define REAPERAPI_WANT_GetParentTrack
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
 
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -429,6 +429,19 @@ const char* getFolderCompacting(MediaTrack* track) {
 	return ""; // Should never happen.
 }
 
+int getTrackDepth(MediaTrack* track){
+	int count = CountTracks(nullptr);
+	int depth = 0;
+	for (int i = 0; i<count; i++){
+		MediaTrack* track_ = GetTrack(nullptr, i);
+		if (track_ == track) {
+			break;
+		}
+		depth += (int)GetMediaTrackInfo_Value(track_, "I_FOLDERDEPTH");
+	}
+	return depth;
+}
+
 const char* getActionName(int command, KbdSectionInfo* section, bool skipCategory) {
 	const char* name = kbd_getTextFromCmd(command, section);
 	if (skipCategory) {
@@ -787,6 +800,11 @@ void postGoToTrack(int command, MediaTrack* track) {
 			// There's no name and track number reporting is disabled. We report the
 			// number in lieu of the name.
 			s << trackNum;
+		}
+		int level = getTrackDepth(track);
+		if (level>0) {
+			separate();
+			s << "level " << level;
 		}
 		if (armed && autoArm) {
 			separate();
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -419,7 +419,7 @@ string formatFolderState(MediaTrack* track) {
 		s << translate("end of folder");
 		// find the folder being ended by this track
 		MediaTrack* folderTrack = track;
-		for(int i = 0; i<(-state); ++i) {
+		for(int i = state; i<0; ++i) {
 			folderTrack = GetParentTrack(folderTrack); 
 		}
 		if(!folderTrack) { // shouldn't happen
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -454,19 +454,6 @@ const char* getFolderCompacting(MediaTrack* track) {
 	return ""; // Should never happen.
 }
 
-int getTrackDepth(MediaTrack* track){
-	int count = CountTracks(nullptr);
-	int depth = 0;
-	for (int i = 0; i<count; i++){
-		MediaTrack* track_ = GetTrack(nullptr, i);
-		if (track_ == track) {
-			break;
-		}
-		depth += (int)GetMediaTrackInfo_Value(track_, "I_FOLDERDEPTH");
-	}
-	return depth;
-}
-
 const char* getActionName(int command, KbdSectionInfo* section, bool skipCategory) {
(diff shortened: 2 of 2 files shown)


jcsteh:
@RDMurray , is there any pending action on this or is this ready for review? For some reason, I thought we were waiting for user feedback or something, but I don't see anything about that here.

RDMurray:
Discussion is in #763. I think this is ready for review as there weren't any objections.

Lo-lo78:
> @RDMurray , is there any pending action on this or is this ready for review? For some reason, I thought we were waiting for user feedback or something, but I don't see anything about that here.
Even though the build installer/osara_pr840-1079,8e572011.exe is not online I had the file already downloaded.
I just installed it and tried the subfolders again and it's perfect!
If others want to try the build:
[osara_pr840-1079,8e572011.zip](https://github.com/jcsteh/osara/files/10987090/osara_pr840-1079.8e572011.zip)

Greetings!

Lo-lo78:
@RDMurray 
👍

Lo-lo78:
@RDMurray 
Forse invece che:
open nested folder
Si potrebbe dire:
open sub folder.
Seems faster for screen reader
What do you think about it?

ScottChesworth:
It's only one syllable shorter, and nested is a more accurate description, so I think we should stay with the reporting style as it is now.

ScottChesworth:
@RDMurray , just took this for a spin and it all seems very excellent! Are you happy with it? Feel free to merge if so.

src/osara.h
```diff
@@ -207,6 +207,7 @@
 #define REAPERAPI_WANT_FreeHeapPtr
 #define REAPERAPI_WANT_TimeMap2_timeToQN
 #define REAPERAPI_WANT_MIDI_GetGrid
+#define REAPERAPI_WANT_GetParentTrack
```
jcsteh, line 210: This is requested but not used. However, I wonder whether we can use it after all; see below.

src/reaper_osara.cpp
```diff
@@ -402,16 +402,41 @@ string formatCursorPosition(TimeFormat format, FormatTimeCacheRequest cache) {
 	return formatTime(GetCursorPosition(), format, false, cache);
 }
 
-const char* formatFolderState(int state, bool reportTrack=true) {
+const string formatFolderState(MediaTrack* track) {
```
jcsteh, line 405: There's no benefit in this being const now. Previously, it was const because this returned a constant defined at compile time, so we didn't want it to get modified. Now, the string is always generated at runtime, so it must always be a different string and thus there's no reason to prevent the caller from modifying it.

src/reaper_osara.cpp
```diff
@@ -402,16 +402,41 @@ string formatCursorPosition(TimeFormat format, FormatTimeCacheRequest cache) {
...
+		s << translate("folder");
+	} else {
+		// Translators: A track which ends its folder.
+		s << translate("end of folder");
+		int trackNum = (int)GetMediaTrackInfo_Value(track, "IP_TRACKNUMBER");
+		// find the folder being ended by this track
+		MediaTrack* folderTrack;
+		int folderTrackNum{0};
+		for(folderTrackNum = trackNum -1; folderTrackNum>0; folderTrackNum--){
+			folderTrack= GetTrack(nullptr, folderTrackNum-1); // getTrack is 0 based
```
jcsteh, line 425: Is there a reason we don't use GetParentTrack here? As it is now, we have to walk backwards through all the intervening tracks. If we use GetParentTrack, we can get the parent's track number after we find it. I realise we still need the folder depth, but that's fine; we can get that after calling GetParentTrack.

src/reaper_osara.cpp
```diff
@@ -770,14 +795,12 @@ void postGoToTrack(int command, MediaTrack* track) {
 	if (trackNum > 0) { // Not master
 		int folderDepth = *(int*)GetSetMediaTrackInfo(track, "I_FOLDERDEPTH",
 			nullptr);
+		const string folderState = formatFolderState(track);
```
jcsteh, line 798: nit: I think we can just call this in the two places we need it, rather than getting it here. This avoids pointless string generation for normal tracks.

jcsteh (changes requested):
Thanks.

It'd be great if you could update the pull request description with a summary of the UX similar to what's documented in the comments of #763.

src/reaper_osara.cpp
RDMurray: There wasn't any good reason. I changed it to use GetParentTrack.

src/reaper_osara.cpp
```diff
@@ -402,16 +402,38 @@ string formatCursorPosition(TimeFormat format, FormatTimeCacheRequest cache) {
...
 	} else if (state == 1) {
 		// Translators: A track which is a folder.
-		return translate("folder");
+		s << translate("folder");
+	} else {
+		// Translators: A track which ends its folder.
+		s << translate("end of folder");
+		// find the folder being ended by this track
+		MediaTrack* folderTrack = track;
+		for(int i = 0; i<(-state); ++i) {
```
jcsteh, line 422: Would this be more readable as something like:

`for (int i = state; i < 0; ++i) {`

jcsteh (approved):
Thanks. One more nit/question, but if you disagree, I'm happy to leave it; I don't feel strongly about it. Feel free to merge once addressed/answered.

src/reaper_osara.cpp
RDMurray: It probably is more readable. That's actually how I wrote it the first time and then I changed it to count from 0.

PR #845, opened 2023-01-24
Updated copyright notices throughout source code and contributors in …

ScottChesworth:
…ReadMe

### DIFF
--- a/installer/osara.nsi
+++ b/installer/osara.nsi
@@ -1,7 +1,7 @@
 ; OSARA: Open Source Accessibility for the REAPER Application
 ; NSIS installer script
 ; Author: James Teh <jamie@jantrid.net>
-; Copyright 2016-2021 NV Access Limited, James Teh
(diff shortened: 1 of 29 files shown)


jcsteh:
FWIW, I tend to update the copyright year on a source file only when that specific source file is changed. For example, controlSurface.cpp hasn't been touched in 2023, so its copyright hasn't changed in 2023 either. Of course, I sometimes forget to update it even when the file does change, which isn't ideal. :)

That said, I don't have a problem with this being updated either.

ptorpey:
With our eyes On success web site, I use include files to show the copyright.  Don’t know if that is valid for code.  Sure makes it more simple to update!

 

--Pete

jcsteh:
For an open source project, include files for copyright headers aren't a great idea because if the file were somehow separated from the project, legally we still want it to be clear that it is under copyright and what license is attached to it. It's not so important for a website because it's generally not open source, so people aren't seeing the code, just the output of said code.

PR #854, opened 2023-01-31
Add OSARA: Cycle shape of selected envelope points action.

### LINKED ISSUE #853: Changing envelope shape with a keystroke
Hi.  I would like to be able to change the shape of the currently selected envelope points with a keystroke rather than having to open up the envelope point dialogue box, tab to the shape radio buttons and change it there.
So I would use alt J and alt K to navigate to my envelope point.  I could at this time also use shift alt J or K to select other adjacent envelope points, or in fact control A to select them all, and then just press a keystroke to cycle through the options of linear, square, slow start slow finish, slow start, fast staart.  Ideally a keystroke to go backwards maybe through this list?  Don't think it should particularly give Bezier unless there is some wizzy way to change the tension without going into the dialogue box.
As for which keystroke.  Could we use a keystroke that might already be used for something but which isn't meaningful when the focus is on an envelope point.  Global automation mode perhaps control shift alt forward slash.  Open to suggestion here.

jcsteh:
Feedback welcome.

Fixes #853.


Lo-lo78:
@jcsteh 
It works very well.
Let's hear what others think.
👍

jcsteh:
@ScottChesworth , are you happy for this to be merged or do you want to play with it a bit more?

Lo-lo78:
Hi for me it works very well this way too.
I don't know if it's useful.
It's a thought that came to me.
For new users, could we put a short warning when no envelope point is selected?
For instance:
"no envelope point selected"

ScottChesworth:
That's easy to do, but it's different from our convension where usually, if nothing happens, nothing is spoken. Is there a reason to treat this action differently?

ScottChesworth:
> @ScottChesworth , are you happy for this to be merged or do you want to play with it a bit more?
> 
It's working a treat here. Just gonna check whether Lolo is on to something, I'll merge if nothing extra is needed. Thanks!

jcsteh:
Yeah, I did think about this. We've made an exception to the "no action, no report" rule in a few cases; e.g. track navigation when there are no tracks, deleting when there is nothing to delete, etc. We've never done this for envelope points though. I don't really have an opinion either way here.

ScottChesworth:
I do, and annoyingly, it's an opinion that's evolved since I requested some of those exceptions early on. I think "no action, no report" is a good pattern to follow, because with the amount we report nowadays, the screen reader not speaking instantly tells you that something about the workflow you're trying to execute has gone wrong.

Lo-lo78:
> 
> It's working a treat here. Just gonna check whether Lolo is on to something, I'll merge if nothing extra is needed. Thanks!

Yes, it's ok for me!

CastleStarter:
This works really nicely.  I also selected multiple points and the first point drives what all the points are set to which makes good sense.  Many thanks.

Issue #383, opened 2020-11-18
Osara is responding slow and the speech lags when using the report soloed tracks action


GianlucaApollaro:
it takes much longer to report how many tracks are soloed if there are many virtual instruments in use. It happened to me when I had more than 10 Kontakt instantiated. unarming tracks or bipassing fx makes no difference

ScottChesworth:
Does this clear up if you disable "Report changes made via control surfaces" in OSARA Config?

jcsteh:
It'd be helpful (and this applies to anyone filing issues) if you could provide exact steps to reproduce the issue. For example, when you say "report how many tracks are soloed", are you talking about track navigation (which seems to be how Scott interpreted it) or are you talking about the report soloed tracks command? Neither of these reports "how many tracks are soloed", hence the confusion.

All of that aside, my guess (and it's very much a guess given the lack of info) is that this is caused by the solo defeat reporting. The REAPER API doesn't properly support querying solo defeat, so OSARA has to inspect the raw track chunk data for that information. I guess if the track data is huge for every track, that might slow things down.

[Here is a build from just before the solo defeat support](https://ci.appveyor.com/api/buildjobs/x7qw7mwo7baqcuu6/artifacts/installer%2Fosara_2020.1pre-305%2C8e2a70c6.exe). Try that and see if it makes a difference. If it does, we'll probably need to revert the solo defeat code and ask Cockos for something better.

GianlucaApollaro:
Hi,
I'm talking about the report soloed tracks command
And It happens mostly when there are many vstIs loaded, to me it happened with Kontakt,  and the tracks had no midi items, just an instrument per Kontakt instance loaded. I had a vst on every track

GianlucaApollaro:
disabling "Report changes made via control surfaces" in OSARA Config made no difference.

GianlucaApollaro:
Hi,
I performed some other tests with the latest build and the speech lags mostly when there are many instances of Kontakt or Komplete Kontrol loaded. Doesn't matter which screen reader is used. I tested with jaws and Nvda and both of them lag. Once with jaws reaper became not responding for about 3 seconds and then it said the soloed tracks. There is no difference how many track you solo, even when there are none you hear "Soloed none" after 5 or more seconds since when the key is pressed.

jcsteh:
And did you test with the earlier build I linked above several days ago?

GianlucaApollaro:
Yes, tested and no lag at all

[...]

PR #855, opened 2023-01-31
Add feedback for horizontal zooming in the MIDI editor

RDMurray:
- View: Zoom in horizontally: NumPad +, Shift+=
- View: Zoom out horizontally: NumPad -, Shift+-

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -50,6 +50,29 @@ int getItemPPQ(MediaItem* item) {
 	return ppq;
 }
 
+// return the midi editor zoom ratio of the item 
+double getMidiZoomRatio(MediaItem* item) {
+	static const regex re("CFGEDITVIEW -?[0-9.]+ ([0-9.]+) ");
+	const size_t startBuffSize = 16384;
+	const size_t maxBuffSize = 1<<24; // 16 Mib
+	for (size_t buffSize = startBuffSize; buffSize <= maxBuffSize; buffSize *= 2) {
+		string buffer(buffSize, '\0');
+		if (!GetItemStateChunk(item, &buffer.front(), buffSize, false)) {
+			return -1;
+		}
+		smatch match;
+		if (!regex_search(buffer, match, re)) {
+			if (buffer[buffSize-2] != '\0') { //chunk bigger than buffer
+				continue;
+			} else {
+				return -1;
+			}
+		}
+		return stod(match.str(1));
+	}
+	return -1;
+}
+
 // Note: while the below struct is called MidiControlChange in line with naming in Reaper,
 // It is also used for other MIDI messages.
 struct MidiControlChange {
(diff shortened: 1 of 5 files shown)


Lo-lo78:
@RDMurray 
It works well.
Now we have a problem.
In the piano roll we have midi ticks but if the zoom works in pixels?
Is it possible to convert from pixels to ticks?

ScottChesworth:
I don't think it's OSARA's job to do that conversion. Zoom is in pixels everywhere in REAPER.

src/midiEditorCommands.cpp
```diff
@@ -50,6 +50,29 @@ int getItemPPQ(MediaItem* item) {
...
+	static const regex re("CFGEDITVIEW -?[0-9.]+ ([0-9.]+) ");
+	const size_t startBuffSize = 16384;
+	const size_t maxBuffSize = 1<<24; // 16 Mib
+	for (size_t buffSize = startBuffSize; buffSize <= maxBuffSize; buffSize *= 2) {
+		string buffer(buffSize, '\0');
+		if (!GetItemStateChunk(item, &buffer.front(), buffSize, false)) {
+			return -1;
+		}
+		smatch match;
+		if (!regex_search(buffer, match, re)) {
```
jcsteh, line 64: This is super clever. That said, I worry about performance here. We had to disable solo defeat reporting back when it used a regexp to search a state chunk; see #383. Of course, this isn't as bad because it will only impact the user if they actually need to have the zoom level reported. On the other hand, the user is more likely to press it many times in rapid succession and I worry this will cause major lag on large MIDI items if the user pressed it many times rapidly.

I guess we'll just have to see how it goes in practice. I agree having this is better than nothing, but I really wish Cockos would just expose this.

src/midiEditorCommands.cpp
```diff
@@ -50,6 +50,29 @@ int getItemPPQ(MediaItem* item) {
...
+			if (buffer[buffSize-2] != '\0') { //chunk bigger than buffer
```
jcsteh, line 65: From this, I gather that the CFGEDITVIEW part is towards the end of the chunk (after the take data), rather than being at the start? Or is this out of an abundance of caution? I'm just double checking in case we can go for a much smaller buffer and avoid retries.

RDMurray: Unfortunately it is after all the midi events, which are in text format so they take up a lot of space.

jcsteh (approved):
Thanks. Let's ship it and see how it flies. :)

jcsteh:
Oops, too fast on the trigger. Let's deal with that in a follow-up I guess.

src/midiEditorCommands.cpp
ScottChesworth: > I agree having this is better than nothing, but I really wish Cockos would just expose this.
> 
Shall I try Cockos again? I'm thinking perhaps Justin might juggle it up the to-do list if I show him that Robbie's done some wizardry but the best way we've got available is still gonna hit a wall performance-wise.

jcsteh: Let's see if users actually complain. If they don't, no point in ringing the bell again. If they do, well, it's more likely to get fixed if we can show it's hurting users.

That said, this take problem Robbie has discovered might be a show-stopper and require us to push on it anyway.

PR #857, opened 2023-02-01
Fix reporting of zoom in the MIDI editor when there is more than one take

RDMurray:


ScottChesworth:
All seems good on the front-end man. Is this ready to be merged?

RDMurray:
Not quite, @jcsteh should probably take a look at this first. I just committed a change that drastically improves performance for large midi items.

Feel free to test this with huge midi items that have multiple takes and over 100000 events. See if you can make it crash. :)

ScottChesworth:
Just done a test where I had 7 takes, each one being a randomly picked bunch of MIDI files glued together. The lightest take had just over 100000 events, the biggest had 277389. Changing the zoom worked fine. It was noticeably more responsive than navigating between notes lol.

Lo-lo78:
> Not quite, @jcsteh should probably take a look at this first. I just committed a change that drastically improves performance for large midi items.
> 
> Feel free to test this with huge midi items that have multiple takes and over 100000 events. See if you can make it crash. :)

I tried with items of 200000 midi events and the problem is that it slows everything down.
I think it depends on NVDA.
If NVDA receives too many events it also slows down Reaper.
I think nothing can be done.
It's the same problem with the Surge XT.
Juce is accessible but Surge has too many events that it sends to NVDA and is therefore a little slow to open the plugin.
My guess is that NVDA has some issues with very large lists and with applications/plugins that have a lot of events it also slows down the application/plugin.
I hope I'm wrong!

src/midiEditorCommands.cpp
```diff
@@ -50,27 +51,33 @@ int getItemPPQ(MediaItem* item) {
 	return ppq;
 }
 
-// return the midi editor zoom ratio of the item 
-double getMidiZoomRatio(MediaItem* item) {
+struct FreeReaperPtr {
+	void operator()(void* p)
+	{
```
jcsteh, line 56:
```suggestion
	void operator()(void* p) {
```

jcsteh (approved):
Ah, this is much nicer. Thank you.

PR #858, opened 2023-02-02
Pan galactic gargle bastard

ScottChesworth:


jcsteh:
I have no idea what's going on at your end, but I just tried your code with no changes, bound control+alt+p to the new action and everything works as expected. So... well done, I guess, except it's not working for you? Curious.

ScottChesworth:
Hmm, clearly I've done something daft. Investigating...

ScottChesworth:
Rebuilt and remapped and it does indeed work. I don't know what I did wrong, but will get the rest of it done today.

ScottChesworth:
@jcsteh, probably gonna need you to resolve those conflicts man, that looks complicated. Mac key map still needs updating, if you can get it to the point where there's an AppVeyor build, @jennykbrennan will do the last part.

jcsteh:
I haven't looked at the code yet, but I see you made a commit about moving track/take earlier in the report. I deliberately put it later based on the principle of the most important info being placed first for efficiency. That is, if the user knows what they're doing, they'll already know what context they're in, so they can hear the envelope name first and ignore or silence the rest. If they're not sure, they can keep reading for the context. I can live with either, but just wanted to flag that this was an intentional decision.

jcsteh:
That said, if the user knows what they're doing, I guess they'll know what action they triggered. Hmm. But keep in mind these same reports are used when toggling last touched FX envelope, for example.

ScottChesworth:
I moved track/take earlier because while the keystrokes are familiar, the context sensitivity of them is new. Figured power users might appreciate having the reminder as early as we can say it without totally breaking form, essentially I'm trying to avoid them feeling like they're being slowed down too much while muscle memory is being reprogrammed.
So we've got two intentional decisions, and either of us could live with the other. What happens now? Do we wrestle? :)

jcsteh:
Wrestling sounds fun. I wonder who would win? :)

In this virtual instance, I will concede to your superior wisdom, especially given you'll be the one having to deal with the users. Muahaha.

ScottChesworth:
I'd tag in @mattymc2 and you'd be in trouble, dude's so greasy it's basically impossible to get him in a choke hold.

jcsteh:
Yeah, but he's Australian, so he'd be on my side. And so would the koalas. They're vicious little bastards you know.

ScottChesworth:
I think you're underestimating the strength of this bond across the pond. Need I remind you we've made a moderately unsuccessful podcast together? It changes people...

jcsteh:
Wouldn't that make it a pondcast?

ScottChesworth:
Are you calling my good friend @mattymc2 pondlife?

ScottChesworth:
I mean, not saying I disagree...

jcsteh:
What ever happened to that podcast anyway? I was just starting to enjoy it and then you disappointed me by not producing further episodes. I mean, it's not like you don't have 5 billion other things to do. Entertain me damn it!

jcsteh:
Also, props for the reference to pan galactic gargle blaster.

ScottChesworth:
Hehe, thought you might like that reference.
Re the podcast, we almost managed a reboot at the end of last year, almost. Could potentially still happen.

PR #859, opened 2023-02-04
Added new context sensitive actions to Mac key map

ScottChesworth:


ScottChesworth:
I think this is the Mac key map. I say think because all that conflict stuff got a bit hairy here for a minute.

jcsteh:
I manually merged this.

Issue #861, opened 2023-02-06
can we do some light reshaping of how controls in the Peak Watcher GUI are named to shore up UX across screen readers?


ScottChesworth:
Navigating the "Hold level" grouping of Peak Watcher's GUI reports differently with NVDA, JAWS and VoiceOver. While absolute consistency probably isn't possible, I'm thinking perhaps there's a way to clarify the UX with some light reshaping of language and one role tweak. Jotting down notes here best I can with a limited technical understanding. If someone can point me at where these controls are named, I'd be happy to sit with the screen readers and twiddle until I get some sort of improvement.

These are the current issues I'm aware of:
1. The edit field that becomes navigable once our radiobutton in this group is set to "for (ms)" doesn't have an accName yet. This causes NVDA to just report "edit" (not a huge deal because what's expected there is only one control away), but with VO navigation that's less clear IMO, and JAWS decides to report "When level reaches" there which is just... er... wrong. Would adding an accName of "Hold for ms" solve this?
2. I think forward and backward navigation through the tab order would be clearer with these language tweaks:
- "Hold level" becomes "Hold level until:"
- "until reset" becomes "the user resets this watcher"
- "for (ms):" becomes "a given amount of time passes"
3. Could we also add an accelerator to the edit field where ms is specified? H seems logical if we go with the accName suggestion above.

Obviously these suggestions make reports longer, but IMO the increased clarity for newcomers is a worthy trade. Power users are familiar enough with this GUI by now that I'm pretty sure it won't slow them down.

Thoughts?

jcsteh:
1. The reason the ms field is unlabelled is that it's effectively labelled by the radio button. It's also inside the Hold for: grouping. I can label it, but that also means we'll have a redundant label visually, since I can't assign non-visual labels in a cross-platform way. I guess that's better than the alternative though.
2. I originally considered labelling the group "Hold level until:". However, this means that for the Disabled option, you get "Hold level until disabled", suggesting that the level is held until the watcher is disabled, which is precisely the opposite of what it does.

jcsteh:
Btw, I'm curious as to why the unlabelled field is less obvious with VO navigation, given that it's right next to the ms radio button?

ScottChesworth:
> I'm curious as to why the unlabelled field is less obvious with VO navigation, given that it's right next to the ms radio button?
> 
It's not. Here's a chronological list of how VO reports controls in the dialog with settings at their default states, starting from where focus lands then moving with VO+RightArrow:
[...]

PR #862, opened 2023-02-08
Commented out call to translateDialog in peakWatcher.cpp 

ScottChesworth:
for GUI comparison, as per Jamie's suggestion in #861


ScottChesworth:
@jennykbrennan, when you surface from your pit and have been adequately doused in coffee, can you compare Peak Watcher using this AppVeyor build? Try VO navigation and the tab order, let us know whether there are any differences. The reports from both navigation approaches with an unmodified snapshot are here for comparison: https://github.com/jcsteh/osara/issues/861#issuecomment-1421303467_

jennykbrennan:
Will do.

jennykbrennan:
Did do. No changes so far as I can tell.

ScottChesworth:
Thanks for checking. Commencing Justin brain probe...

PR #863, opened 2023-02-08
Don't pass page keys to Transient Sensitivity dialog, 

ScottChesworth:
as they don't serve any useful purpose within the dialog, and our usual mappings can still be used if these are allowed to reach main section.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3854,8 +3854,6 @@ int transDetect_translateAccel(MSG* msg, accelerator_register_t* accelReg) {
 		case VK_LEFT:
 		case VK_UP:
 		case VK_DOWN:
-		case VK_PRIOR:
-		case VK_NEXT:
 		case VK_HOME:
 		case VK_END:
 		case VK_SPACE:


jcsteh:
They don't do large steps on the sliders?

ScottChesworth:
Freed these up because it was irking me that I kept reaching for the usual keystrokes out of habit. Leaving for @jcsteh to merge though, just in case he had a reason for redirecting the page keys.

ScottChesworth:
> They don't do large steps on the sliders?
> 
Nope they don't. Another SWELL thing I keep meaning to ask Justin about.

ScottChesworth:
Come to think of it though, actually this is a SWELL thing I've already asked Justin about. Twice if memory serves. It hasn't appeared yet.

Lo-lo78:
With NVDA I was able to read the elements to split with inline exploration within the object.

I wanted to point out that in the Dynamic split items window when we press the button:
The "Set transient sensitivity" new window is no longer accessible with the tab or shift tab key.Quando ci accediamo direttamente con l'azione:
"Transient detection sensitivity/threshold: Adjust...".
Strangely it works but it's the same window.

jcsteh:
So... if that gets fixed in SWELL, would we want to revert this change? Will we still want some version of this change at that point? I'm not quite sure where to go here.

ScottChesworth:
> in the Dynamic split items window when we press the button: The "Set transient sensitivity" new window is no longer accessible with the tab or shift tab key.
> 
It never was. The hack that makes it possible to navigate with tab relies on detecting that the dialog is open. We can detect that when loading the dialog via the action on Control+Alt+Shift+T, but we can't detect the change when it loads within the dynamic split dialog.

ScottChesworth:
> I'm not quite sure where to go here.
> 
Think I sent a PR in a state of excitement just because something worked the way I thought it would on the first try LOL. Leave it open for a bit, I'll revisit and either merge or close once I've had a chat with Justin about SWELL improvements.

jcsteh:
Cockos will probably know what you're talking about regardless, but just a clarification: this is more WDL wingui than swell. Swell is the Mac emulation layer, whereas wingui implements these custom widgets used on both Windows and Mac.

ScottChesworth:
Gotcha, thanks. Every extra bit of seeming like I know what I'm talking about helps here. :)

jcsteh:
We eventually want to get pageUp/pageDown moving by step on these sliders, so we don't think it makes sense to merge this now and then unmerge it later. Abandoning.

PR #864, opened 2023-02-08
Fixed truncated text for Follow check box and Loudness Range combo bo…

ScottChesworth:
…x entries


ScottChesworth:
Got a set of eyes to give the GUIs a once-over on Win and Mac before reporting the control ordering issue we're having to Cockos, figured I'd see whether there are any visual differences. There aren't, but it turned out our check box label was only showing the first couple of characters and the U of Loudness Range LU in the combo box was truncated. This fixes both issues by adjusting the width of those controls.

PR #873, opened 2023-02-25
Add feedback for View: Zoom in/out vertical, Toggle track zoom to minimum/maximum height.

### LINKED ISSUE #872: Can we report vertical zoom?
A request for this to be reportable came in from @dgl1984 via the WhatsApp group. It would be useful when tutoring/presenting REAPER to an audience with usable vision, making screen caps, creating YouTube content etc. Likely also helpful when working alongside sighted peers.

I took a stab at it, but this API has more arguments than I'm used to, scons was like "ROFL, are you typing with your toes again"?
Here's some notes I was working from.

Reckon this is the relevant API:
void adjustZoom(double amt, int forceset, bool doupd, int centermode)
forceset=0,doupd=true,centermode=-1 for default

These actions seem like they'll be useful:
40111 View: Zoom in vertical
40112 View: Zoom out vertical
_SWS_VZOOMFIT SWS: Vertical zoom to selected tracks
_WOL_SETVZOOMC_TRACKCVIEW SWS/wol: Options - Set "Vertical zoom center" to "Track at center of view"
_WOL_SETVZOOMC_TOPVISTRACK SWS/wol: Options - Set "Vertical zoom center" to "Top visible track"
_WOL_SETVZOOMC_LASTSELTRACK SWS/wol: Options - Set "Vertical zoom center" to "Last selected track"

No idea what the API is gonna return, suspect some feedback is better than none though.

Thanks in advance to anyone who can take a run at this. I'm happy to do tedious stuff here if that helps at all.

jcsteh:
Fixes #872.

Can we get folks to try this out and see if it's useful?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -229,6 +229,10 @@ Most of these are actions built into REAPER, but a few are useful actions from t
 #### Zoom
 - View: Zoom out horizontal: - or NumPad-
 - View: Zoom in horizontal: = or NumPad+
+- View: Zoom out vertical
+- View: Zoom in vertical
+- View: Toggle track zoom to minimum height
+- View: Toggle track zoom to maximum height
 
 #### Options
 - Options: Cycle ripple editing mode: Alt+Shift+P
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1873,6 +1873,33 @@ void postToggleTakePreservePitch(int command) {
 		translate("disabled preserve pitch when changing item rate"));
 }
 
+void postChangeVerticalZoom(int command) {
+	int size = 0;
+	int index = projectconfig_var_getoffs("vzoom2", &size);
+	assert(size == sizeof(int));
+	int zoom = *(int*)projectconfig_var_addr(nullptr, index);
+	switch (zoom) {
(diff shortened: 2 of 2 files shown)


jcsteh:
@ScottChesworth , thoughts? Are we abandoning this or do we want it?

ScottChesworth:
@dgl1984, have you tried this and is it useful for your presentations?
My take is that even if some investigation is still required on the user's part to understand what the feedback means in context, it's better to have some feedback than not, so I think we should keep it.

dgl1984:
It's proven to be awesome actually.

It's helped me know what people can actually see.

It turns out that we can manipulate tracks and items when no one can see 
it happening.  The changes are obvious when we zoom out, but if we have 
one track displayed, but are looking at another in peak watcher, we can 
tell cited people what the levels are on things we can't see.

Same with automation, or any changes made on a control surface.

Although every system I've used gives different visual results with 
similar numbers, it doesn't take much time to get someone to take a look 
and tell you what the numbers mean on your desktop/laptop.

ScottChesworth:
Thanks for confirming. Gonna merge this then.

PR #877, opened 2023-03-10
Annotate the FX chain dialog and property pages so NVDA doesn't try to guess the caption.

### LINKED ISSUE #843: Tidy up extraneous speech when navigating tab order in FX window;
I'm not sure whether this is one best handled by OSARA or Cockos (suspect the latter), but there are examples of extraneous speech when navigating the tab order in the FX window that I'd love to get cleaned up. I regularly see new users being bamboozled by this when they're not yet familiar enough with the layout to fly around on muscle memory, some tightening here would do a lot of good for the UX.

1. Create a track and instantiate ReaEQ (note that these examples are specific to the FX window, not ReaEQ).
2. From the FX chain list (the default landing position), press tab three times to reach the Presets combo box. Do this slowly and notice how the first two presses provide clean reports, but on the third press when you reach the Presets combo box, the useful part of the report (IE, that the Presets combo box is now focused and which preset is selected) is prefixed with a bunch of status stuff from elsewhere in the window. Here's an example report:
"0.0%/0.0% CPU 0/0 spls  property page  A2 Gain: ReaEQ," and only then do we hear "combo box  No preset  collapsed".
3. Now hit Tab once more, and start Shift+Tabbing back through the same controls. Moving in this direction, reporting is clean as a whistle. Ideally, this is the UX I'd like us to have travelling forward.
4. Another example, tab until you reach the unlabelled check box (already asking Cockos to fix that), and notice on the next press, we have another extraneous prefix:
"property page  G2," then we hear the actual control, "Band frequency  slider  100.0".

The fact that these examples are split into two separate reports in NVDA's speech buffer leads me to believe it happens when we move between groups in the object hierarchy or some such thing, but I don't really know enough yet to be sure I'm asking Cockos for the right stuff. Do we need improvements to the roles in the FX window? If we're going to hear prefixes, it would be good for them to be clarified, like "Presets grouping", "FX controls" and so on.

Any advice welcome. I'm happy to do the asking, watch the dev builds like a hawk and provide feedback once this gets moving, I just wanna be sure that what I'm requesting is correct and concise.

jcsteh:
This caption guessing can make it very slow to focus effects with a lot of parameters such as Surge.
Fixes #843.

### DIFF
--- a/src/fxChain.cpp
+++ b/src/fxChain.cpp
@@ -62,12 +62,6 @@ bool maybeSwitchToFxPluginWindow() {
 	if (!(window = FindWindowExA(window, nullptr, "#32770", nullptr))) {
 		return false;
 	}
-	// This is a property page containing the plugin window among other things.
-	// set property page name, to avoid CPU/PDC label audition after switching
-	char name[8];
-	if (GetWindowText(window, name, sizeof(name)) == 0) {
-		SetWindowText(window, " ");
-	}
 	// Descend. Observed as the first or as the last. 
 	// Can not just search, we do not know the class nor name.
 	if (!(window = GetWindow(window, GW_CHILD)))
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4519,6 +4519,40 @@ void annotateAccRole(HWND hwnd, long role) {
 		var);
 }
 
+// Several windows in REAPER report as dialogs/property pages, but they aren't really.
+// This includes the main window.
+// Annotate these to prevent screen readers from potentially reading a spurious caption.
+void annotateSpuriousDialog(HWND hwnd) {
+	annotateAccRole(hwnd,
+		hwnd == mainHwnd || hwnd == GetForegroundWindow() ? ROLE_SYSTEM_CLIENT :
+		ROLE_SYSTEM_GROUPING);
+	// If the previous hwnd is static text, oleacc will use this as the name.
+	// This is never correct for these windows, so override it.
+	if (GetWindowTextLength(hwnd) == 0) {
+		accPropServices->SetHwndPropStr(hwnd, OBJID_CLIENT, CHILDID_SELF, PROPID_ACC_NAME, L"");
+	}
+}
+
+void annotateSpuriousDialogs(HWND hwnd) {
+	annotateSpuriousDialog(hwnd);
+	for (HWND child = FindWindowExW(hwnd, NULL, L"#32770", NULL); child;
+		child = FindWindowExW(hwnd, child, L"#32770", NULL)
+	)
+		annotateSpuriousDialogs(child);
+}
+
+UINT_PTR annotateFxDialogTimer = 0;
+void CALLBACK annotateFxDialog(HWND hwnd, UINT msg, UINT_PTR event,
+	DWORD time
+) {
+	KillTimer(nullptr, annotateFxDialogTimer);
+	annotateFxDialogTimer = 0;
+	if (!GetFocusedFX(nullptr, nullptr, nullptr)) {
+		return;
+	}
+	annotateSpuriousDialogs(GetForegroundWindow());
+}
+
 HWND prevForegroundHwnd = nullptr;
 DWORD prevForegroundTime = 0;
 HWND prevPrevForegroundHwnd = nullptr;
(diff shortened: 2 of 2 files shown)


ScottChesworth:
Looking good here, Thanks for the hackage! Does this need stress testing before merging? If so, what sort of stuff could break it?

jcsteh:
There are undoubtedly race conditions where the dialog or the property page could get focus before the OSARA timer runs and cleans them up. However, even if that happens, I don't think there's anything I can do about it. REAPER seems to take a while to update the API when an effect gets focus.

I'd like to get some more people testing to make sure it addresses the problem across various systems before merging.

ScottChesworth:
Gotcha. I'll keep poking around busy GUIs here then.
Paging @castlestarter, @pitermach and @lo-lo78. Can you folks hammer on this in some taxing projects and report back?

Lo-lo78:
@jcsteh 
@ScottChesworth 
@vick08 
@leonardder 
Perfect!
Works very well with Surge.
Exiting the context menu is also fast when using the Surge XT.
I am very happy because it helps a lot to lighten the CPU.
In very large projects upon opening Surge the computer suffered.
With this build everything is much smoother.
We hear what others say.
👍👍👍

Lo-lo78:
There's just one small problem.
When you press esc or select a parameter from the context menu, and exit the screen reader does not read the focus.
The object navigator doesn't work but if I tab or shift+tab now it's much faster and this is already a great result.
I think it doesn't depend on the screen reader because even Narrator doesn't read anything when you exit the context menu.
However the improvement is excellent!

CastleStarter:
Hi, had to go read the thread to see what this was about.  But I downloaded latest OSARA and in Surge the control to select the filter type now seems more responsive.  Significantly more responsive.

Is that the right sort of test you were after?  I can spend more time this evening taking it for a drive.

Best, Castle

pitermach:
I haven’t yet had time to test this change but I want to say the issue with having to tab after closing a menu is a juce bug not related to reaper or OSARA. Anyway, going by what the others said this change has indeed greatly improved performance so thanks for the fix!Sent from my iPhoneOn 10 Mar 2023, at 09:02, Lo-lo78 ***@***.***> wrote:﻿
There's just one small problem.
When you press esc or select a parameter from the context menu, and exit the screen reader does not read the focus.
The object navigator doesn't work but if I tab or shift+tab now it's much faster and this is already a great result.
I think it doesn't depend on the screen reader because even Narrator doesn't read anything when you exit the context menu.
However the improvement is excellent!

ScottChesworth:
> the issue with having to tab after closing a menu is a juce bug...
> 
Think I've already raised that with them a while back, but will check and/or chase.

jcsteh:
Now I'm deeply confused. People were saying that exiting a context menu would lag when returning to the UI. But now people are saying that exiting a context menu doesn't return to the UI at all. Huh?

pitermach:
It did lag, just that juce doesn’t fire the event when the menu closes so nvda realizes it after the focus is moved with tab and then you get the lag. More contexthttps://github.com/surge-synthesizer/surge/issues/6426Sent from my iPhoneOn 10 Mar 2023, at 12:39, James Teh ***@***.***> wrote:﻿
Now I'm deeply confused. People were saying that exiting a context menu would lag when returning to the UI. But now people are saying that exiting a context menu doesn't return to the UI at all. Huh?

ultrasound1372:
Should this be tested with the UI in generic mode? I also think there's at least one effect which actually uses a property page in its exposure, though I can't for the life of me remember its name now.

ScottChesworth:
So far as I know the generic UI doesn't display anything beyond what's available in OSARA's Parameters dialog, so IMO improving the UX with that UI active isn't a priority, we've already got a robust and more navigable alternative.

ultrasound1372:
Lag is still present in bridged mode, when changing filter type or when loading a new patch I get a few seconds of nothing. I've also applied this property page hack to NVDA and it doesn't seem to change anything, so that might be a separate issue. I use it in bridged mode because alt+comma/period to navigate the Surrge UI don't work in the chain.

ScottChesworth:
Last I checked, Alt+Comma/Period worked in Surge XT's GUI everywhere other than when the patch browser has focus. Is that not the case for you?

ultrasound1372:
Surge doesn't receive it, it goes to Reaper. Says no items selected.

ironcross32:
Noticeably improved.

ironcross32:
@ultrasound1372 What you can do is before you load Surge, app key on it from the add FX dialog, then set send all keyboard input to plugin from the default settings for new instances submenu. You will no longer have this issue as long as you ensure Surge's UI is in focus and as @ScottChesworth has pointed out, not in the patch section.

ironcross32:
@ScottChesworth I'd ordinarily agree. There's no point using Surge in generic UI mode. However, if, for some reason, the UI is deactivated, it's a number of seconds for NVDA to speak if you tab into the generic UI. This also happens *every* time that you tab into the UI. I've only gotten into this situation once, but it was very annoying.

CastleStarter:
Have you set up Surge XT to receive all keyboard input from Reaper?  You have to go to default settings in the context menu obtained when Surge XT is highlighted as a search result in the FX chain.

Regards, Chris G

Lo-lo78:
> Lag is still present in bridged mode, when changing filter type or when loading a new patch I get a few seconds of nothing. I've also applied this property page hack to NVDA and it doesn't seem to change anything, so that might be a separate issue. I use it in bridged mode because alt+comma/period to navigate the Surrge UI don't work in the chain.

It also works well in bridge mode. Just like in automatic mode for 64 bits.
I have not found any problems.

jcsteh:
So we seem to have some people this helps and others who say it doesn't help. Are we taking this, are we throwing it out or is there more feedback from others needed here?

jcsteh:
This should also fix most of #843.

ScottChesworth:
Re-reading feedback, seems like Surge in generic mode is still the only consistent lag. That's not a configuration that anyone is advised to choose in any resources I know of and there's nothing to be gained by choosing to run Surge that way, so I don't think OSARA needs to sweat over shoring up that situation. Good to go with it here.

AZSlow3:
Unfortunately the commit breaks SIBIAC logic... It discards role DIALOG from FX dialog and I was using it as a cheap (at least I think it is) way to detect some window is FX dialog and some other window is inside it, I mean parent top level dialog is FX dialog.

If you don't want declare FX dialog as a dialog, do you have any suggestion how should I search for it quickly in NVDA scripting?

jcsteh:
Ug. This is all very ugly.

I guess you could search for windowClassName `"#32770"` instead of role dialog?

AZSlow3:
Yes, role PANE, class  "#32770" with class of parent's parent "#32769" works. But somehow ugly.

I understand there had to be a good reason discard dialog flag from the dialog. But I think if it does not work good with screen readers, something else should be changed instead. I mean FX dialog is a top level dialog. REAPER window and sub-dialogs are not really dialogs, but this one is real.

PR #879, opened 2023-03-10
Make OSARA view parameters command open FX parameters anywhere in the FX chain, not just on the FX list.

### LINKED ISSUE #878: UX tightener: can Alt+P perform a single function throughout the FX window?
When the FX window is open, Alt+P has two results at the moment:
- If pressed with keyboard focus in the FX chain list, it goes to the FX parameters dialog.
- If pressed with keyboard focus anywhere else in the FX chain window, it loads track or take params depending on which is last-touched.

Could anyone shore up the UX a bit here? My thinking is that it should always load the FX params dialog because:
1. Context sensitivity. We're interacting with an effect any time we're in the FX window, ergo it's logical that Alt+P would only load FX parameters. It seems weird to me when it loads track/take params even though it might've been quite a while since I was last in the TCP.
2. It would be a handy workflow to choose a preset in the FX window, hit Alt+P to tweak params and then hit Escape to land back in the Presets Combo box. Best I can tell, isFXListFocused in fxChain.cpp already handles this nicely in one spot.
3. With partially accessible GUIs becoming more prevalent, there are times where I'll encounter a control with a value that I can read in the GUI, but can't readily adjust from the keyboard. It would be neat to be able to hit Alt+P and check whether that's also available as a parameter in these situations.

Thanks in advance for any tweaks. I did take a few shots before requesting, but so far I've only managed to annoy scons. Would be happy to keep poking at it if anyone's short on time to implement but does have time to share tips on how it could be done.

jcsteh:
Fixes #878.

@ScottChesworth , is this what you're after?

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1122,14 +1122,14 @@ class ItemParams: public ReaperObjParamSource {
 
 void cmdParamsFocus(Command* command) {
 	unique_ptr<ParamSource> source;
-	if (isFxListFocused()) {
-		int trackNum, itemNum, fx;
-		int type = GetFocusedFX(&trackNum, &itemNum, &fx);
+	int trackNum, itemNum, fx;
+	int fxType = GetFocusedFX2(&trackNum, &itemNum, &fx);
+	if (fxType && !(fxType & 4)) {
 		MediaTrack* track = trackNum == 0 ?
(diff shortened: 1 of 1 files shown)


ScottChesworth:
That's exactly it. Thanks!

PR #886, opened 2023-03-13
When moving the pitch cursor, the preview now respects the configured insertion note length.

### LINKED ISSUE #884: Can note previews respect length for next inserted note when using pitch cursor?
When navigating chords in the MIDI editor, note previews already respect the length of each note. However, when adjusting the pitch cursor with Alt+Up, Alt+Down, Alt+Shift+Up or Alt+Shift+Down, a generic preview length is played, I think we use a timer at the moment.
Could previews respect the current setting for lenth of next inserted note (Control+top row numbers)? Hearing contextual previews would make manual insertion of notes a more intuitive task, particularly for users who don't have a strong grasp of theory.

jcsteh:
Fixes #884.
As is so often the case with REAPER, this was trivial once I figured out what pieces in the massive jigsaw puzzle needed to fit in the right places. Of course, figuring that out is often non-trivial...

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1123,7 +1123,15 @@ void postMidiMovePitchCursor(int command) {
 	int pitch = MIDIEditor_GetSetting_int(editor, "active_note_row");
 	int chan = MIDIEditor_GetSetting_int(editor, "default_note_chan");
 	int vel = MIDIEditor_GetSetting_int(editor, "default_note_vel");
-	previewNotes(take, {{chan, pitch, vel}});
+	double start = GetCursorPosition();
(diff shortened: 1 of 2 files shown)


ScottChesworth:
(sidles in dressed as a waiter and quietly tops up your glass of genius juice)

ScottChesworth:
Just done some griddy stuff and this worked great. Does it need any specific testing or can we merge?

jcsteh:
The only obscure thing (that didn't work with my initial attempt) is when next inserted note length is set to grid, which has to be handled a bit differently to when there's a specific note length configured. However, i tested that and it seems to work, so I think we're good here.

PR #889, opened 2023-03-25
Add an Include unnamed parameters check box to Parameters dialogs.

### LINKED ISSUE #887: Consider introducing an option in the "FX parameters" dialog to allow filtering out the nameless parameters
There are quite a few instrument plugins out there that provide a massive FX parameter list with half or more of those parameters  being empty. Komplete Kontrol is one such instrument. This results in the "FX parameter" list being huge and very tedious to scroll through.
This is a proposal to add an option to the "FX parameter" dialog, possibly a checkbox, that would allow the user to filter out empty, i.e. with no names, parameters from the parameter menu. In doing so, the menu can be significantly reduced, thus reducing the number of keyboard presses and increasing the efficiency of browsing / discovering the parameters that actually do something.

jcsteh:
It is checked by default.
When unchecked, unnamed parameters are excluded from the parameter list. Currently, this means parameters named with the empty string or the single character '-'.
Fixes #887.


jcsteh:
@vick08 Please take this for a spin and let me know if it works for you. I haven't tested it with KK here, as it's on a different machine.

vick08:
No, unfortunately, that didn't change anything. I just turned on "all punctuations" in VOiceOver and, indeed, I hear "dash" in all empty parameters. This is just using Komplete Kontrol for now.
Thanks for trying!

jcsteh:
Just to clarify, this was with the new check box unchecked?

vick08:
Yes, after opening the fx parameter dialog, the checkbox was checked by default with a label "include empty parameters". I unchecked the checkbox and pulled down the menu again -- no change.

jcsteh:
The next build coming up in a few minutes should hopefully fix this.

vick08:
This worked very well and I am now able to reduce the list of 500 menu items to just 52, or something around that number. Thank you!
I wonder if instead of the checkbox, we had an edit field where a user could type a character sequence for parameter exclusion, perhaps even a regular expression?
I know this is a likely unpopular idea because it probably will only be used by very advanced users.
Outside of that, what should we do now? Thoughts?

jcsteh:
I think it would need to be a regular expression to be useful, but I also think that's far too advanced for most DAW users.

Internally, this uses a regular expression, so I guess we could just carefully extend that over time as new patterns are discovered. If you can come up with regular expression patterns that match unnamed parameters in other plugins that aren't so broad that they will break other things, I'm happy to add them.

Once you're happy with this, I guess I'll just merge it and see what happens.

vick08:
Sorry, forgot to comment here. Things look good as far as Komplete Kontrol goes.
Not sure if you would like to continue with other patterns after we check in this one.
Just in case, the other two patterns I observed are:
- letter P followed by a number (Guitar Rig)
- number sign followed by a number (Kontakt)

jcsteh:
Is that definitely a capital p, no spaces, etc.? Could you perhaps copy and paste one each of these names so we can be as precise as possible?

vick08:
Just double-checked.
For the Guitar Rig use case, we have

Uppercase "p" followed by the number.

For Kontakt we have

# (number sign) followed by the number.

jcsteh:
How many digits in each case?

vick08:
I have not seen more than three digits after the letter, but I noticed that, for example, for VST3 plugins, the number of parameters can go up to 3000 or 4000. Should we provision for that?

jcsteh:
Are the names like `#001` and P001? In that case, it's unlikely they'd go to 4 digits because they always use 3, even for numbers that don't require it.

jcsteh:
@vick08, could you please test this with Kontakt and Guitar Rig? Thanks.

vick08:
Thanks @jcsteh!
Just tested this commit with the mentioned plugins and everything behaves as expected. When the checkbox is unchecked, the empty parameters in those plugins are removed from the menu.

PR #892, opened 2023-03-26
Fix take volume in Parameters dialog when take polarity is flipped.

### LINKED ISSUE #891: When item phase is inverted, OSARA incorrectly displays take volume as -inf in its Item Parameters dialog
Passing along a report from @GianlucaApollaro, raised during a live Q&A.
Reproduction steps:
1. Insert some audio onto a track.
2. Select the item, and twiddle Control+Up/Control+Down to observe that the take volume reports as expected. It'll be at 0dB by default. All good.
3. With the item selected, hit Alt+P to view OSARA's Item Parameters dialog, observe that the slider and edit field associated with take volume param are both displaying expected values, then dismiss the dialog.
4. Hit Shift+F2 for Item Properties, Alt+I to check the "Invert phase" check box, then dismiss the dialog with Enter.
5. Retry Control+Up/Control+Down. Still all gravy, take volume is still reported as expected even with item phase inverted. But...
6. Hit Alt+P again for OSARA's Item Parameters dialog, and now, you'll find the slider and edit field associated with take volume param are both displaying a value of -inf. DOH!
7. Set item phase back to normal, the take volume values in our dialog also return to what's expected.

Going on my haphazard commenting out of random stuff, it seems like this line of code in reaper_osara.cpp is what keeps Control+Up/Control+Down reporting as expected when item phase is inverted:
"volume = fabs(volume);"
I couldn't figure out how to incorporate that into paramsUi.cpp though, the templating stuff in there broke my brain.

Thoughts?

jcsteh:
In this case, the raw value is negative. We need to flip it accordingly when both retrieving and setting it. It was easiest to do this in ReaperObjVolParam, even though this only affects takes. We check whether the current raw value is negative during construction and set a flag so we know whether the sign needs to be flipped thereafter. This avoids checks each time we set the value; i.e. we only check this the first time the parameter is constructed.
Fixes #891.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -170,10 +170,18 @@ class ReaperObjVolParam: public ReaperObjParam {
 		this->step = 0.002;
 		this->largeStep = 0.1;
 		this->isEditable = true;
+		if (this->getValue() < 0) {
+			// Take volume raw values are negative when the polarity is flipped.
+			this->flipSign = true;
+		}
 	}
 
 	double getValue() {
-		return *(double*)this->provider.getSetValue(nullptr);
+		double result = *(double*)this->provider.getSetValue(nullptr);
+		if (this->flipSign) {
+			result = -result;
+		}
+		return result;
 	}
 
 	string getValueText(double value) {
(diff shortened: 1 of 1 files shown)


ScottChesworth:
Working nicely here, thanks man! Is it merge-ready?

PR #924, opened 2023-08-19
Fix focusing the MIDI Filter Events window in non-English REAPER.

### LINKED ISSUE #922: Filter: Show/hide filter window
Hi,
2 users have reported to me that the following action no longer works with OSARA in the MIDI editor.
Filter: Show/hide filter window...
I tested here on 3 machines, indeed, the focus of NVDA no longer goes in the filter window on 2 of the 3 machines tested.
All tests were performed with the latest version of Reaper and OSARA on Windows 10.
But it works perfectly on one of my 3 PCs, why, no idea.
I asked a sighted person and he confirmed that there is no window blocking access or error message on my 2 problem PCs.
The filter is well displayed on the screen but the focus of NVDA does not move in the window.
I then moved the REAPER.ini file to another location so that Reaper builds a new .ini file and I have exactly the same result on both machines.
On the other hand, the machine that works is a machine on which I always have a fresh installation of Reaper.
As soon as the new version of Reaper is available, I completely uninstall Reaper and delete all files including those in Roaming folder.
This allows me to do tests with a clean installation in order to compare strange behavior.
I then tried with a portable version of Reaper and I have the same behavior on both machines.
So, I don't know what to do anymore to solve this problem.
Any ideas?

Thanks.

jcsteh:
The only way we can currently find this window is via its title. Previously, this title was a hard-coded English only string. Instead, we now use REAPER's own localisation functionality to get the string from the REAPER lang pack.
Fixes #922.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1613,7 +1613,8 @@ void cmdMidiFilterWindow(Command *command) {
 	HWND editor = MIDIEditor_GetActive();
 	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
 	// TODO: we could also check the command state was "off", to skip searching otherwise
-	HWND filter = FindWindow(WC_DIALOG, "Filter Events");
+	HWND filter = FindWindowW(L"#32770",
+		widen(LocalizeString("Filter Events", "midi_DLG_128", 0)).c_str());
 	if (filter && (filter != GetFocus())) {
 		SetFocus(filter); // focus the window
 	}
--- a/src/osara.h
+++ b/src/osara.h
@@ -208,6 +208,7 @@
 #define REAPERAPI_WANT_TimeMap2_timeToQN
 #define REAPERAPI_WANT_MIDI_GetGrid
 #define REAPERAPI_WANT_GetParentTrack
+#define REAPERAPI_WANT_LocalizeString
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
 


jcsteh:
This is a speculative fix based on @RDMurray's suggestion in https://github.com/jcsteh/osara/issues/922#issuecomment-1682971705. I haven't tested it myself with non-English REAPER.

@leejul, would you mind taking a build for a spin when it arrives here?

cn-peter:
I am a Chinese user. I thought the midi filter was a reaper problem.
let me try.

cn-peter:
This version is still the same as before, no changes.

jcsteh:
Oh, also, focusing Filter Events only works on Windows. WE can't make it work on Mac.

jcsteh:
Actually, please try again with the next build that will arrive here in a few minutes. I realised the previous code might not work with languages where there are non-ASCII characters in the translated string, which is certainly true for Chinese.

leejul:
Sorry, I was gone for a few days, so here it is.
I tried osara_2023.1pre-1168,0387fafe and the problem is still not solved.
But do I have to install a specific version?
Thanks.

ScottChesworth:
The proposed solution hasn't been merged into a snapshot yet. Here's the build to test:
https://ci.appveyor.com/api/buildjobs/65h6hfqvuaa0xehq/artifacts/installer/osara_pr924-1167,d29c048a.exe

leejul:
Yes! It works perfectly. Thank guys.

cn-peter:
very perfect

PR #932, opened 2023-09-18
Add feedback for Track: Duplicate tracks

RDMurray:
trivial change to add feedback for Track: duplicate tracks, Alt+D. Feedback is the same as when pasting tracks.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -86,6 +86,7 @@ Most of these are actions built into REAPER, but a few are useful actions from t
 - SWS: Select nearest next folder: Alt+PageDown
 - SWS: Select nearest previous folder: Alt+PageUp
 - Track properties: Toggle free item positioning: Alt+Shift+F2
+- Track: Duplicate tracks: Alt+D
 
 #### Adjusting Track Parameters
 - Track: Mute/unmute tracks: F5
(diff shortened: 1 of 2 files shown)


ScottChesworth:
Thanks!

ElCrackLoko:
Hello, we translators have problems with this feedback. At least in Spanish Mexico.
This bug comes from a long time ago when only the word "add" was added for one or more tracks when pasting them with CTRL+V.
In Spanish, add as such is a transitive verb and we cannot say, "una pista agregar", but rather "una pista agregada" or, better yet, "se agregó una pista" "una=one" singular, if there are 2, 3, 4 tracks or more, the word in conjunction with the phrase changes, it would no longer be singular "se agregó una pista" but rather "se agregaron varias pistas", "varias", plural. Now, as a translator I thought the best for everything without having to bother you and instead of translating add as "agregar", a better phrase was added for "desde el portapapeles", in English, from the clipboard, since it doesn't matter there If it is one or more, that phrase coincides in plural and singular.
But now that same phrase is being used to duplicate tracks so it loses all meaning since when duplicating one or more tracks with Alt+D, we do not do it from the clipboard with CTRL+V. IF you add a line, a different comment like, a duplicate clue for the singular or 2 duplicate clues for the plural, that would be great.

jcsteh:
The message that should be used when pasting or duplicating tracks is "{} track added" or "{} tracks added", which should allow you to specify the correct plural. Are you not seeing the singular/plural message in the translation for some reason?

jcsteh:
Ignore that. I'm wrong. Sorry.

I'd prefer not to have separate messages for pasting and duplication, as the net impact for the user is the same. The correct fix is to fix the message so that it is "{} track added" and "{} tracks added". I'll do that when I get some time.

jcsteh:
Ah damn it. Now I realise why we didn't do that in the first place. It's possible to paste both tracks and items at the same time, so you can get "2 tracks 1 item added". Hmm. Not sure how to deal with this yet.

ElCrackLoko:
Don't worry, I realized that too and that's why in Spanish Mexico we decided to add the message from the clipboard when pasting clues, because the items are masculine and the clues are feminine, so in Spanish they would never combine . You can create the same word of add to the action of duplicate and simply in Spanish we would change it to the word duplicate.

jcsteh:
Sure, and I appreciate that, but this isn't necessarily going to be consistent across languages. I'd prefer a more generic solution if I can come up with one.

PR #937, opened 2023-09-30
Make shift+space toggle item selection in the MIDI Editor Event List.

### LINKED ISSUE #869: Can Shift+Space become the keystroke for toggling selection in list box controls?
It's been raised on RWP that as things stand, it's not possible to pause when in the MIDI event list view. That's a list box control where on Windows, Control+Space toggles selection of the highlighted event instead of pausing. That's a sticky wicket, because ideally we need to be able to do both of the things Control+Space does.

I know it's not standard for the OS, but would it be possible to hack it so that Shift+Space toggles selection of the highlighted entry when in list box controls? That way, even when it's not OSARA's non-contiguous selection mode handling the workflow, users would at least have a consistent keystroke.

Another alternative could perhaps be to switch around what's bound to Control+Space and Shift+Space, but that would invalidate so many established training resources, I'd rather avoid going that route if possible.

Thoughts? How much gruntwork is it? As per usual I don't know how this would be done, but am happy to tinker and research if someone wants to point me at relevant code.

jcsteh:
This allows control+space to play/pause as usual.
Fixes #869.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1648,6 +1648,16 @@ void maybeHandleEventListItemFocus(HWND hwnd, long childId) {
 	previewNotes(take, {note});
 }
 
+void toggleListViewItemSelection(HWND list) {
+	const int item = ListView_GetNextItem(list, -1, LVNI_FOCUSED);
+	if (item == -1) {
+		return;
+	}
+	UINT prevState = ListView_GetItemState(list, item, LVIS_SELECTED);
+	ListView_SetItemState(list, item,
+		prevState == LVIS_SELECTED? 0 : LVIS_SELECTED, LVIS_SELECTED);
+}
+
 #endif // _WIN32
 
 void postMidiChangeVelocity(int command) {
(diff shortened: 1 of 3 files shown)


jcsteh:
@ScottChesworth, I tried this out and I have to say I don't love it. When I'm selecting list items non-contiguously with control+arrows, my muscle memory leads me to press control+space, which of course plays/pauses now.

An alternative to this could be to bind shift+space to play/pause, but only in the MIDI Event List. That avoids breaking muscle memory and training material everywhere else, but still means the Event List is inconsistent.

With either approach, we cause problems for muscle memory. I guess the question is whether it's more common to play/pause or non-contiguously select in the Event List. Since you have more dealings with users, I'm happy for you to make that call; I can't make a compelling argument either way.

LeonarddeR:
I came across this and looking at the outstanding issue, I have no problem with the event list being a bit inconsistent here, so I'd vote for shift+space for play/pause. The event list is a syslistview anyway, so I wouldn't break the native non-contiguous selection behavior for that.

PR #940, opened 2023-10-06
Add a .clang-format file

### LINKED ISSUE #923: Provide .ClangFormat for auto formatting
When writing OSARA code, all of us format code manually as far as I can see.
I think adding a basic .ClangFormat file can be beneficial for us using an IDE that supports auto formatting with Clang, such as VS Code.

LeonarddeR:
I invested an hour or two writing down a .clang-format that more or less follows the code style that is now common in OSARA.
@jcsteh What would you like? Should I add a reformatted version of one file as well? Then you and probably others can comment on changed lines you dislike.
Before merge we can revert these changes, then after merge I can provide a pr that should be with the whole code base formatted, including a [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view). We should merge that using rebase and merge or a regular merge commit, not squashing.

Fixes #923


.clang-format
```diff
@@ -0,0 +1,104 @@
...
+BinPackArguments: false
+# Do not pack function parameters together.
+# A function declaration’s or function definition’s parameters will either all be on the same line
+# or will have one line each.
+BinPackParameters: false
+# Add a space after the colon in bit fields.
+BitFieldColonSpacing: After
+# Break before binary operators that are not assignments.
+# E.g. break before "+", but not before "+=".
+BreakBeforeBinaryOperators: NonAssignment
```
jcsteh, line 50: Hmm. We're pretty inconsistent on this one. Sometimes, we put `&&` at the end of the line, but other times, we put it at the start. That's mostly (maybe entirely?) my fault; I think my perspective on this has changed over time. Do you have a preference personally?

.clang-format
```diff
@@ -0,0 +1,104 @@
...
+AllowShortEnumsOnASingleLine: false
+# Only allow empty functions on a single line.
+AllowShortFunctionsOnASingleLine: Empty
+# Do not allow short if statements on a single line.
+AllowShortIfStatementsOnASingleLine: Never
+# Do not allow short loops on a single line.
+AllowShortLoopsOnASingleLine: false
+# Do not pack function call arguments together.
+# A function call’s arguments will either be all on the same line or will have one line each.
+BinPackArguments: false
```
jcsteh, line 41: I'm not sure about this one or BinPackParameters. I don't love splitting a function call into many lines when there are really short parameters. I also don't think this is something we do often in current OSARA code, so this one seems like it might cause a lot of churn.

.clang-format
```diff
@@ -0,0 +1,104 @@
...
+BitFieldColonSpacing: After
+# Break before binary operators that are not assignments.
+# E.g. break before "+", but not before "+=".
+BreakBeforeBinaryOperators: NonAssignment
+# Attach opening braces to the same line as the statement or declaration it belongs to.
+BreakBeforeBraces: Attach
+# Break string literals if they are too long.
+BreakStringLiterals: true
+# Maximum column width before code will be wrapped or formatted.
+ColumnLimit: 110
```
jcsteh, line 56: I'm weird about maximum line widths in general, but is there a reason you chose 110? There seems to be a fair amount of standardisation around 80 and I'm trying to conform to this in newer code, even though I didn't in older code. I don't really think 80 should be significant these days, but in the absence of a good argument, I'm inclined to lean towards the most common convention.

.clang-format
```diff
@@ -0,0 +1,104 @@
...
+# Sort includes (CaseSensitive, CaseInsensitive, or None).
+SortIncludes: CaseInsensitive
+# C++ standard to be followed (e.g., c++98, c++11, c++14, c++17, c++20).
+Standard: c++20
+# Add a space before opening parentheses in control statements.
+SpaceBeforeParens: ControlStatements
+# Do not add a space before square brackets.
+SpaceBeforeSquareBrackets: false
+# Width of a tab character.
+TabWidth: 4
```
jcsteh, line 102: I wonder if we should base things around 2 spaces per tab rather than 4? If we were using spaces instead of tabs, I'd be moving to 2.

.clang-format
```diff
@@ -0,0 +1,104 @@
...
+# Separate definition blocks (Always, Never, or Leave).
+SeparateDefinitionBlocks: Always
+# Number of spaces before trailing comments.
+SpacesBeforeTrailingComments: 1
+# Do not add a space after a C-style cast.
+SpaceAfterCStyleCast: false
+# Do not add a space after the template keyword.
+SpaceAfterTemplateKeyword: false
+# Sort includes (CaseSensitive, CaseInsensitive, or None).
+SortIncludes: CaseInsensitive
```
jcsteh, line 94: I think there are some edge cases where include order matters; e.g. see the inclusion of osara.h in fxChain.cpp and others. Will this break that?

jcsteh (commented):
Thanks very much for working on this.

I think it might be worth seeing the diff for reaper_osara.cpp. It's the biggest and oldest file, so this will give us a good idea of how much churn there will be and how we might minimise it.

.clang-format
leonardder: All right, I'll remove that one and we'll see what this does.

.clang-format
leonardder: I think i preferp putting it at the start. For example, in:
```cpp
if (
    something
    && something2
//    && something3
)
```

You can easily comment out lines, even the last predicate of the check.

.clang-format
leonardder: I choose 110 because it is used in NVDA 😳
How about leaving it out? Then we'll stick to Google's default, whatever it is. Probably 79 or something.

.clang-format
leonardder: I'll have a look and will act accordingly.

.clang-format
leonardder: AH, that's fine with me.

leonardder:
I addressed your comments and added one commit with the whole code base formatted, so you can have a look on what it will do.

.clang-format
```diff
@@ -0,0 +1,95 @@
...
+# Author: James Teh <jamie@jantrid.net>
+# Copyright 2023 Leonard de Ruijter
+# License: GNU General Public License version 2.0
+
+# Style based on Google's C++ style guide.
+BasedOnStyle: Google
+# Offset for access modifiers (public, protected, private).
+AccessModifierOffset: 0
+# Align code after an open bracket with block indentation.
+AlignAfterOpenBracket: BlockIndent
```
jcsteh, line 12: I think we should use DontAlign here. Right now, a lot of code puts arguments on the same line as the function call, only breaking the line when more would make the line too long. Using DontAlign would result in less churn.

```suggestion
AlignAfterOpenBracket: DontAlign
```

jcsteh: Ug, this causes long if statements to have their final line indented to the same level as the inner block. That's... annoying. We could fix this with ContinuationIndentWidth: 4. That affects function calls as well, but that's probably okay.

jcsteh:
We're going to need to tweak makePot so that it can operate across lines. Right now, I think it expects translate, translate_plural, etc. to be on a single line. When we re-format, translate_plural in particular might get split across lines. Unfortunately, that's going to be non-trivial because makePot currently walks line by line.

jcsteh:
One way to do that might be to create separate regexps to match just the beginnings of these functions; e.g. RE_CPP_TRANSLATE_PLURAL_BEGIN. If we match RE_CPP_TRANSLATE_PLURAL_BEGIN but not RE_CPP_TRANSLATE_PLURAL, we repeatedly add the next line until RE_CPP_TRANSLATE_PLURAL matches the whole string. Though... that may cause problems if we have multiple translate calls on a single line. Urgh.

leonardder:
May be a strange question, but why are we actually using a special script to create pot files rather than xgettext?

jcsteh:
Not a strange question at all. The reason is that xgettext doesn't support some of the things we need such as .rc files and string lists (translateFirstString, etc.).

leonardder:
Actually with my last change, I think we are now pretty safe. WhitespaceSensitiveMacros will just leave alone the translate and translate_plural calls.

src/uia.cpp
```diff
@@ -241,11 +239,10 @@ bool sendUiaNotification(const string& message, bool interrupt) {
...
-		uiaProvider,
-		NotificationKind_Other,
-		interrupt ? NotificationProcessing_MostRecent : NotificationProcessing_All,
-		SysAllocString(widen(message).c_str()),
-		SysAllocString(L"REAPER_OSARA")
-	) == S_OK);
+	return (
+		uiaCore->RaiseNotificationEvent(uiaProvider, NotificationKind_Other,
+			interrupt ? NotificationProcessing_MostRecent
+								: NotificationProcessing_All,
```
leonardder, line 245: I wish I knew how we could avoid these random tabs being added here. I'd say AlignOperands would do the trick nicely here, but it seems it doesn't.

jcsteh: So AlignOperands: DontAlign doesn't work?

.clang-format
```diff
@@ -0,0 +1,100 @@
...
+# Do not add a space before square brackets.
+SpaceBeforeSquareBrackets: false
+# Width of a tab character.
+TabWidth: 2
+# Always use tabs for indentation.
+UseTab: Always
+# Don't touch lines used for translation.
+WhitespaceSensitiveMacros: [
+	'translate_plural',
+	'translate'
```
jcsteh, line 99: translate_ctxt should probably be added here too. Maybe I'll fix this properly one day and we can get rid of this hack. :)

src/uia.cpp
leonardder: We definitely shouldn't touch ContinuationIndentWidth, that makes things worse.

.clang-format
leonardder: I have investigated that. Personally I really dislike the style where the closing paren after a list of function parameters isn't at the same level as the function name. For that, we really need BlockIndent.

leonardder:
I just took a different tack and took the Microsoft style as a basis. This seems to be going a lot better.

.clang-format
jcsteh: But this is already what we have. There are a lot of places where the code does things like this:

```
    foo(bar, baz,
        blahblah);
```

Codifying this means we get less churn.

leonardder: This works for function calls, but for function definitions, it is either:

```
int foo(
	int bar, int bazz) {
```

or

```
int foo(int bar,
	int bazz) {
```

Wherea's I'd probably prefer:

```
int foo(int bar,
	int bazz
) {
```

As we use in several places.

jcsteh: If I understand correctly, DontAlign should mean we get this if we set continuation indentation to 4:

```
int foo(int bar,
		int baz) {
	...
}
```

So it's possible to distinguish between the function parameters and the body because the parameters are further indented than the body is. That said, I agree it does look particularly weird for function definitions, and it doesn't look like we can differentiate between definitions and calls. That's unfortunate because breaking calls up like that does mean a lot of extraneous nearly empty lines. Ah well.

src/envelopeCommands.cpp
```diff
@@ -184,23 +185,24 @@ optional<int> currentEnvelopePoint{};
...
-		translate("fast end"),
-		// Translators: A shape for an envelope point.
-		translate("bezier"),
-	};
+			// Translators: A shape for an envelope point.
+			translate("linear"),
+					// Translators: A shape for an envelope point.
+					translate("square"),
+							// Translators: A shape for an envelope point.
+							translate("slow start/end"),
```
jcsteh, line 193: Wo. What on earth is going on here? It seems to keep adding continuation indents even though each translate call is closed. I wonder if WhitespaceSensitiveMacros is interfering here somehow?

jcsteh: It does seem that AlignAfterOpenBracket: DontAlign makes this worse. But it's definitely very confused. If I add random indentation to these lines, it preserves that even with AlignAfterOpenBracket: BlockIndent. So I guess the whitespace sensitivity applies to the line before the macro, not just inside the macro, which is very unfortunate.

I think I'm going to need to fix makePot... somehow.

jcsteh:
This severely breaks translate firstString comment blocks. I guess the only way to fix that is to disable clang-format for those sections, which seems pretty bad. Alternatively, i guess we could switch to a no-op macro purely for makePot which signifies that a string is translatable without actually trying to translate it. That might also allow us to switch to xgettext, though we'd still need custom code for the rc file.

src/envelopeCommands.cpp
leonardder: I'm not sure. There's also some black magic going on in uia.cpp I referenced earlier, which I think is related. It really looks like we are missing an alignment setting somewhere, but I'm pretty sure we have covered them all.

I'm starting to regret this all.

jcsteh: I really do appreciate you working on this. Even if we end up shelving it for a while, we can always come back to it.

I'll look at fixing makePot at some point to cope with multi-line calls. I have some ideas on how it might be done and that's probably a good idea for other reasons anyway.

Losing translateFirstString is a bit of a shame, but I guess that was always a bit sketchy anyway, and not having to worry about code formatting eventually (which is a total waste of people's time) probably outweighs that loss.

leonardder: Sure. I will mark this as draft for now.

.clang-format
```diff
@@ -0,0 +1,94 @@
...
+# Author: James Teh <jamie@jantrid.net>
+# Copyright 2023 Leonard de Ruijter
+# License: GNU General Public License version 2.0
+
+# Style based on Google's C++ style guide.
+BasedOnStyle: Google
+# Offset for access modifiers (public, protected, private).
+AccessModifierOffset: 0
+# Align code after an open bracket with block indentation.
+AlignAfterOpenBracket: DontAlign
```
jcsteh, line 12: Hmm, I don't love what this does to some data structures. I'm pondering switching back to BlockIndent here. I do feel BlockIndent wastes lines in some cases - a function call that could be 2 lines ends up being 4 - but at least it's consistent across the board. I wonder if this would also fix the weirdness in uia.cpp. I'll have to try it.

leonardder: > I wonder if this would also fix the weirdness in uia.cpp.

I'm afraid not. This might just be a weird bug in clang-format itself.

.clang-format
leonardder: Right, we definitely want to leave include ordering alone.

jcsteh:
I wonder if we can get around the include problem with tightly scoped clang-format ignores. It'd be nice if we didn't have to worry about include ordering except in very specific cases.

src/controlSurface.cpp
```diff
@@ -214,17 +212,18 @@ class Surface: public IReaperControlSurface {
 		postGoToTrack(0, track);
 	}
 
-	virtual int Extended(int call, void* parm1, void* parm2, void* parm3) override {
+	virtual int Extended(int call, void* parm1, void* parm2, void* parm3)
+		override {
```
jcsteh, line 216: Oof. This is a bit ugly. Shouldn't the override be indented to continuation width at least so that the function body and the header are at different levels? It's not a huge deal, but it's not great.

src/reaper_osara.cpp
```diff
@@ -2589,15 +2881,11 @@ bool maybeFixTabInSaveDialog(bool previous) {
...
+// some windows.
 LRESULT CALLBACK keyboardHookProc(int code, WPARAM wParam, LPARAM lParam) {
 	const bool isKeyDown = !(lParam & 0x80000000);
-	if (!isKeyDown || code != HC_ACTION || (
-		wParam != VK_APPS && wParam != VK_CONTROL &&
-				wParam != VK_F10 && wParam != VK_RETURN &&
-				wParam != VK_F6 && wParam != 'B' &&
-				wParam != VK_TAB && wParam != VK_DOWN &&
-				wParam != VK_UP && wParam != VK_SPACE)) {
+	if (!isKeyDown || code != HC_ACTION || (wParam != VK_APPS && wParam != VK_CONTROL && wParam != VK_F10 && wParam != VK_RETURN && wParam != VK_F6 && wParam != 'B' && wParam != VK_TAB && wParam != VK_DOWN && wParam != VK_UP && wParam != VK_SPACE)) {
```
jcsteh, line 2888: What the actual hell? What happened to the maximum line length? This actually makes this much harder to read.

src/reaper_osara.cpp
```diff
@@ -915,15 +934,16 @@ void postInvertTrackPhase(int command) {
 	MediaTrack* track = GetLastTouchedTrack();
 	if (!track)
 		return;
-	outputMessage(isTrackPhaseInverted(track) ?
-		translate("phase inverted") :
-		translate("phase normal"));
+	outputMessage(
+		isTrackPhaseInverted(track) ? translate("phase inverted")
+																: translate("phase normal")
```
jcsteh, line 939: Definitely something wacky going on with ternary operators. It's like it's trying to align horizontally instead of using continuation indentation.

src/reaper_osara.cpp
```diff
@@ -4153,245 +4513,653 @@ void cmdReportRegionMarkerItems(Command* command) {
...
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to next CC")}, "OSARA_NEXTCC", cmdMidiMoveToNextCC},
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to previous CC")}, "OSARA_PREVCC", cmdMidiMoveToPreviousCC},
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to next CC and add to selection")}, "OSARA_NEXTCCKEEPSEL", cmdMidiMoveToNextCCKeepSel},
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to previous CC and add to selection")}, "OSARA_PREVCCKEEPSEL", cmdMidiMoveToPreviousCCKeepSel},
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to previous midi item on track")}, "OSARA_MIDIPREVITEM", cmdMidiMoveToPrevItem},
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to next midi item on track")}, "OSARA_MIDINEXTITEM", cmdMidiMoveToNextItem},
-	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Select all notes with the same pitch starting in time selection")}, "OSARA_SELSAMEPITCHTIMESEL", cmdMidiSelectSamePitchStartingInTimeSelection},
-	{ MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Mute next message from OSARA")}, "OSARA_ME_MUTENEXTMESSAGE", cmdMuteNextMessage},
+	{MAIN_SECTION,
+	 {DEFACCEL, _t("OSARA: Move to next item (leaving other items selected)")},
```
jcsteh, line 4790: And here we don't seem to be using block indent, despite AlignAfterOpenBracket. IN fact, we end up using a space here instead of a tab, probably because of the single brace. I don't get it. We should be using block indent or continuation indent, never horizontal align.

jcsteh:
Oh. clang-format has this to say about AlignAfterOpenBracket: BlockIndent:

> This currently only applies to braced initializer lists (when Cpp11BracedListStyle is true) and parentheses.

I mean... I would have thought if conditions count as parentheses, but apparently not. But it does explain why it doesn't impact braces.

leonardder:
> I wonder if we can get around the include problem with tightly scoped clang-format ignores. It'd be nice if we didn't have to worry about include ordering except in very specific cases.

I will look into that later today/this week.

src/reaper_osara.cpp
leonardder: Wow this is extremely weird indeed.

src/controlSurface.cpp
leonardder: Slightly nasty indeed. We probably need an additional rule for this to avoid breaking.

leonardder: I actually reverted the continuation width as part of switching back to BlockIndent, but that's probably not what we want. It's back at 4 now.

src/reaper_osara.cpp
leonardder: I actually fixed continuation indent and now it seems to break up properly again.

leonardder:
Ii now chose the Microsoft style as a basis, which has a 120 character column limit. At least that saves us some very ugly code breaking.

src/reaper_osara.cpp
leonardder: It now adds breaks again, but insists to add an extra space to the second line:

```
	if (!isKeyDown || code != HC_ACTION ||
			(wParam != VK_APPS && wParam != VK_CONTROL && wParam != VK_F10 && wParam != VK_RETURN && wParam != VK_F6 &&
			 wParam != 'B' && wParam != VK_TAB && wParam != VK_DOWN && wParam != VK_UP && wParam != VK_SPACE)) {

```

I'm utterly confused!

jcsteh:
Which breaks were bothering you with the Google style? If they're a real problem, can we just increase the max line length there?

leonardder:
We're now back at Google, and indeed, it formats some stuff slightly nicer than Microsoft.

Timtam:
Just coming here because i'm noticing how long it takes to fix things like indentation all the time. I noticed that you seem to struggle quite alot with getting clang-format to work. Is there any specific reason why you absolutely want to go with this over any other, newer solution like a prettier or something similar? AFAIK clang-format works, but is pretty old and most likely outdated compare to more modern tools. We already have  Python toolchain with scons, we could go for a Python-based implementation of a similar tool instead, maybe that would work better? Just a thought though.

jcsteh:
I'm curious as to what you're suggesting is a more modern tool. As I understand it, clang-format is the industry standard these days.

The main problem is that OSARA's existing formatting isn't really compatible with any of the more prevalent C++ coding styles. We could just reformat the whole code base without caring, but that isn't great for anyone already familiar with contributing to the project. Also, there are sections of the code that have been formatted the way they have for very specific reasons, so we need to consider the trade-offs there.

Timtam:
> I'm curious as to what you're suggesting is a more modern tool. As I understand it, clang-format is the industry standard these days.

I'm not too deep into "the industry" anymore. Last time I was working for a tech company they were doing all the formatting with a pre-commit hook through the IDE, namely IntelliJ. Dunno if that one uses clang-format under the hood, but I kinda doubt it. I know other people who do all the formatting in IDE too, tools like clang-format, prettier and such usually just are fallbacks there, but obviously that is nowhere near the entire picture.

> We could just reformat the whole code base without caring, but that isn't great for anyone already familiar with contributing to the project.

IMO the advantage of not having to care about the entire formatting at all is way, way more important than keeping the current formatting in tact. I watched Scott and Jen navigate the codebase multiple times by now, kinda all that matters is that you can count the braces and parens there, indentation is there but usually gets skipped anyway, and nooone is entirely sure how to indent properly anyway in specific cases like line breaks within a function call. IMO the actual way of formatting isn't as important as having the ability to just write down your code without having to care about indents and stuff, hit a button before committing and just ignore code formatting at all. That is basically what all my sighted colleagues do too from what I can tell. Noone nowadays cares about counting tabs and breaking lines correctly.

> Also, there are sections of the code that have been formatted the way they have for very specific reasons, so we need to consider the trade-offs there.

Just put them onto a ignore list and don't format them at all? I know, sounds easier than it probably is, and I know you discussed that already. Could you give me a rundown why exactly that doesn't work properly with clang-format? I think I remember something along the lines of clang-format messes other things up if you do that? That might be one reason why it might be worth investigating other tools to see if they handle this case better.

jcsteh:
There's also the fact that this would break all existing open pull requests, some of which we can't merge yet for various reasons.

LeonarddeR:
@jcsteh What do you think about this? At some point, we might just want to settle on something that is not ideal per se

jcsteh:
I made a few tweaks here. I set the ColumnLimit to 80, which does introduce a bunch more lines, but I think it's nicer than lines which drag on forever. It's not that I specifically care about 80 character lines, more that very long lines can get really intense. I actually couldn't figure out what value was implicitly being used here. Do you know what it was?

Do you know if there's a way to prevent clang-format from trying to align comments when it splits a trailing comment across lines? Even though we have AlignTrailingComments: never, it still wants to align them if it has to split them; e.g. in the command maps. IMO, visual alignment (vs semantic alignment) is never something we want, but I can't find a way to disable this case. That said, if I get rid of the ColumnLimit, I guess this problem goes away, so... hmm.

jcsteh:
Oh, looks like it's 120. Maybe that's fine and I should leave this alone. What do you think?

jcsteh:
I did a before and after with scons pot and it generates the same .pot file, minus comment line wrapping differences. So that's good.

LeonarddeR:
@jcsteh I (or rather, Claude) picked this up again. The file is rewritten, and the pull request now also contains the reformat of the whole code base and a formatting check, so you can see exactly what the file does; the description is updated. 

Things that changed:
- the column limit is now 100, between 80 and 120
- operators at the start of wrapped lines
- `# include` indentation for nested directives

May be we should diverge the formatting to its own pr, then we can just do a fast forward merge with a follow up containing the formatting and the blame ignore refs.

jcsteh:
Thanks. I'll take a look at this. I keep looking at it every now and then, but there's always something about the formatting that really annoys me and looks far worse than it is already.

I think 80 characters is closer to where the code base is already at. Certainly, I've been formatting manually with 80 characters for years. The major exceptions are the action tables, which honestly look far worse than they do now regardless of how we configure clang-format.

> When a call or lambda is an operand of << or || and its contents wrap, those contents are indented from the operand's column, three characters right of the operator. About 64 lines in the code base look like that.

Hmm, that really isn't ideal. We don't want space indentation for anything at all.

> A comment that is the first thing after { is pulled onto the line of the brace. Commit 2 moves the three translator comments this affected; one ordinary comment in a regex literal list in paramsUi.cpp stays where clang-format puts it.

That is... an odd choice. I'm confused though. If clang-format moves the comments, how do you deal with that for cases where you don't want that behaviour? It's also not something I've seen clang-format do before - e.g. it doesn't do that in the Mozilla code base - so I'm wondering whether there's an option for it.

LeonarddeR:
Thanks for looking into it, this makes sense. The following was changed now:

1. Back to 80 columns

2. The space indentation is a clang-format bug really: when a call wraps *inside* a stream chain or a braced list, clang-format indents the contents from the operand's column, which lands three characters right of the operator and cannot be expressed in tabs. It is reported as [#59797](https://github.com/llvm/llvm-project/issues/59797), with [#67897](https://github.com/llvm/llvm-project/issues/67897) and [#65008](https://github.com/llvm/llvm-project/issues/65008) covering the same thing for `AlignOperands`; all still open, the oldest since January 2023.

	What does avoid it is not writing the construct that triggers it. So there is now a commit that rewrites the 29 statements involved (18 stream chains, 8 lambdas, 2 conditionals and one chained assignment); statements clang-format already indents with tabs are left alone:

	- a chain that inserts several parts becomes one statement per part, so `s << a << b;` where a part wraps is written as `s << a;` and `s << b;`
	- a conditional that wraps inside a chain is assigned to a variable first
	- a lambda passed inside a braced list is given a name before the call

	This ensures there's no space indentation left at all.

3. The comment moved onto the brace line is clang-format behaviour too. A comment that is the first token inside a braced initializer is placed on the line of the brace, and the elements after it are aligned with the comment. The way to keep such a comment on its own line is to put it above the statement that opens the brace. The preparation commit does that for the two translator comments where it mattered, because makePot only recognises `// Translators:` at the start of a line. To make sure this cannot regress silently, makePot now fails with an error if a translators comment ends up after code on a line. The one remaining case is an ordinary comment inside a regex in paramsUi.cpp, which is harmless.

jcsteh:
> a lambda passed inside a braced list is given a name before the call

Can you give me an example? I don't quite follow this.

> The one remaining case is an ordinary comment inside a regex in paramsUi.cpp, which is harmless.

The comment is below the brace because it's intended to be specifically relevant to the regex fragment directly below it. With it on the same line as the brace, it looks like that comment is relevant to the entire expression. So I don't think it's harmless, but it's also one minor instance in the entire code base, so I guess I can live with it; don't let the perfect be the enemy of the good and all that. :)

The other thing I forgot to mention is operands at the start of lines. The code base is currently inconsistent in that regard. However:

```
$ git grep '^\s*[&|]' | wc -l
29
$ git grep '[&|].$' | wc -l
63
```

That makes it fairly clear that the prevailing style is for `&&` and `||` to be at the end. On the other hand:

```
$ git grep '^\s*+' | wc -l
59
$ git grep '+.$' | wc -l
4
```

So for `+`, the prevailing style is at the start of the line. So that makes it annoyingly tricky to decide on the correct path here. :) I'm assuming the same directive handles both, so if we're considering both, I guess start of line wins.

LeonarddeR:
Here is a lambda example:

```cpp
options.push_back(
	{translate("Show/hide &envelope"), [this] {
		 return this->showHideEnvelope();
	 }}
);
```

Becomes:

```cpp
auto showHideEnvelope = [this] { return this->showHideEnvelope(); };
options.push_back({translate("Show/hide &envelope"), showHideEnvelope});
```

The same for the other seven lambdas, all in paramsUi.cpp. I think I like this a lot actually.

The comment in the regex was a leftover from the 100 columns experiment. Now the declaration on its own is long enough that clang-format breaks after the brace, so the comment stays on its own line, directly above the fragment it describes:

```cpp
static const regex RE_UNNAMED_NONAUTOMATABLE_PARAM{
	// Any one of several strings...
	"(?:MIDI CC|MIDI Controller|Program Change|CC|Pitch Bend|Pitchbend"
```

`BreakBeforeBinaryOperators` indeed is a single setting for every binary operator, so `&&`, `||`, `+`, `<<` and the parts of a conditional cannot be treated differently.

jcsteh:
Hmm, I don't love the lambda thing. It makes the code unnecessarily verbose, especially if you ended up with a sequence of lambdas or something like that. For example, there's a case where we do options.insert instead of options.push_back, and having named lambdas separates the code for each action from its name.

Given that this is an open clang-format bug, I guess we can hope it might get fixed some day? I guess I need to decide whether I can live with the spaces until then. :)

LeonarddeR:
> especially if you ended up with a sequence of lambdas or something like that.

Fair point, I hadn't considered that.
That said, I find mixed tabs/spaces terribly disgusting.

I really don't understand at all why clang-format makes it so incredibly difficult for us to do what we want, while black, ruf, and rust fmt naturally have a style that aligns much more with common sense. I'd almost start to think that clang-format is just not the tool we want.

jcsteh:
I get the impression C++ coders are a lot more comfortable/happy with aligning using spaces for some reason and clang-format seems to be designed around that. The ability to disable that seems to be kinda bolted on or hacked in, or at the very least, an afterthought.

That said, I'm sorry I'm being so nitpicky about this. On one hand, I totally support automated formatting: it saves a lot of wasted time, tedium and pointless code review nits. On the other hand, it's really hard to watch some of the code you've carefully formatted over multiple years (albeit inconsistently in places) get uglier. So I do really appreciate the effort, but I'm having trouble coming to grips with the fact that even though this is "for the greater good", some of it actually looks like a downgrade to me.

PR #941, opened 2023-10-14
Translation framework fixes to allow for automatic code formatting

jcsteh:
This pull request includes changes to the translation framework necessary to make it possible for us to format the code base with clang-format (#940).

- [x] Support translation calls split over multiple lines.
- [x] Add an additional translation macro for marking strings in compile time data structures as translatable for makePot. This will be a no-op in the C++ code.
- [x] Replace translate firstString comment markers with this new macro.
- [x] Remove support for translate firstString.


jcsteh:
CC @leonardder. This isn't finished yet (see the checklist), but I thought you might be interested in this first batch of work.

leonardder:
Thanks for taking this. I think I like the proposed no-op approach, it's probably the most elegant solution.

leonardder (approved):
I think this is good to go. Thanks!

PR #943, opened 2023-10-17
Added feedback when toggling or momentarily switching to alt key map …

ScottChesworth:
…layers in Main section, updated ReadMe

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -320,6 +320,44 @@ Most of these are actions built into REAPER, but a few are useful actions from t
 - Tempo: Increase current project tempo 01 BPM: alt+=
 - SWS/BR: Options - Toggle "Playback position follows project timebase when changing tempo": Alt+Backspace
 
+#### Alternative Key Map Layers (requires Reaper 7.0 or newer)
+- Main action section: Set override to default
+- Main action section: Toggle override to recording
+- Main action section: Toggle override to alt-1
+- Main action section: Toggle override to alt-2
+- Main action section: Toggle override to alt-3
+- Main action section: Toggle override to alt-4
+- Main action section: Toggle override to alt-5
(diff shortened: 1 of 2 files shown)


Lo-lo78:
@ScottChesworth 
To complete the work there should also be feedback from this action:
"Main action section: Set override to default"
Thank you!

Lo-lo78:
@ScottChesworth 
I noticed that if you change layers there is no longer any feedback.
I was trying to make 16 default key maps with a few basic actions.
But I can only get feedback from the action:
"Main action section: Set override to default".
Only from the default layer is there feedback.
If you go from alt-1 to default there is no more feedback.

ScottChesworth:
Hmm, seems the way I tried to report this is inadequate and needed more testing before I merged it. Sorry about that.

@jcsteh, situation is that these actions handling key map switching are toggle actions that expose state when viewed in the Main and Main (alt recording) sections, but they're not toggles that expose state in any of the alt key maps. Moreover, no existing toggle actions are displaying state in any of those 16 alt sections. Ergo right now, none of the work you did to have consistent reporting of toggles/making those reports easy to refine works when those actions are triggered from an alt map.
Confusingly, toggle states are displayed in the actions list when viewing the Main (alt recording) map, but the feedback I added around switching earlier today still isn't spoken when running those actions from there, even though I've checked the IDs remain consistent.

I thought about trying to write a function to intercept the switching actions, but haven't started on it yet because those toggle states not being visible across the board seems like a way bigger issue for our current reporting. Is that correct, and if so, the change needs to come from Cockos right?

jcsteh:
I fixed all the things in 5f1e10f.

ScottChesworth:
Thanks for the rescue!

PR #948, opened 2023-10-18
When a container is focused in an FX chain list, pressing f6 now focuses the container's FX list.

### LINKED ISSUE #947: Is it possible to switch from FX chain to FX container chain with F6?
I find the workflow to be a bit slow.
Going from the FX chain to the FX container is a bit slow.
Is it possible to create an F6 key exception for container plug-ins that directly passes through the container's FX chain?
Thank you!

jcsteh:
Fixes #947.

### DIFF
--- a/src/fxChain.cpp
+++ b/src/fxChain.cpp
@@ -102,11 +102,23 @@ void shortenFxName(const char* name, ostringstream& s) {
 #ifdef _WIN32
 
 bool maybeSwitchToFxPluginWindow() {
-	HWND window = GetForegroundWindow();
 	MediaTrack* track;
 	MediaItem_Take* take;
 	int fx;
 	if (!getFocusedFx(&track, &take, &fx)) {
+		return false;
+	}
+	// Find the nearest ancestor FX chain parent window. This might be the top
+	// level FX chain or it might be a container. This allows f6 to focus FX
+	// inside a focused container.
+	HWND window = GetFocus();
+	do {
+		window = GetParent(window);
+		if (isClassName(window, "#32770")) {
+			break;
+		}
+	} while (window);
+	if (!window) {
 		return false;
 	}
 	// Descend. Observed as the first or as the last.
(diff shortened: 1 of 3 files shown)


jcsteh:
CC @Lo-lo78 for testing. Note that f6 won't work for FX inside containers at this point, including sub-containers. However, f6 should at least now allow you to access top level container chains.

ScottChesworth:
Is this cross platform or specific to Windows?

Lo-lo78:
@ScottChesworth 
> Guess I'm just worried about destabilizing the handy focus trick that's on F6 at the moment. Could Control+Tab either cycling through containers or placing focus on the list of containers be a thing?
I like the idea!
Go to the fx chain of the next fx container and then ctrl+tab again to go to the sub fx container.
With ctrl+shift+tab go back to the previous fx container and if possible you could return to the main fx chain where the first fx container is located.
This way we leave F6 as it always has been and You could add ctrl+tab and ctrl+shift+tab just for the fx container plugin.
The important thing is that it is not in conflict with Reaeq.

jcsteh:
This fy6 hack has always been Windows specific.

I really don't understand the desire to make this control+tab instead. This new f6 behaviour for containers will only happen when a container is focused in the FX list. If some other effect is focused, f6 will behave as it always did. There is no ambiguity here. Unless I'm missing something?

Lo-lo78:
@jcsteh 
Yes, to do a clean job, perhaps ctrl+tab could confuse the user who is used to conceiving ctrl+tab for tables on Windows.
I agree and the latest build works but I think we should make the most of Reaper's potential.
If it is possible to use the container and sub containers where the limit is the CPU, it is possible that in the future it will happen more and more often to work in this way.
I found myself very comfortable using the presets.
I prefer container FX presets over saving FX chain..
The FX container for plugin chains is much faster.

jcsteh:
Whether it's f6 or control+tab, sub-containers are going to need additional, separate work. The keystroke we choose doesn't change that.

PR #955, opened 2023-10-27
Make it possible to access effects in containers via sub-menus when using OSARA: View FX parameters for current track/take.

### LINKED ISSUE #952: View FX parameters for current track/take (depending on focus) (proposal new design)
Since the arrival of containers in Reaper, I think it would be good to think about another design of the menu "View FX parameters for current track/take (depending on focus)".
Jamie has already accessed and fixed all the bugs when we are directly in the FX window.
But, personally, I would use a keyboard shortcut other than F6 which would be dedicated to the container.
This is how I see it:
Shortcut in the FX window:
Shift+Right Arrow - If container is selected, moves focus in its FX list, otherwise states, no container is selected
Shift+Left Arrow - Moves focus to the previous container, otherwise states, you are not in a container.
Should this be a loop, maybe yes, maybe no, personally a loop would be fine for me but some users might find it a bit confusing.
So for these users a stopper would be better for them.
If this is feasible, navigation would be extremely fast in the FX window.
Otherwise for the rest of the work already done in the FX window is very good, it is very responsive.

Now for the menu with Ctrl+F when the track is selected in the TCP.
Here is an example project:
I have 2 containers on my track.
1. Named EQ, it contains an EQ and a compressor.
2. Named Delay, it contains a delay and a compressor.
If I select my track and open the OSARA FX parameters list, I have 2 elements, the 2 containers.
I don't have access to the plugin parameters in the containers directly in the menu.
So I need to open the FX window, select the container, press F6, select the plugin and press Ctrl+Shift+P.
So, I propose to create submenus that would have the same name as the containers in this menu And the plugins would be in the submenus.
If there is a container in a container, create a submenu in the submenu etc.
To open the list which contains the 3 parameters of the container (Bypass, Wet and Delta), we select it, right arrow to open it, and the first in the list would always be the container itself. So you just press Enter.
Otherwise, you move on to the first plugin in this container.
Obviously, if the container does not contain a plugin, just press Enter only, once it is selected.
With submenu that would have the same name as the containers, it would be really quick to locate the container with a down arrow or up arrow, and a right arrow swipe to locate the plugin in that container.
Please note, it is also possible to add a container in position 1, a VST in position 2 and another container in position 3 in the FX list of the FX window.
So in this case, The menu must therefore follow the tree structure of this list of FX.
The overhaul of this menu is essential for me because it no longer reflects the needs of the future.
Additionally, as mentioned by Scott, "It avoids lag while some GUIs are drawn when arrowing through the FX chain list."
It avoids an annoying UX blip where some JUCE plug-ins/instruments grab keyboard focus as soon as they get selected in the FX chain list.
It avoids breaking like, 10000 mentions of it across resources.
And I would add that we have to think for later, I think that users will overuse the containers so opening the FX window will take a very long time.
It's not uncommon for users to have huge lists of presets for each plugin, which slows down opening the window. So imagine what it will be like with 4 containers which contains 4 plugins.

jcsteh:
This was probably the trickiest thing I've had to code for OSARA in a while. It took me hours! The API for addressing container effects is truly awful. Nevertheless, I think I have it working.

@leejul, please test once a build is available.

Fixes #952.

### DIFF
--- a/src/fxChain.cpp
+++ b/src/fxChain.cpp
@@ -196,6 +196,7 @@ bool maybeReportFxChainBypass(bool aboutToToggle) {
 	}
 	outputMessage(enabled ? translate("active") : translate("bypassed"),
 		/* interrupt */ false);
+	if (fx >= 0x2000000) outputMessage(format("{}", fx - 0x2000000));
 	return true;
 }
 
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1364,7 +1364,7 @@ void fxParams_begin(ReaperObj* obj, const string& apiPrefix) {
 			// container itself.
 			itemInfo.fMask = MIIM_TYPE | MIIM_ID;
 			itemInfo.fType = MFT_STRING;
-			itemInfo.dwTypeData = (char*)translate("(container)");
+			itemInfo.dwTypeData = (char*)translate("(Container Parameters)");
 			itemInfo.cch = strlen(itemInfo.dwTypeData);
 			// We add 1 to wID because 0 means cancelled.
 			itemInfo.wID = fx + 1;
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1181,85 +1181,215 @@ void cmdParamsFocus(Command* command) {
 	new ParamsDialog(std::move(source));
 }
 
-typedef vector<pair<int, string>> FxList;
+// Iterates through effects, including effects in containers.
+template<typename ReaperObj>
+class FxIterator {
+	public:
+	FxIterator(ReaperObj obj): obj(obj) {
+		StackItem item;
+		// The first call to next() should move to the first effect, index 0.
+		item.indexInContainer = -1;
+		if constexpr(is_same_v<ReaperObj, MediaTrack*>) {
+			item.containerCount = TrackFX_GetCount(obj);
+		} else {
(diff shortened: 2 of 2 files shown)


ScottChesworth:
Working nice here on Windows. Thanks! I've just tweaked the topmost menu item for each container, aiming for a bit of extra clarity is all.

ScottChesworth:
Yeah I chewed over the same thought, think this is clearer though.

jcsteh:
Could someone take this for a spin on Mac? OSARA hasn't done sub-menus before, so I just want to make sure it behaves correctly.

leejul:
Wow, good job Jamie, I couldn't ask for more.
The presentation is very clear.
It's good to add the word Parameters for the one at the top of the list who is the container. This will be very clear to the user.
Big thanks.

pitermach:
Looks good on Mac here. I made a track which had both a container with 2 FX and one more effect which wasn't in a container, and I was able to access all of them in the parameter menu.

mattymc2:
This is fantastic for both platforms. You’re a wizard. Sent from my iPhoneOn 27 Oct 2023, at 11:51 pm, James Teh ***@***.***> wrote:﻿
Merged #955 into master.

ElCrackLoko:
Hello guys, what function does this have? How can I test it?

PR #967, opened 2023-11-21
very Experimental support for fixed lanes

RDMurray:
This is just a first attempt at supporting the new fixed lanes feature in Reaper 7. It works similar to envelopes in that after focusing a track you can use OSARA: next/previous lane actions to cycle through the fixed lanes on the track. this restricts item navigation to the items on the selected lane.

For better support we would need Cockos to make some additions to the API:

* to get/set the name of a lane
* to tell if a lane is a comp lane and if comping is enabled
* ability to show lane context menus
* In general it would just be nice if lanes were better supported by reaper actions so that they can be selected, pasted on to etc.

From what I can tell a comp area is just an item on a comp lane. After selecting one the actions to move comp area up/down work.

This is just a draft as I am sure there are bugs and it is far from complete.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -517,8 +517,19 @@ bool isPosInItem(double pos, MediaItem* item) {
 	return (start <= pos && pos <= end);
 }
 
-bool isFreeItemPositioningEnabled(MediaTrack* track) {
-	return *(bool*)GetSetMediaTrackInfo(track, "B_FREEMODE", nullptr);
+enum class FreeMode {
+	none,
+	free,
+	fixed
+};
+
+FreeMode getTrackFreeMode(MediaTrack* track) {
+	int* freeMode = (int*)GetSetMediaTrackInfo(track, "I_FREEMODE", nullptr);
+	if(freeMode) {
+		return static_cast<FreeMode>(*freeMode);
+	} else {// Reaper before version 7
+	return *(bool*)GetSetMediaTrackInfo(track, "B_FREEMODE", nullptr) ? FreeMode::free : FreeMode::none;
+	}
 }
 
 const char* automationModeAsString(int mode) {
@@ -728,12 +739,14 @@ const char* (*NF_GetSWSTrackNotes)(MediaTrack* track) = nullptr;
  */
 
 bool shouldMoveToAutoItem = false;
+int trackFixedLane = -1;
 
 void postGoToTrack(int command, MediaTrack* track) {
 	fakeFocus = FOCUS_TRACK;
 	selectedEnvelopeIsTake = false;
 	shouldMoveToAutoItem = false;
 	SetCursorContext(0, nullptr);
+	trackFixedLane = -1;
 	if (!track)
 		return;
 	ostringstream s;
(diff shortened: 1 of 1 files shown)


jcsteh:
Discussed this on Zoom. A few things to do here:

1. Get this rebased. Sorry @RDMurray. :)
2. When we switch lanes, it probably makes sense to change the actively playing lane accordingly. Not quite sure how to do this, but maybe C_LANEPLAYS on the item does it?
3. New actions and API things have been added since this PR. We should take a look at those and figure out whether this gives us stuff we need.
4. We could potentially get someone with sight to help testing things, but we should come up with a bunch of API queries we want to make after each step so we know how each step is reflected in the API.

RDMurray:
I added reporting of lane name for osara: next/previous fixed item lane.

I did some experimentation, some random observations:

* It seems there's a bit more to comp areas than just items on the comp lane, and unfortunately we don't have access to them through the API or actions as far as I can tell.

Say you record a long item, then punch in a shorter item in the middle, creating 3 comp areas. The comp lane contains three items. If you move the middle comp area up, switching it to the first recording the comp lane now only has one item, but there are still three comp areas.

They are stored in the RPP as `LINKEDLANE` lines. 

From my test project:
```
    ITEMLANES 3
    LINKEDLANE 0 0.8821875 2 0 -1 0.01 0.01
    LINKEDLANE 0.8721875 3.696375 1 0 -1 0.01 0.01
    LINKEDLANE 3.686375 5.5441875 2 0 -1 0.01 0.01
```

The first two fields are start and end time and the 3rd is lane number.

* It is possible to get the take style behaviour when recording if you run "Track lanes: Comp into new empty lane" first. It makes a new comp area each time you record.

* I don't know if reaper has a concept of selected comp areas. In my experimenting "Fixed lane comp area: Move down" and "Fixed lane comp area: Delete comp area" have acted on all comp areas.
* "Fixed lane comp area: Split comp area at edit cursor" works and also splits the corresponding item on the comp lane.

ScottChesworth:
Bumping this as Cockos are keen to remove take comps in favour of lanes. Is there enough info available via the API and suitable actions to get those going? @RDMurray, what do you reckon we need nowadays?

RDMurray:
I just looked through the API docs from 6.47. They have added lane names to GetSetMediaTrackInfo_String:

P_LANENAME:n : char * : lane name (returns NULL for non-fixed-lane-tracks)

I don't see anything to identify a comp lane. I'll play with it again and look at the rpp files to try to figure out how comp lanes are identified.

We would also have to carefully consider the user interface, because unlike takes, items in lanes can be positioned independently of each other and I think you can for example change takes in the middle of an item without splitting. I am guessing that this results in the comp lane containing split items that are copies of those in the source lanes.

ScottChesworth:
I've asked Cockos whether that's possible.

RDMurray:
@ScottChesworth  are they definitely intending to remove takes entirely, or just the "little used old-style named take comp feature" removed in the most recent dev build?

The main problem with lanes and comp areas is the lack of actions to select them. Most actions operate on comp area or lane under mouse.

PR #970, opened 2023-12-11
Add OSARA: Set phase normal for all tracks action to replace a custom…

ScottChesworth:
… action, this reports nicer

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4210,17 +4210,16 @@ void cmdSelectFromCursorToStartOfProject(Command* cmd) {
 	outputMessage(translate("selected to start of project"));
 }
 
-void cmdSetPhaseNormalAllTracks(Command* cmd) {
+void cmdSetPhaseNormalAllTracks(Command* command) {
 	int count = CountTracks(nullptr);
 	if (count == 0) {
 		outputMessage(translate("no tracks"));
 		return;
 	}
-	int phaseVal = 0;
 	for (int t = 0; t < count; ++t) {
 		MediaTrack* track = GetTrack(nullptr, t);
 		if (isTrackPhaseInverted(track)) {
-			GetSetMediaTrackInfo(track, "B_PHASE", &phaseVal);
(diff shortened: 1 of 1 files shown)


src/reaper_osara.cpp
```diff
@@ -4210,6 +4210,22 @@ void cmdSelectFromCursorToStartOfProject(Command* cmd) {
 	outputMessage(translate("selected to start of project"));
 }
 
+void cmdSetPhaseNormalAllTracks(Command* cmd) {
```
jcsteh, line 4213: cmd -> command for consistency. I messed this up in the last action I added, so this is my fault, not yours. :)

```suggestion
void cmdSetPhaseNormalAllTracks(Command* command) {
```

src/reaper_osara.cpp
```diff
@@ -4210,6 +4210,22 @@ void cmdSelectFromCursorToStartOfProject(Command* cmd) {
...
+	int count = CountTracks(nullptr);
+	if (count == 0) {
+		outputMessage(translate("no tracks"));
+		return;
+	}
+	int phaseVal = 0;
```
jcsteh, line 4219: 1. This will work, but B_PHASE is actually a bool, not an int, so we should use a bool.
2. As it happens, OSARA already defines a bool called bFalse for this purpose, so we can just use that.

```suggestion
```

src/reaper_osara.cpp
```diff
@@ -4210,6 +4210,22 @@ void cmdSelectFromCursorToStartOfProject(Command* cmd) {
...
+	for (int t = 0; t < count; ++t) {
+		MediaTrack* track = GetTrack(nullptr, t);
+		if (isTrackPhaseInverted(track)) {
+			GetSetMediaTrackInfo(track, "B_PHASE", &phaseVal);
```
jcsteh, line 4223:
```suggestion
			GetSetMediaTrackInfo(track, "B_PHASE", &bFalse);
```

jcsteh (approved):
Thanks.

PR #976, opened 2023-12-13
Added reporting for Item properties: Toggle lock. Closes #944

### LINKED ISSUE #944: Speech feedback for the action, Item properties: Toggle lock.
I see Osara reports an item's locked state when moving to an item. Would it be possible to get speech feedback when toggling locking on the current item? Idealy, this would also work for multiple selected items, for example saying "3 items locked."

ScottChesworth:


jcsteh:
Hi hi. It looks like your tinygettext submodule is out of date. Git doesn't update those automatically. You'll want to run git submodule update so that doesn't accidentally get committed in future. Thanks!

ScottChesworth:
Sorry about that. Did git submodule update, but am still seeing this message about tinygettext when I do git status:
> On branch master
Your branch is up to date with 'upstream/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   include/tinygettext (untracked content)

no changes added to commit (use "git add" and/or "git commit -a")
> 

When I tried either git add or git commit as suggested there, it griped that my master was out of whack with upstream, so I've amended that for the time being.
Any ideas?

jcsteh:
Hmm. I would run git submodule update again. Then cd include\tinygettext, git status, and see what is untracked there. My guess we can just get rid of the untracked files, but I guess we'd better check.

ScottChesworth:
> Doing that gives me:
HEAD detached at 552dbbe
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        external/
> 

Figured I might learn something more by doing cd external and git status again, which says "./" is untracked, so cd ./, git status again, nothing else changes and that's when I start to suspect I don't know my anus from my elbow. :)

Thoughts?

jcsteh:
Weird. Is anything in the external/ directory? I guess you could try just removing the whole directory, as we don't need it.

PR #991, opened 2024-02-03
Add OSARA: Go to master track. Make OSARA: Go to first/last track scroll the track into view.

jcsteh:
Go to master track makes the master track visible if it isn't already. It also scrolls to the first track before selecting the master in the hope that this will also scroll the master into view.


jcsteh:
Can someone confirm whether the go to master track action scrolls the master track into view if it was previously scrolled off? For example, if you're on track 30 or something like that and you use the OSARA: Go to master track action, is the master track now visible?

ranetto:
Hi, I can't find the action go to the master track in the last osara that I just downloaded, however if I go to track number 100 with the osara action go to the last track, I continue to see the master until I I move the arrow up or down and select the last track. If I go to the first track I don't see the master until I move with the arrows selecting the first track. Can you explain to me why I don't see the action you mentioned? Thank you!

jcsteh:
Sorry, this change isn't in the main OSARA snapshots yet. You'll need to test with [this test build](https://ci.appveyor.com/api/buildjobs/hqrtjy468vh2rnhi/artifacts/installer/osara_pr991-1276,b79cba4a.exe). (For future reference, in pull requests, you'll see AppVeyorBot add a comment about the build succeeding. That comment includes links to the test build for the change.)

ScottChesworth:
Had eyes check here, master gets made visible, scrolled into view placed near the top of the screen and it gets selected. Jen and I will get key map adjustments added to this branch tomorrow and then it's good to go.

PR #992, opened 2024-02-04
Include track, item, take and FX info in the title of OSARA Parameters dialogs.

### LINKED ISSUE #743: Is it possible to put the name of the track in the title of the OSARA FX Parameters?
Hello,
after so many hours in the mixing session I happened to forget the name of the track while I was setting up an FX on Reaper.
I tried with NVDA + t but it was just telling me FX Parameters.
One might think for example:
"FX Parameters: Track 1 "piano"
Maybe I'm stoned but I think it can come in handy.
Greetings!

jcsteh:
Fixes #743.

@Lo-lo78, is this what you're after?
@ScottChesworth, are you happy with this?

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -754,9 +754,7 @@ class FxParams: public ParamSource {
 		}
 	}
 
-	string getTitle() {
-		return translate("FX Parameters");
-	}
+	string getTitle();
 
 	int getParamCount() {
 		// Any named config params come first, followed by normal params.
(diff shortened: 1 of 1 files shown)


ScottChesworth:
Can FX name be stripped here? Would it be possible to do stripped FX name on track/take?
So instead of what we have now:
FX Parameters: 1 Bass: VST: ReaEQ (Cockos)
We'd have:
FX Parameters: ReaEQ on 1 Bass
I'd like to report the name of the effect sooner and move track/take name last in the report, because take names are verbose time stamps by default.

jcsteh:
I'm not a big fan of the "effect on track" notation because it's not ideal for translations. I guess I can make it "{fx} on {object}" or something, but it's far easier if we can just avoid trying to make it a phrase altogether.

What about "FX Parameters: ReaEQ (1 Bass)"?

jcsteh:
Or "FX Parameters: ReaEQ, 1 Bass"

ScottChesworth:
The second one I reckon, takes slightly less space on a braille display and can be said a bit faster if speech verbosity is high.

Lo-lo78:
@jcsteh 
> Or "FX Parameters: ReaEQ, 1 Bass"

👍
As for the braille display let's hear from the others.
Thank you!

Lo-lo78:
@jcsteh 
@ScottChesworth 
👍 
For me it's OK! Let's hear from the community!

PR #998, opened 2024-02-18
Optionally configure REAPER for optimal screen reader accessibility.

### LINKED ISSUE #343: Consider undocking the Media Explorer by default
By default, when you open the Media Explorer, it gets docked and does not get focus. This makes it pretty hard to access. When you're done with it, dismissing it when docked is also not intuitive or trivial. It might be worth OSARA disabling this automatically.

I'm generally pretty reluctant to mess with user settings without asking, as a user might have configured things that way for a reason, might be working with a sighted user, etc. It's also just rude. In this case, though, I wonder whether the difficulty of accessing this with default settings justifies this.

Alternatively, we could have OSARA ask the user whether to make some recommended setting changes at startup and give them the option. That does add complexity, though.

Technical notes:

1. There's an action in the Media Explorer section to dock the Media Explorer. It doesn't seem to have a toggle state unless the Media Explorer is open. I'm also not sure whether SetToggleCommandState will work on this.
2. This is controlled by the docked key in the reaper_explorer section of reaper.ini. Setting to 0 undocks. However, I don't know whether REAPER will reflect this setting automatically if it's modified at runtime.

jcsteh:
The user will be prompted once at startup to do this. Alternatively, they can run the action OSARA: Configure REAPER for optimal screen reader accessibility at any time. This currently does the following:

- Undocks the Media Explorer by default so that it gets focus when opened.
- Enables legacy file browse dialogs so that REAPER specific options in the Open and Save As dialogs can be reached with the tab key.
- Enables the space key to be used for check boxes, etc. in various windows.
- Shows text to indicate parallel, offline and bypassed in the FX list.

Fixes #343.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -883,6 +883,7 @@ This list is worth referencing when making your own key map additions, assigning
 - OSARA: Toggle Report time movement during playback/recording
 - OSARA: Toggle Report track numbers
 - OSARA: Toggle Report transport state (play, record, etc.)
+- OSARA: Configure REAPER for optimal screen reader accessibility
 
 ### Muting OSARA Messages in Custom/Cycle Actions
 The action "OSARA: Mute next message from OSARA" can be used in custom/cycle actions to mute OSARA feedback for the next action.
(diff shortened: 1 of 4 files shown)


ScottChesworth:
Awesome to see this taking shape! Hang fire for a day or two before merging, gonna put a note out on RWP to see whether anyone suggests things we hadn't considered yet.

PR #1002, opened 2024-02-26
Include win32_utf8 in config.cpp to fix Unicode characters in the REAPER optimal configuration dialogs.

jcsteh:


jcsteh:
This problem was reported on RWP. I didn't get a reply from the reporter, but I'll merge it anyway, since it's definitely more "correct" even if it doesn't fix the bug.

PR #1006, opened 2024-03-02
Include source/destination audio channel for sends in the Track Parameters dialog.

### LINKED ISSUE #1003: Can we include channel choices for sends/receives in the OSARA track parameters dialog?
We've had a bunch of requests over the years to simplify the routing around sends and receives, the main thing people seem to want to avoid is repetitive diving in and out of menus. If we can improve this in some way, ideally I'd like to move away from the context menu on the Delete button being a necessary step while we're at it, have seen a lot of learners forget that's what to do.
Instead, could we add channel configuration choices to the track params dialog? Looks like it might be possible via the API but I don't think I'm smart enough to code it yet.
Thoughts?

jcsteh:
Fixes #1003.
I really hope this was worth it. It took me more than 4 hours to implement. This multi-channel stuff really hurts. :)
I'd appreciate some testing with various multi-channel scenarios before i merge this.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1044,9 +1044,13 @@ class TrackSendParamProvider: public ReaperObjParamProvider {
 		ReaperObjParamProvider(displayName, name, makeParamFromProvider),
 		track(track), category(category), index(index) {}
 
-	virtual void* getSetValue(void* newValue) override {
+	void* getSetValue(const char* name, void* newValue) {
 		return GetSetTrackSendInfo(this->track, this->category, this->index,
-			name.c_str(), newValue);
+			name, newValue);
+	}
+
+	virtual void* getSetValue(void* newValue) override {
+		return this->getSetValue(name.c_str(), newValue);
 	}
 
 	private:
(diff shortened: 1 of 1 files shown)


Lo-lo78:
@jcsteh 
Don't kill me for this!(rofl)
But is creating a "(New channels on receiving track)" button with a shortcut "alt+n" in the Track Parameters window a problem?
It is true that this can already be done with Reaper in the "Routing for track" dialog.
But if it were in the OSARA Track Parameters window the session is much faster.
In Italian there is an expression: la butto la!
it means I try to make a proposal and then if it's possible it's great if it's not possible we'll move on to other adventures.

jcsteh:
The dialog is designed to be pretty generic for all kinds of parameters, so it's not really feasible to have buttons specific to particular parameters. For one thing, the button would have to appear or disappear when those parameters are selected or unselected, which is a slightly weird UX. It's not just like the text box which gets enabled or disabled: it would have to appear or disappear, since there would be different buttons for different parameters.

A context menu on the slider might be better, but I'm not sure if I can make that work on Mac. I guess we could have an Additional actions button or similar which opens a menu.

In any case, I think we should deal with anything like that separately.

PR #1017, opened 2024-03-10
Provide the ability to add new channels for the source and destination audio channel parameters in the OSARA Track Parameters dialog.

### LINKED ISSUE #1010: A button or applications key for the source/destination audio channel for sends in the Track 3/4 Parameters dialog and etc...
https://github.com/jcsteh/osara/pull/1006#issuecomment-1974928950

>  For one thing, the button would have to appear or disappear when those parameters are selected or unselected, which is a slightly weird UX. It's not just like the text box which gets enabled or disabled: it would have to appear or disappear, since there would be different buttons for different parameters.
> 
> A context menu on the slider might be better, but I'm not sure if I can make that work on Mac. I guess we could have an Additional actions button or similar which opens a menu.
> 
> In any case, I think we should deal with anything like that separately.

I don't know if it can be done.
You might think about creating an Options button that we could improve over time by adding functionality.
This way the Options button is always active and if nothing is there nothing happens.
In the future we could add what might be needed step by step.
An open door for the future.

jcsteh:
This can be accessed via the new context menu on the parameter value slider.

Fixes #1010.

I don't know if the context menu can be accessed on the Mac with VO+shift+m. If it can't, we'll probably need to add a More button or similar which can access it. Could someone test this?


jennykbrennan:
No joy I'm afraid. VO Shift M gives me nothing but a polite little noise telling me, nothing to see here.

jcsteh:
I hate supporting the Mac. I really do.

jcsteh:
I've added a "More" button to access the same menu. Hopefully that works on Mac. Builds should be up soon.

Two questions now:

1. Are we happy with the name "More"? I considered "Options", but every parameter value is technically an option. :)
2. Do we get rid of the ability to bring up this context menu with the applications key or shift+f10 on Windows, since it apparently doesn't work on Mac (even though I tried to implement it in a cross-platform way)? This would mean you would use the More button on both platforms, rather than having two options (More button or apps key) on Windows.

jcsteh:
You also don't have accelerators on Mac as far as I know, so you'll have to tab to the more button rather than just hitting alt+m, but there's only so much I can do to compensate for a platform with inherently crappy keyboard support.

jennykbrennan:
I don't have any strong opinion either way. I'll leave that to those with opinions. I'm so used to doing things just slightly, and awkwardly differently on the mac that one more thing wouldn't bother me. Although yeah, having even more differences to learn isn't very appealing. You hate supporting the mac, I hate that I'm still using it. There has to be a support group for that.

jcsteh:
Just to be clear, i really, really want to love Mac. The hardware is better, the OS is better in many ways, it's far better for audio for sure. But the bolted on keyboard support and the ridiculous bugs/quirks in VoiceOver and just...

Now this is going way off topic. I'm going to bed. Thanks for your help. :)

jcsteh:
(As a side note for anyone else watching, the context menu not working is probably a problem in SWELL; it's certainly possible to get VO+shift+m working with native Mac widgets. Part of the problem here is that we're dealing with Mac + SWELL weirdness.)

ScottChesworth:
I reckon keep both approaches in, might be able to get Justin to take a look at the SWELL weirdness.

jennykbrennan:
Oh, there's much to like, unfortunately. Back on topic, this build.  It works well. One more button didn't end up causing personal injury or mental collapse. All good. :)

pitermach:
For the record, I seriously appreciate all of the what must be incredibly frustrating work you're putting in to make OSARA work on Mac and personally find the Windows audio stack and work culture/heat management of even modern Windows systems much worse so I'm happy I can use Reaper on a Mac. Anyway, getting back on-topic I came here to say from my experiences with helping to make plugins accessible, unless you're on the web, VO+Shift+M doesn't simulate a right click and instead calls some function that developers have to implement. This seems to happen automatically for native applications, but when someone implements accessibility from scratch developers have to support it themselves (IE Juce has an API to abstract this Show Menu action). In Reaper this works in most places, including on many toolbar buttons which I just tried, but there is at least one place where it doesn't (the paths and databases tree of the Media Explorer)

jcsteh:
Yeah, I'm familiar with that API (accessibilityPerformShowMenu) because of Mozilla's work on Mac accessibility in Firefox. In this case, though, I'm fairly sure REAPER creates native widgets for most things on Mac, rather than custom widgets. (SWELL maps Win32 UI calls to native Cocoa calls.) So, I don't understand how to tell a native Mac button that it has a context menu. Maybe you just have to implement accessibilityPerformShowMenu even for native widgets, in which case there's nothing I can do because I don't have access to the actual Mac widget.

This working correctly on the toolbar makes sense because it's one of the few places where REAPER uses custom widgets. I just checked the WDL code and it does indeed implement accessibilityPerformShowMenu for those widgets.

SWELL does implement accessibilityPerformShowMenu for list view controls, so there's already some precedent. However, there's probably no way for it to know whether an emulated Win32 widget supports a context menu or not, since there's no way to declare that you have a context menu other than responding to the event when asked to open one.

PR #1018, opened 2024-03-10
Disable the Windows specific OSARA hack to report active/bypass in the FX chain list in REAPER 7.06 and later, since REAPER now makes this accessible properly.

### LINKED ISSUE #1016: Is it possible to update the old OSARA FX Bypassed/active functionality?
Hello with the new versions of Reaper we now have feedback of the status of the plugin in the FX chain.
Now if you bypass a plugin it will be repeated twice.
I believe one from OSARA and the second from Reaper.
In this case I would give precedence to Reaper.
When it is activated I think it says OSARA.
In this case we can decide if it is good or not good.
when it is active it is reported by OSARA and not by Reaper. We might as well leave it.
In my opinion, a choice needs to be made here.
There is also a second problem.
When you open the menus there is a bit of confusion:
Add container  3 di 39. bypassed 
And this is even stranger:
Run selected FX in parallel with previous FX,  activated3 di 39. active
In my opinion it needs to be fixed up a bit. 🙂

jcsteh:
Fixes #1016.
I haven't actually tested this on REAPER 7.05 or earlier to check whether the hack still works there. If someone gets a chance before I do, that'd be awesome.

### DIFF
--- a/src/fxChain.cpp
+++ b/src/fxChain.cpp
@@ -176,6 +176,11 @@ bool maybeSwitchToFxPluginWindow() {
 // If the FX list in an FX chain dialog is focused, report active/bypassed for
 // the selected effect.
 bool maybeReportFxChainBypass(bool aboutToToggle) {
+	string version = string(GetAppVersion(), 0, 4);
+	if (stod(version) >= 7.06) {
+		// REAPER >= 7.06 made this properly accessible.
+		return false;
+	}
 	if (!isFxListFocused()) {
 		return false;
 	}
--- a/src/osara.h
+++ b/src/osara.h
@@ -212,6 +212,7 @@
 #define REAPERAPI_WANT_TakeFX_GetNamedConfigParm
 #define REAPERAPI_WANT_GetTrackSendName
 #define REAPERAPI_WANT_AddExtensionsMainMenu
+#define REAPERAPI_WANT_GetAppVersion
 #include <reaper/reaper_plugin.h>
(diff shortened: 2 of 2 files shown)


Lo-lo78:
@jcsteh 
I can give it a try immediately.
I'll give you feedback in 10 minutes.

Lo-lo78:
@jcsteh 
I confirm!
With the reaper705 version there are only OSARA feedback.
From version 706 everything is ok.

Lo-lo78:
It works but unfortunately we are only at Reaper 7.11.
Maybe we'll have to wait a bit.
Or alternatively you could leave the latest version compatible with 7.5 on the OSARA snapshots page. ;)

jcsteh:
Can i clarify that this worked as expected for you in 7.05; i.e. you got the OSARA feedback? In contrast, in 7.06 and alter (including 7.11), you don't get the OSARA feedback, but you instead get the REAPER feedback. In that case, we can merge this, since users of older versions are still okay, which is exactly what this is intended to do.

Thanks for your help.

Lo-lo78:
Exactly!
Just as you asked!

Lo-lo78:
The test is as follows:
1. install reaper portable 7.05
2. I install the version of OSARA pr1018-1347,d5047150
3. I try any plugin
4. Result: Works only with OSARA
5. install reaper portable 7.06
6. I install the version of OSARA pr1018-1347,d5047150
7. I try any plugin
8. Result: Only works with Reaper
It seems ok to me!

PR #1021, opened 2024-03-13
Track Parameters dialog: For send/receive parameters, add an option in the context menu to go to the send/receive track.

### LINKED ISSUE #569: Is it possible to add a button in the track parameters that opens a menu with all the sends and receives of the selected track and if you select a send or receive track it takes you into the TCP in the right track?
Hello,
while mixing I feel the need to switch to the FX Chain of the track send or the Receive. In this way it would be possible to pass from the source track to the receiving track, from the receiving track to the source; to get to the desired FX Chain list faster.

For instance:

Parameter:  Combo box  2 Reverb send volume  
Value:  progress cursor  0.00dB
0.00dB
FX Chain: 2 Reverb  Button  ALT + p
Filter:  ALT + f
Close  Button  ALT + c

It should also include the Pan, Mute and Mono of all sends or Receives.

In my opinion this improves the work a lot and makes it faster.
Greetings!

jcsteh:
Fixes #569.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1126,7 +1126,8 @@ class TrackSendParamProvider: public ReaperObjParamProvider {
 				translate("Go to send destination track"),
 				[this] { return this->goToTargetTrack("P_DESTTRACK"); }
 			}};
-		} else if (this->category == -1) {
+		}
+		if (this->category == -1) {
 			return {{
 				translate("Go to receive source track"),
 				[this] { return this->goToTargetTrack("P_SRCTRACK"); }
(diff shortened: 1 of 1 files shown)


Lo-lo78:
It works very well.
It would be faster if there was the possibility of using the application button among the parameters.
Or even press enter to go to the track.
Or even a button that if pressed takes you into the track.
With the benefit of an alt+ accelerator.

jcsteh:
A dedicated button isn't realistic for reasons given previously; it would only apply to some parameters and it's silly to have a go to track button always visible (even if disabled) for parameters that don't support it.

Enter doesn't make sense either, since many people probably dismiss the dialog with enter. We don't want people accidentally losing their place in the track list by doing this.

You can already do this pretty quickly by opening the context menu, then pressing g. Note that you can open the context menu by pressing alt+m, or by moving to the value slider and pressing the applications key.

Lo-lo78:
Yes, you are actually right about the possibility of accidentally finding yourself on the wrong track.
I just tried it and it actually works well.
For me it's OK!
Let's see what others say.

PR #1025, opened 2024-03-13
Parameters dialog: Allow the parameter context menu to be opened using the applications key on the Parameter combo box.

jcsteh:


Lo-lo78:
👍

Lo-lo78:
Thank you!
You are fantastic!

ranetto:
Hi everyone, I would like it possible to add a button to the contextual menu to delete the send in question, I'll explain, when I have many sends to a track, for example when I create a return bus of a reverb, I would like to be able to delete all the send tracks to those buses, in the reaper routing window you have to navigate a lot to get to all the buses to be eliminated, as each time the focus shifts as we have eliminated one of the buses, what do you think about adding an entry to the context menu to delete the bus we are positioned on? Thanks in advance!

jcsteh:
You can already do this effectively using the routing context menu (alt+applications in the track view).

ranetto:
I CERTAINLY CAN DO IT, BUT THAT IMPLIES HAVING TO SEARCH ALL THE TRACKS AND ALSO THE FOLDERS, WHILE IN THIS CASE SIMPLY HAVING THE POSSIBILITY OF DELETING A SEND IN THE CONTEXTUAL IT WOULD ONLY REFER TO THE TRACK ON WHICH I AM WORKING

PR #1026, opened 2024-03-13
Fix issue with reporting the length of a time selection when there are tempo/time signature changes. 

### LINKED ISSUE #839: When reporting length of time selection in bars/beats, OSARA doesn't factor in tempo/time signature changes
Similarly to #249, Osara doesn't report the correct time selection length in bars/beats when there are tempo/time signature changes in the project. This issue is medium priority at most, but fixing it at some point would be nice.

RDMurray:
fixes #839

refactor FormatTime and add formatLength.


RDMurray:
Oops, pushed with merge conflicts by mistake.

jcsteh:
Sorry. That was probably my cleanup patch. :(

RDMurray:
My fault, I only noticed the first conflict and didn't build again before commiting.

RDMurray:
There's a possible problem with one call that I changed to formatLength in `ReaperObjLenParam::getValueText:

```
	string getValueText(double value) final {
		static string lastText;
		string text = formatLength(0, value);
		if (text.empty()) {
			// formatTime returned nothing because value produced the same value text as the last call.
			// Therefore, we cache the text and return it here.
			return lastText;
		}
		lastText = text;
		return text;
	}
```

I don't really understand the need for caching. `formatLength` never returns an empty string, and afaik `formatTime` with `FT_NO_CACHE` doesn't either. Is this from before `FT_NO_CACHE` was a thing? The comment was committed 8 years ago. I take it the caching in this method can safely be removed?

There's also an unrelated issue with ReaperObjLenParam. If the ruler is bars.beats the edit field in the item params dialog is 1.1.0 instead of 0.0.0. It's because of the use of `format_timestr_pos`.

there are `format_timestr_len` and `parse_timestr_len` but they need an offset which would be a problem for fadeout time because the offset would be unknown when parsing.

I wonder if these times should just be in seconds regardless of the ruler?

jcsteh:
> I don't really understand the need for caching. formatLength never returns an empty string, and afaik formatTime with FT_NO_CACHE doesn't either.

But ReaperObjLenParam calls formatTime without a cache parameter, so it defaults to FT_CACHE_DEFAULT.

> Is this from before FT_NO_CACHE was a thing? The comment was committed 8 years ago.

No. FT_NO_CACHE existed back then, albeit in a slightly different form (useCache=false).

> I take it the caching in this method can safely be removed?

It can't. The easiest way to understand this is as follows:

1. Create a blank project.
2. Press shift+insert (twice if you're using NVDA) to create a MIDI item of 1 bar in length.
3. Control+leftArrow to move to that item.
4. Alt+p.
5. Down arrow to fade in length.
6. Page up to increase the value to 5 bars.
7. Right arrow to increase the value to 5 bars 4%.

Notice that OSARA just says "4%", not "5 bars 4%". While this is arguably a bit weird, it makes adjusting length a lot easier, since you don't have to continually here, for example, "5 bars 0%", "5 bars 4%", etc. So, that's why we use caching here.

jcsteh:
> There's also an unrelated issue with ReaperObjLenParam. If the ruler is bars.beats the edit field in the item params dialog is 1.1.0 instead of 0.0.0. It's because of the use of format_timestr_pos.

Oof. Nice catch... and that's ugly.

> there are format_timestr_len and parse_timestr_len but they need an offset which would be a problem for fadeout time because the offset would be unknown when parsing.

Do we not have the same inaccuracy with formatLength in this case though, since we give it an offset of 0? We could give it the item end position if that would help.

> I wonder if these times should just be in seconds regardless of the ruler?

We could do that. I believe that's what Item Properties does. That said, I think there's maybe some utility in being able to see this in the ruler unit, but I'm not sure. Maybe @ScottChesworth has thoughts on that.

ScottChesworth:
> there's maybe some utility in being able to see this in the ruler unit, but I'm not sure. Maybe @ScottChesworth has thoughts on that.
> 
I'd say there is yeah, if you're thinking in terms of bars and beats for your arrangement stuff then it's easier to decide fade lengths in the same unit. IE, I know I want the entire last chorus to be a fade-out and I also know the length of that chorus, done.

jcsteh (changes requested):
Thanks for working on this. Requesting changes because formatLength needs to be able to cache as explained in https://github.com/jcsteh/osara/pull/1026#issuecomment-1999275732.

RDMurray:
This should work in the item params dialog now. I also rewrote formatLength to match reaper's behaviour. There may still be edge cases where it doesn't work.
To test I do the following:
- make a time selection and read it's length with osara
- insert midi item and check that length in item properties matches

I noticed that when zoomed in far enough, the item start isn't always exactly at the selection start.

jcsteh:
Thanks. You did the item length stuff a lot more cleanly than I envisaged. I was thinking we could just compare the name string and do everything in the ItemLenParam class, rather than having additional providers. That's less code, but your solution is definitely cleaner.

There are still some regressions with those parameters though:

1. Open a new project.
2. Press shift+insert (twice if you're using NVDA) to insert an empty MIDI item, then control+leftArrow to select it.
3. Press alt+p.
4. Press f to move to the fade in length.
5. Press right arrow. Observe: "4%"
6. Press left arrow.
    - Expected: "0%"
    - Actual: Nothing
7. Press home. Observe: "250 bars"
8. Press left arrow. Observe: "249 bars 3 beats 96%"
9. Press right arrow.
    - Expected: "250 bars 0 beats 0%"
    - Actual: "250 bars"
10. Press end.
    - Expected: "0 bars"
    - Actual: Nothing

jcsteh:
Aha. The issue is that formatLength is calling formatTimeMeasure with includeZeros set to false. Is there a reason you're setting it to false there rather than true?

jcsteh:
Hmm, I guess I can see cases where you might not want zeros. The whole reason includeZeros was introduced was to deal with MIDI note lengths, where it's not useful to report zeros, since you're not dealing with deltas. Maybe we could set includeZeros to true only when we're using the cache?

That said, I don't know if anywhere else ever sets this to false now that formatNoteLength has been split out. I wonder if we should just remove includeZeros; i.e. so it's permanently true.

RDMurray:
> Aha. The issue is that formatLength is calling formatTimeMeasure with includeZeros set to false. Is there a reason you're setting it to false there rather than true?

No, I just didn't think it through.

> That said, I don't know if anywhere else ever sets this to false now that formatNoteLength has been split out. I wonder if we should just remove includeZeros; i.e. so it's permanently true.

formatNoteLength calls formatTimeMeasure with includeZeros set to false.

Thinking about it, we might actually need to add an includeZeros parameter to formatLength because it should be false when reporting the time selection length.

jcsteh (approved):

PR #1034, opened 2024-03-19
Track Parameters dialog: For send/receive/hardware output parameters, add an option in the context menu to delete the send/receive/hardware output.

jcsteh:
Because removing a send obviously removes its parameters, this involved teaching the Parameters UI code how to fully rebuild the parameter list due to a parameter being invalidated.

CC @ranetto.


ranetto:
Thanks and sorry for the work of rebuilding the parameter list

PR #1055, opened 2024-04-18
Added threshold to the information reported in Peak Watcher context menu

ScottChesworth:


src/peakWatcher.cpp
```diff
@@ -362,7 +362,7 @@ void Watcher::description(ostringstream& s) {
 	} else {
 		describeTarget(target, s);
 	}
-	s << " " << translate(this->levelTypeInfo().name);
+	s << " " << translate(this->levelTypeInfo().name) << " " << this->notifyLevel << " " << translate("threshold");
```
jcsteh, line 365: This one probably needed code review. The problem with translating this way is that there's no context for the word "threshold", which might cause problems for some languages and certainly makes things harder for translators. For future reference, I probably would have done this:

```
s << " " << translate(this->levelTypeInfo().name) << " " <<
	// Translators: Used in the Peak Watcher menu to indicate the notification threshold for a watcher.
	// {} will be replaced with the level. For example: "0 threshold"
	format(translate("{} threshold"), this->notifyLevel);
```

There's probably no point doing this now since some translators have already translated it as is, but it's worth keeping in mind for future changes like this.

ScottChesworth:
Ah my bad. Thanks for explaining. Understood now. I'll mark the next handful of times where I add translation strings for review first then, still got a lot to learn.

jcsteh:
Unfortunately, with something like OSARA, exactly what to do with translatable strings is a bit of an imprecise science. Ideal translation would mean that we never merge strings together - s << blah << blah << blah - but that's just totally impractical for something like OSARA, where we're often providing multiple pieces of information more as a list of tokens rather than trying to make a proper sentence (and where making a proper sentence would actually be worse for the user). So there are going to be many cases which are a bit of a grey area and where we could go one way or another. Anyway, best for us to discuss cases like these for a while until we can hopefully come up with some solid guidance.

jcsteh:
Another thing to consider here: does it make sense to report the threshold if automatic notification is disabled for both channels? In that case, the threshold is irrelevant, unless I'm missing something.

PR #1064, opened 2024-05-03
Add closing Media Explorer using escape key to optimal config

### LINKED ISSUE #1061: Configure REAPER for optimal screen reader accessibility question
Is disabling the following action part of parameters list of Configure REAPER for optimal screen reader accessibility?
In media explorer action list: Options: Close Media Explorer on escape key
If so, is it possible not to disable this action?
If it doesn't, there is something disabling this action and I don't know where it comes from
Thanks.

ScottChesworth:
Closes #1061 if I did it right.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -248,7 +248,7 @@ void cmdConfigReaperOptimal(Command* command) {
 		// Undock Media Explorer
 		{"reaper_explorer", "docked", 0, 0},
 		// Enable "Close window on escape key" option in Media Explorer menu
-		{"reaper_explorer", "autoplay", 0, 17},
+		{"reaper_explorer", "autoplay", 1 << 4, 17},
 		// Enable legacy file browse dialogs for more accessible check boxes
 		{"REAPER", "legacy_filebrowse", 0, 1},
 		// Prefs -> Keyboard/Multi-touch: Allow space key to be used for navigation in various windows
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -215,7 +215,7 @@ struct ReaperSetting {
 };
 // If any settings are added, changed or removed below, this number should be
 // increased.
-constexpr int REAPER_OPTIMAL_CONFIG_VERSION = 1;
+constexpr int REAPER_OPTIMAL_CONFIG_VERSION = 2;
 const char KEY_REAPER_OPTIMAL_CONFIG_VERSION[] = "reaperOptimalConfigVersion";
 
 void cmdConfigReaperOptimal(Command* command) {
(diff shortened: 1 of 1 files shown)


src/config.cpp
```diff
@@ -244,13 +245,17 @@ void cmdConfigReaperOptimal(Command* command) {
 		return;
 	}
 	const ReaperSetting settings[] = {
+		// Undock Media Explorer
 		{"reaper_explorer", "docked", 0, 0},
+		// Enable "Close window on escape key" option in Media Explorer menu
+		{"reaper_explorer", "autoplay", 0, 17},
```
jcsteh, line 251: Unfortunately, I'm fairly sure this is going to break things. The fact that we need the value 17 here suggests this is one of those annoying config values that combines multiple settings into one using bit flags. That means that if someone has one of the other affected settings set to something other than the default, this change will clobber it.

My educated guess is that the flag we want is either 1 or 16 (1 << 4), since 17 is most likely a combination of those two bits.

You could try disabling this option again and checking the value of the autoplay key, but it's possible REAPER just removes it when all flags are set to the default. In that case, you'll need to find some other setting that also impacts this autoplay key (probably Media Explorer autoplay settings). After that, you can disable the "Close window on escape key" setting and see what the difference is. For example, say you find some other setting which changes that autoplay key from 17 to 25. Now you disable "Close window on escape key" and the value goes down to 9. Now, you know the bit flag you want is 16 (1 << 4).

jcsteh (changes requested):

ScottChesworth:
Already compared that. Autoplay started out set to 1, changed to 17 when I enabled the menu option we wwant, goes back to 1 again when I disable it.
Seeing as that was stock ini I figured we'd be safe to write the 17, is that not the case?
Would it be easier to ask Justin how this autoplay thingy works? I'm already lost lol

jcsteh:
No, because someone might have a non-stock ini. We only want to change the settings we say we're going to change, not other settings that are stored in the same key. I don't know what settings this controls, but we know it's at least two because there's 1 and there's 16. We now know that 16 controls the setting we want. I don't know what 1 controls, but the point is that if someone disables 1 (so they get autoplay = 0), writing autoplay = 17 would reset the user's change.

I wish REAPER didn't combine multiple settings into one key like this, but here we are.

jcsteh:
All of that said, given your findings, the line should be as follows:

```
		{"reaper_explorer", "autoplay", 1 << 4, 17},
```

Note that 1 << 4 is equivalent to 16. It's just easier to read that way when dealing with bit flags because it makes it clear that 16 is bit 4 in binary.

Don't look at me like that, it's Justin's fault.

ScottChesworth:
I checked with Cockos because I'm kinda nervous about clobbering existing settings. Justin said:
> the low two bits, &3 is the autoplay/stop preview/previous playback 

&4 = don't auto-mark on preview 

&8 = crossfade 

& 16 = close media explorer on escape 

should be safe to read the integer, or 16 with it, and write it back
> 
This makes my brain feel as efficient as a 486 with a fan that's jammed full of cat hair. What would you type into the computer?

jcsteh:
Exactly as I said before:

```
		{"reaper_explorer", "autoplay", 1 << 4, 17},
```

My code already handles the "or" of the flag with the existing value. What basically happens is this:

1. If the key already exists, we read the key, "binary or" (similar to mathematical addition) the first value with it and write it back. For example, if the key contains 1 and the first value in the code is 16, we write 17.
2. If the key doesn't exist at all, we write the second value, 17. That incorporates both the setting we want plus any defaults.

jcsteh:
> This makes my brain feel as efficient as a 486 with a fan that's jammed full of cat hair.

Don't worry too much about it. When you signed up for helping with OSARA code, I don't think you signed up for learning binary, not to mention binary arithmetic. 99% of the time in OSARA code, you don't need to know it. Unfortunately, because of the way REAPER stores some settings, this is in the 1% of cases where you ideally do.

jcsteh (approved):

Issue #179, opened 2018-10-31
Request: action to speak current tempo


ScottChesworth:
Would it be possible to add an action that announces the current tempo? Sighted folk have this info readily available to the right of the transport bar in the TCP.

Obviously not a high priority request, but it came to me through WhatsApp and seemed worth cataloging.

ranetto:
I would say that this action could pronounce the tempo if pressed once, like "tempo 110", and the time if pressed twice, like "4 quarters" what do you think?
Scott, It doesn't seem like a good method to decrease and increase the time to know what it is in real time.
This could lead to problems if perhaps at that moment something slows down and an increase or decrease is missing.

ScottChesworth:
I like the idea of one press for tempo, two presses for time sig. Not so keen on the idea of trying to interpret time sigs in words though, that language could get super weird super fast with more complex time signatures. Just report what's in Project Settings I guess, like "4/4".

jcsteh:
What's the advantage of the double press here? Would you ever need to know
the time signature in a hurry such that hearing the tempo first would be
too verbose? I'm thinking a single press could say, for example, "120, 5/8".

The double press isn't too hard to implement, but I also don't want to make
this more complicated than it needs to be from a UX perspective.

Definitely agree that we shouldn't try to interpret the time signature.

ranetto:
Hi James and Scott, no, maybe you were misled by the fact that I wrote the numbers in words, but I absolutely did not intend to translate the numbers into words:
I think 3/4 and the like is fine.
james, if you think that the double pressure can create problems, the single pressure is also fine where it is said first 120 and then 4/4.
I had given the indication of the single pressure for the time and the double pressure for the 4/4 only for the fact that the user could have chosen whether he wanted to hear only the time or only the time.
However, it is also fine with a single push to say both.
Thanks James and Thanks Scott.

ScottChesworth:
There's a new action - "OSARA: Report tempo and time signature at play cursor; press twice to add/edit tempo markers" - that addresses everything here. Guess we forgot to close when merging. Doh!

PR #1073, opened 2024-05-10
Implement tempo and time signature reporting

day-garwood:
Introduces a new action to report the current tempo and time signature of the project, specifically the tempo reported at the position of the play cursor.

see issue #179 for discussion.

Note: This isn’t linked to the keymap yet. Still figuring that part out.

Any feedback would be appreciated since I'm new here. Thanks.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3583,6 +3583,18 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
 		outputMessage(translate("unselected tracks/items/envelope points"));
 }
 
+void cmdAnnounceCurrentTempoAndTimeSig(Command* command)
+{
+ReaProject* proj=EnumProjects(-1, nullptr, 0);
+if(proj==nullptr) return;
+double tempo=0;
+int timesig1=0;
+int timesig2=0;
+double pos=GetPlayPosition();
+TimeMap_GetTimeSigAtTime(proj, pos, &timesig1, &timesig2, &tempo);
+outputMessage(formatDouble(tempo, 1, false)+", "+to_string(timesig1)+"/"+to_string(timesig2));
+}
+
 void cmdSwitchProjectTab(Command* command) {
 	ReaProject* oldProj = EnumProjects(-1, nullptr, 0);
 	Main_OnCommand(command->gaccel.accel.cmd, 0);
@@ -4852,6 +4864,7 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Configure REAPER for optimal screen reader accessibility")}, "OSARA_CONFIGREAPEROPTIMAL", cmdConfigReaperOptimal},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Check for update")}, "OSARA_UPDATE", cmdCheckForUpdate},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Open online documentation")}, "OSARA_OPENDOC", cmdOpenDoc},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report current tempo and time signature")}, "OSARA_TEMPOTIMESIG", cmdAnnounceCurrentTempoAndTimeSig},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Enable noncontiguous selection/toggle selection of current chord/note")}, "OSARA_MIDITOGGLESEL", cmdMidiToggleSelection},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to next chord")}, "OSARA_NEXTCHORD", cmdMidiMoveToNextChord},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to previous chord")}, "OSARA_PREVCHORD", cmdMidiMoveToPreviousChord},
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4873,7 +4873,7 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Configure REAPER for optimal screen reader accessibility")}, "OSARA_CONFIGREAPEROPTIMAL", cmdConfigReaperOptimal},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Check for update")}, "OSARA_UPDATE", cmdCheckForUpdate},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Open online documentation")}, "OSARA_OPENDOC", cmdOpenDoc},
-	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report tempo and time signature at play cursor (press twice to add/edit tempo marker)")}, "OSARA_MANAGETEMPOTIMESIGMARKERS", cmdManageTempoTimeSigMarkers},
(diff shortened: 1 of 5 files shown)


ScottChesworth:
Woo, it's excellent to see you landing code here dude! Let's have a think about where we want to put it on the key map. FYI doing that requires access to Windows and Mac, we've got @jennykbrennan handling the Mac side at the moment.
Here's how to add to the key map on Windows though, just so you know in future:
1. Make sure you're starting with an unmodified copy of the latest OSARA map. Easiest way to do that is backup reaper-kb.ini if you've got any customizations you want to preserve, then install latest OSARA and say Yes to the key map replacement prompt.
2. Open REAPER, add the new binding as usual.
3. Take reaper-kb.ini from your resource path, go to src\config\windows in your local clone of OSARA, delete the existing copy and replace it. You'll likely find an ini.bak version there too, I usually delete that as well. It'll be recreated in the next step.
4. Now run python tools\sortKeymap.py config\windows\reaper-kb.ini. This is a tool Robbie wrote that makes the diff easier to track.
5. Add the name of the action and its new binding to readme.md.

That's it. Commit, push, switch back to your prior reaper-kb.ini to restore customizations if needed.

day-garwood:
To be fair, I couldn't actually find anywhere I would put it - of course the logical choice would be something linked to T but everything I tried was taken, so I think I'll leave that to the experienced keymappers for now.
Good to know for future though. Thanks.

ScottChesworth:
Adding tempo time sig markers is on Shift+C. The C part isn't strong but it's been there forever and T is already jam packed so I guess that'll be staying where it is.
What if that becomes double-pressable? Once to hear tempo and time sig, twice to bring up the dialog for that marker type?

day-garwood:
I've just pushed a new commit so you can see what that would look like.

jcsteh:
Welcome, @day-garwood. I just wanted to let you know that I've seen this and I'm keen to review it when I can. Just a bit flat out here with very little headspace at the moment, so it might take me a few more days to get to it. Thanks for your work.

src/reaper_osara.cpp
```diff
@@ -3583,6 +3583,26 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
 		outputMessage(translate("unselected tracks/items/envelope points"));
 }
 
+void reportTempoTimeSig() {
+ReaProject* proj=EnumProjects(-1, nullptr, 0);
```
jcsteh, line 3587: While it's not necessary to get the code to compile, code blocks should be indented for good style. We use a tab character for each level of indentation. Unfortunately, we don't have automatic code formatting working yet, so this has to be done manually.

src/reaper_osara.cpp
```diff
@@ -3583,6 +3583,26 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
...
+if(proj==nullptr) return;
+double tempo=0;
+int timesig1=0;
```
jcsteh, line 3590: Please rename these to be more descriptive; e.g. num and denom. This makes it clearer what each of these actually is: numerator and denominator.

src/reaper_osara.cpp
```diff
@@ -3583,6 +3583,26 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
...
+int timesig2=0;
+double pos=GetPlayPosition();
+TimeMap_GetTimeSigAtTime(proj, pos, &timesig1, &timesig2, &tempo);
+outputMessage(formatDouble(tempo, 1, false)+", "+to_string(timesig1)+"/"+to_string(timesig2));
```
jcsteh, line 3594: Rather than concatenating strings like this, it would be cleaner and easier to follow to use format(). This way, it's much easier to see what the output will look like. Something like this:

```
outputMessage(format("{}, {}/{}",
	formatDouble(tempo, 1, false), timesig1, timesig2));
```

You could even name the format template arguments, though that might be overkill for only three arguments for an untranslatable message:

```
outputMessage(format("{tempo}, {num}/{denom}",
	"tempo"_a=formatDouble(tempo, 1, false),
	"num"_a=timesig1, "denom"_a=timesig2));
```

I could go either way.

src/reaper_osara.cpp
```diff
@@ -3583,6 +3583,26 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
...
+}
+
+void cmdManageTempoTimeSigMarkers(Command* command) {
+if(lastCommandRepeatCount==0)
+{
```
jcsteh, line 3599: Brace on the same line as the statement please, like this:

```
if (lastCommandRepeatCount == 0) {
```

src/reaper_osara.cpp
```diff
@@ -3583,6 +3583,26 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
...
+reportTempoTimeSig();
+return;
+}
+Main_OnCommand(40256, 0);
```
jcsteh, line 3603: It'd be great if you could add a comment at the end of this line to explain what action this is:

```
Main_OnCommand(40256, 0); // Tempo envelope: Insert tempo/time signature change marker at edit cursor...
```

jcsteh (changes requested):
Thanks for your work on this, and thanks for putting time and effort into contributing to OSARA. Functionally, this is great, but there are a few code style and readability nits. I tend to be a bit picky about code readability and polish. I can make these changes myself if this is overwhelming, but I figured you might like to learn from the code review.

day-garwood:
Thanks for the suggestions. I will make these changes now.
The code formatting will definitely take some getting used to, I'm not used to dealing with indentations with a screen reader (especially using Notepad2 which can sometimes butcher a text selection workflow if you're not aware that a line is indented) and I'm certainly not used to the braces on same line method. It's not impossible but it does slow me down a lot. Not sure if that's because I'm primarily a TTS user rather than Braille or whether I'm just an oddball who struggles to process all that extra verbosity when you do things like turn all punctuation and indentation information on in addition to reading the actual text.
I'd be interested to know how you and other contributors work with some of these quirks being screen reader users yourselves.
Cheers.

jcsteh:
Ideally, we'd have auto code formatting so we just wouldn't have to worry about it. There's been some work on that in #940, but it causes some problems and so isn't quite ready to merge yet. Sorry about that.

Personally, I actually find it easier to follow code with indentation, but it did take some practice. I use NVDA and I have indentation reporting set to tones, which means I can hear the indentation simultaneously with speech. That cuts down on both time and speech clutter.

I always have my punctuation set to all, not just when I'm coding. But I'm also ridiculous in that I generally have my speech rate set somewhere between 650 and 960 WPM. :)

src/reaper_osara.cpp
```diff
@@ -3583,6 +3583,25 @@ void cmdUnselAllTracksItemsPoints(Command* command) {
 		outputMessage(translate("unselected tracks/items/envelope points"));
 }
 
+void reportTempoTimeSig() {
+	ReaProject* proj=EnumProjects(-1, nullptr, 0);
+	if(proj==nullptr) return;
```
jcsteh, line 3588: nit: Brace if body to avoid ambiguity, boolean check rather than comparing against nullptr.

```suggestion
	if (!proj) {
		return;
	}
```

jcsteh (approved):
Thank you.

jcsteh:
@ScottChesworth, shall I land this now and we can update the Mac key map later or do you want to get the Mac key map updated here? Normally, I'd just land it and update both key maps later, but it feels different to have one key map updated and one not.

ScottChesworth:
Don't land just yet, will get @jennykbrennan to do Mac mapping tomorrow.

day-garwood:
Ooh nice. Since this isn't my project thought I'd go for clarity over brevity, but I also prefer boolean checks. Yay.

PR #1077, opened 2024-05-11
I179

ScottChesworth:


ScottChesworth:
Hey @day-garwood, just commited some small tweaks, figured I'd explain in a comment as well seeing as I know you're new-ish to GitHub and I don't know how comfortable you are reading diffs yet.
1. I've tweaked the names of the two functions to keep them aligned with others. We use report instead of announce. Your functions are now called reportTempoTimeSig and cmdManageTempoTimeSigMarkers.
2. Minor presentation tweak, we do "void functionNameCamelCased() {", opening on the same line.
3. We leave a single blank line between each function, even if they're closely related.
4. Adjusted the name of the new action to make the double press functionality clear in shortcut help, because I suspect a grand total of about 0.2 percent of users will look this up in the readme before screaching about a bug on RWP lol.
5. Also tweaked the action ID just for consistency.

That's all I got for now. Thanks again!
Gonna wait for @jcsteh to look this over before merging, only because it's a first contribution.

day-garwood:
I know nothing about diffs so that was very helpful. Thanks.

jcsteh:
@ScottChesworth, you unfortunately can't submit a pull request against a pull request like this. If you wanted to do that, you'd need to submit a PR to @day-garwood's repository, which @day-garwood would then have to merge. The easier way to do this is to just push commits directly to @day-garwood's branch, which you should have access to do because you're an OSARA collaborator. I'll do this myself now and close this PR, but just dropping this note here for future reference.

ScottChesworth:
Oops, didn't know that. Thanks for tidying up!

Issue #1066, opened 2024-05-07
recognize the type of item, audio, midi, video


ranetto:
Hi everyone, I don't know if this problem has already been dealt with, it frequently happens to me that I don't recognize a type of item at all, as a reaper, I often call frizzy items which are actually audio midi freeze. To recognize them, I try to open them with the midi editor, if they don't open they will be audio, then with control f2 I try to see their properties and if it gives me the properties of a wave file, then it means that I am in the presence of an audio item. Would it be possible to make sure that when I select an item after the name and everything that is announced it could tell us if it is an audio item, a midi item, or a video item? Furthermore, this would also be more consistent with the behavior of the automation items which are actually announced as such.

ScottChesworth:
I don't know whether this is possible, can imagine it being useful information to hear though. Nice idea.

jcsteh:
Where do you imagine this would get reported in the item info? I'm concerned this would make that report longer, taking more time to get to info users might want quickly.

I'm not certain, but this can probably be done by:

1. Getting the take.
2. Getting its PCM_source by calling GetSetMediaItemTakeInfo with P_SOURCE.
3. Calling GetType() on the PCM_source.

There are open questions though:

1. GetType() can return WAV, MIDI, etc. However, I suspect it will return something different if, for example, the source is an MP3. We don't want to have to learn about all those cases separately.
2. Maybe we could call GetSampleRate() on the PCM_source instead. That should return 0.0 for MIDI.
3. I have no idea at all what happens with video items. I still don't really understand how video items work in REAPER. Are they separate from the audio or is the audio and video combined? If combined, I don't think we can detect them this way.

ScottChesworth:
> Where do you imagine this would get reported in the item info?
> 
After dynamic things like mute/group/lock, before take name.

jcsteh:
Is it okay to hear "WAVE", "MP3", etc. or must it be "audio"? Note that the former would not be translatable. But also: what do sighted users see here; e.g. for a wav item vs an mp3 item? Do they look any different?

jennykbrennan:
Been pondering this one. For me who use primarily audio, and work with huge numbers of items, I'm not sure. I agree, this would be a good idea. But after experimenting a bit it gives me pause.
As I understand it, the report would be something like 
7 wav Chorus-45.wav 6 takes bar 87 beat 2 86%
[...]

PR #1079, opened 2024-05-14
Report source type in item navigation

day-garwood:
When navigating through items, the source type (audio, video, midi) is announced when reporting take information (see issue #1066).

In short, "AUDIO" is reported by default for anything that isn't tagged as "VIDEO" or "MIDI". Basic tests show this to be working nicely:
Wav, MP3 and FLAC files report "AUDIO", as do frozen MIDI items which seem to have additional content announced naturally to suggest that it is a frozen MIDI item.
Giff files specify "VIDEO".
MIDI reports MIDI, as expected.

The one thing I haven't tested are container files like MP4 or ogg, or other video formats that inherently support embedded audio (simply because I have none to test with).

Hopefully I've nailed the formatting for this one.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3157,6 +3157,13 @@ void cmdGoToMasterTrack(Command* command){
 	postGoToTrack(0);
 }
 
+string getReportableSourceType(const char* type) {
+	if((!type)||(!*type)) return "";
+	string reported_type=type;
+	if((reported_type=="VIDEO")||(reported_type=="MIDI")) return reported_type;
+	return "AUDIO";
+}
+
 void moveToItem(int direction, bool clearSelection=true, bool select=true) {
 	unsigned int undoMask = getConfigUndoMask();
 	bool makeUndoPoint = undoMask&1;
@@ -3236,6 +3243,9 @@ void moveToItem(int direction, bool clearSelection=true, bool select=true) {
 		}
 		MediaItem_Take* take = GetActiveTake(item);
 		if (take) {
+			if (auto* source = (PCM_source*)GetSetMediaItemTakeInfo(take, "P_SOURCE", nullptr)) {
+				s << " " << getReportableSourceType(source->GetType());
+			}
 			s << " " << GetTakeName(take);
 		}
 		int takeCount = CountTakes(item);


ScottChesworth:
I'm not sure why AppVeyor choked on this. Presumably it built fine locally as you've already been testing it @day-garwood?

day-garwood:
How embarrassing. Sorry about that...
Accidentally did my work on the i179 branch. So I made the mistake of switching back to master and then just pasting the file with my latest work in there. Of course that's going to fail because my cpp file still had the tempo changes in which the h file didn't declare the API for! I'll fix that now.
Sorry again.

ScottChesworth:
Glad you figured it out, apparently I still need more practice reading the AppVeyor output at this end. From here it looked like it went build... building... almost built... BONK!

ScottChesworth:
M4A reports as video at the moment. Is that an easy one to resolve? Haven't looked at the code yet.
Ogg reports audio, opus reports audio.
Mp4 reports video, other image formats jpg and png both report video. I don't think we need to distinguish between video and image, a catchall for items with visual content seems fine to me, we can expect users to know roughly what's where in their own projects.

day-garwood:
Surprised that M4A reports video. Maybe that's because that's usually the audio format used in MP4 videos? Either way it seems that's what the source type is reporting.

The code simply checks if the source type reports video or MIDI and returns that, or returns audio for everything else.

Does ogg return audio even when it contains video data?

Guessing we'll either have to expand my getReportableSourceType function or else find another way of determining the type if it does end up being released.

jcsteh:
My guess is that REAPER doesn't have an audio codec for AAC audio, so it uses the video codec instead, which does support it. I'll wager that this also looks like a video item to sighted users, aside from the fact that it doesn't display any actual video of course.

IMO, providing more info here than sighted users see is out of scope in this case.

ScottChesworth:
> I'll wager that this also looks like a video item to sighted users,
> 
Can get that confirmed tomorrow.

ScottChesworth:
Oops, looks like it was waiting on verification of which source types are distinguishable visually. I said I'd get that looked at over a year ago. Will try to do it today.

Issue #471, opened 2021-02-06
Repeating audible indicator when in ripple all tracks mode


CastleStarter:
Feature request.
When in ripple all tracks mode, and when not playing, recording or scrubbing, a short beep sound is played periodically.  Much like the noise they do in space telecommunications so you know the channel is open or closed.  e.g. Roger that beep.  Suggest something like a beep every 12 seconds.
Problem being solved is that it is so easy to completely mess up a whole project if editting and unintentionally in ripple all tracks mode.  I know there is always control Z but prevention is better than cure.
Perhaps option in OSARA options to enable /disable this feature.

Flameborn:
I am not sure how unintentional it can be if you have speech feedback, letting you know that you are indeed in *Ripple All Tracks* mode. There are other, very similar ways a project can be "messed up", for example if you select all tracks, or if you have multiple items selected across tracks.

I am not trying to blur the importance of this feature request, there should be a very good reason for it to be submitted after all, however, I am not sure I agree with it being an unintentional action.

At any rate, if this is added, it should be a toggle.

MatejGolian:
@Erion, I agree with you.

2021-02-06 9:46 GMT+01:00, Erion <notifications@github.com>:

CastleStarter:
Hi, I use ripple all tracks quite rarely and agree that you don't enter it by mistake.  The spoken feedback of the ripple mode is good.  Unlike item or track selection, the ripple state stays the same for a long time.  Well, until you change it.  You can do some editting in ripple mode, then go away and have a cup of tea, come back ready to record your new track, and forgetting you are in ripple all track mode, mess up the project.
That's my scenario.  As I say, we do have control Z so perhaps that should suffice but you do then loose that new section of audio you were inspired to make whilst drinking your tea.
So my suggestion is to to prevent it going wrong in the first place.  And if its a toggle option in control f12 then it doesn't have to annoy people who don't like beep noises smily face.
Anyway, thought I'd put the suggestion out there.

Flameborn:
I think this is beyond the goals of OSARA, namely making Reaper more usable via accessibility software/hardware.

Having said that, a script could be easily written to warn you when such a scenario happens, such as for  Ripple Mode and Track Selection, since the selection tends to stay active as well.

I suggest the ingenious name of *Tea Time*, based on your use case, as well as on the fact that tea is the best beverage anyone could wish for ☺️

ScottChesworth:
[...]

PR #1080, opened 2024-05-17
Inform users if ripple is enabled when they cut or delete items

ScottChesworth:
Scenarios where hearing some sort of reminder that ripple is enabled would've been helpful appear on my radar often, especially when teaching newcomers. Not knowing this information has also tripped up a more experienced user on RWP recently.
Please try removing items with each ripple mode, tell me whether the additional feedback is useful. Paging @ptorpey as I know he's interested in this.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -1510,6 +1510,16 @@ void postCycleRippleMode(int command) {
 	}
 }
 
+string getRippleMode(int command) {
+	if (GetToggleCommandState(40310)) {
+		return translate("ripple per-track");
+	} else if (GetToggleCommandState(40311)) {
+		return translate("ripple all tracks");
+	} else {
+		return translate("ripple off");
+	}
+}
+
 void reportRepeat(bool repeat) {
 	outputMessage(repeat ?
 		translate("repeat on") :
@@ -3510,11 +3520,9 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
 					count = countAffected(GetSelectedMediaItem, selItems);
 				}
 				if(GetToggleCommandState(40310) || GetToggleCommandState(40311)) { // Set ripple editing per-track or Set ripple editing all tracks
-				// Translators: used for  "Item: Cut selected area of items" and "Item:
-				// Remove selected area of items" when ripple is enabled.
-				// {} is replaced by the number of items affected.
+				// Inform users that a ripple mode is enabled.
 				outputMessage(format(
-					translate_plural("selected area of {} item removed with ripple enabled", "selected area of {} items removed with ripple enabled", count), count));
+					translate_plural("selected area of {} item removed, " + getRippleMode(command->gaccel.accel.cmd), "selected area of {} items removed, " + getRippleMode(command->gaccel.accel.cmd), count), count));
 				} else {
 				// Translators: used for  "Item: Cut selected area of items" and "Item:
 				// Remove selected area of items".  {} is replaced by the number of items
@@ -3534,10 +3542,9 @@ void cmdhRemoveItems(int command) {
 	Main_OnCommand(command, 0);
 	int removed = oldCount - CountMediaItems(0);
 	if (GetToggleCommandState(40310) || GetToggleCommandState(40311)) { // Set ripple editing per-track or Set ripple editing all tracks
-		// Translators: Reported when items are removed and ripple is enabled. We inform users that ripple is on because that can influence the resulting positions of items on their timeline.
-		// {} will be replaced with the number of items; e.g. "2 items removed with ripple enabled".
+		// Inform users that a ripple mode is active because that can influence the resulting positions of items on their timeline.
 		outputMessage(format(
-			translate_plural("{} item removed with ripple enabled", "{} items removed with ripple enabled", removed),
+			translate_plural("{} item removed, " + getRippleMode(command), "{} items removed, " + getRippleMode(command), removed),
 			removed));
 	} else {
 		// Translators: Reported when items are removed. {} will be replaced with the
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3506,14 +3506,23 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
 				int count = 0;
 				if(selItems == 0) { // these commands treat no item selection as if all items are selected
 					count = countAffected(GetMediaItem, CountMediaItems(nullptr));
-				} else {
+				} else if (selItems > 0) {
 					count = countAffected(GetSelectedMediaItem, selItems);
 				}
+				if(GetToggleCommandState(40310) || GetToggleCommandState(40311)) { // Set ripple editing per-track or Set ripple editing all tracks
+				// Translators: used for  "Item: Cut selected area of items" and "Item:
+				// Remove selected area of items" when ripple is enabled.
+				// {} is replaced by the number of items affected.
+				outputMessage(format(
+					translate_plural("selected area of {} item removed with ripple enabled", "selected area of {} items removed with ripple enabled", count), count));
+				} else {
 				// Translators: used for  "Item: Cut selected area of items" and "Item:
 				// Remove selected area of items".  {} is replaced by the number of items
 				// affected.
 				outputMessage(format(
 					translate_plural("selected area of {} item removed", "selected area of {} items removed", count), count));
+				}
+				break;
 			}
 		}
 	}
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3512,20 +3512,20 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
 				break;
 			}
 			default: {
-				ostringstream s;
 				int count = 0;
 				if(selItems == 0) { // these commands treat no item selection as if all items are selected
 					count = countAffected(GetMediaItem, CountMediaItems(nullptr));
 				} else {
 					count = countAffected(GetSelectedMediaItem, selItems);
 				}
+				ostringstream s;
 				// Translators: used for  "Item: Cut selected area of items" and "Item:
 				// Remove selected area of items".  {} is replaced by the number of items
 				// affected.
 				s << format(
 					translate_plural("selected area of {} item removed", "selected area of {} items removed", count), count);
 				maybeAddRippleMessage(s, command->gaccel.accel.cmd);
-				outputMessage(s.str());
+				outputMessage(s);
 			}
 		}
 	}
@@ -3543,7 +3543,7 @@ void cmdhRemoveItems(int command) {
 		translate_plural("{} item removed", "{} items removed", removed),
 		removed);
 	maybeAddRippleMessage(s, command);
-	outputMessage(s.str());
+	outputMessage(s);
 }
 
 void cmdRemoveItems(Command* command) {
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3524,11 +3524,17 @@ void cmdhRemoveItems(int command) {
 	int oldCount = CountMediaItems(0);
 	Main_OnCommand(command, 0);
 	int removed = oldCount - CountMediaItems(0);
-	if (GetToggleCommandState(40310 || GetToggleCommandState(40311)) { // Set ripple editing per-track or Set ripple editing all tracks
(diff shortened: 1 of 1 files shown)


ptorpey:
Just tested this out.  Very appropriate feedback with all ripple modes.  I think that feedback would have clued me in to what was going on with the original problem I posted to the RWP list.

 

Thanks a bunch from me and others.  You the man!

 

--Pete

 

From: ScottChesworth ***@***.***> 
Sent: Thursday, May 16, 2024 8:25 PM
To: jcsteh/osara ***@***.***>
Cc: ptorpey ***@***.***>; Mention ***@***.***>
Subject: [jcsteh/osara] Inform users if ripple is enabled when they cut or delete items (PR #1080)

 

Scenarios where hearing some sort of reminder that ripple is enabled would've been helpful appear on my radar often, especially when teaching newcomers. Not knowing this information has also tripped up a more experienced user on RWP recently.
Please try removing items with each ripple mode, tell me whether the additional feedback is useful. Paging @ptorpey <https://github.com/ptorpey>  as I know he's interested in this.

  _____  


  https://github.com/jcsteh/osara/pull/1080


Commit Summary


*	fad9ac7 <https://github.com/jcsteh/osara/pull/1080/commits/fad9ac73a2175972a5d7e1e162f0e29687f04c9d>  Inform users if a ripple editing mode is enabled when they cut or delete items


File Changes 


(1 <https://github.com/jcsteh/osara/pull/1080/files>  file)

*	M src/reaper_osara.cpp <https://github.com/jcsteh/osara/pull/1080/files#diff-9e5cb6cc5a801eb7565d8fa0f7c1175fd9fe5ef046d0f736cc2f0e6127a876e9>  (24) 


Patch Links:


*	https://github.com/jcsteh/osara/pull/1080.patch
*	https://github.com/jcsteh/osara/pull/1080.diff

ScottChesworth:
Cool so a few more questions:
1. Are there other scenarios where we need to prompt? I'm only looking to add this extra feedback in scenarios where resulting positions could be unexpected.
2. @jcsteh, I've made the new longer reports localizable and tried to explain the purpose of each clearly in translator comments. New to this stuff though, figured I'd check whether you're ok with this approach before I go any further?

jcsteh:
I hope we don't have too many more cases we need to handle like this, as I don't love that this requires two additional messages for each action we tweak, but I don't have a better approach to suggest and I can live with it.

Do we really need to have separate messages for all tracks and per track ripple though? I know they're different, but these feel very long, and as I understand it, it's mostly just about giving the user a hint. We already have a command to quickly report the ripple state. In other words, can we just have "2 items removed with ripple"?

jcsteh:
See also #471.

ScottChesworth:
> Do we really need to have separate messages for all tracks and per track ripple though?
> 
Did that because |I was pretty sure there's a visual indication of which ripple setting is active, but I've just checked and nope, the button goes green and there's a radar scan effect circling it, stays the same irrespective of whether per track or all tracks. I'll simplify this then.

ScottChesworth:
Kept enabled in the reporting (as in "did this thing with ripple enabled") because I think that makes it clearer we're informing users about the setting being on, not whether the setting being on has had an actual influence on the result. Got the new feedback going for removing selected area of items now too.
I'm finding it tricky to work this into cmdPaste without a lot of repeated code. Been trying to factor it in with the mixed reports of items and tracks being added, but hmm, it's causing this terrible pain in all the diodes down my left side...

jcsteh:
Ug. Paste is one of those cases where we don't - and can't - really do translation properly in the sense that we combine messages. So we essentially take three strings: "1 track", "4 items" and "added" and combine them together. Doing otherwise isn't possible because of the fact that there are two numbers which could have different plurals, and it would also be a lot of alternative messages; track without item, just track, just item, etc.

This is making me wonder whether we should be less grammatical about this and make it more like phrases. So rather than saying "2 items removed with ripple enabled", we just say "2 items removed, ripple enabled". So it's more like two related messages rather than a single linguistic message. That should make the paste case a lot easier to deal with, since that second bit wouldn't be part of the count translation string; it'd just be tacked onto the end.

jcsteh:
If we do tack it onto the end, that also means we can have a function - maybeAddRippleMessage or something - which handles the ripple part of the message, rather than repeating that code everywhere. Then we just call that function as needed, passing it the string stream.

Btw, do we need to report ripple when engaging record? Can't remember whether record is affected by ripple or not...

ptorpey:
Btw, do we need to report ripple when engaging record? Can't remember whether record is affected by ripple or not...

[PT] There is a “ripple record” script from Reapacdk that I use quite frequently.  Too bad it isn’t part of Reaper’s built in record function so others could easily find it.

Anyway, much as I like that script, I don’t know whether I would like the extra verbiage being reported since when I use that function I just want recording to start immediately and not be interrupted by speech.

On the other hand, this is a similar situation to deleting items that would give some users a clue as to why they might be getting unexpected results.  I could live with it either way.

 

--Pete

ScottChesworth:
The script does its stuff after recording has ended and also doesn't work reliably with some combinations of settings, I'm not gonna muddy the water by reporting that.

ScottChesworth:
> So it's more like two related messages rather than a single linguistic message.
> 
IMO the pause makes it easier to miss the additional report, and I don't like it without the pause. Hmm.
Could the function approach say "with ripple enabled"? Of course I'd be clear about how it'll be assembled in the translator comments, but I don't know enough about languages to know whether construction would be difficult.

jcsteh:
"with ripple enabled" as a separate string is a really slippery slope. It's like you're trying to make it grammatical in English because you don't like it being grammatically incorrect, but if it's not grammatical in other languages, tough. :) I don't know the specifics of other languages, but I'd guess the likelihood of this being grammatically incorrect in at least one language is pretty high.

In contrast, "ripple enabled" is just a separate chunk (or token) of the message in all languages. We're not biasing grammatical correctness towards English because it's not intended to be grammatically correct.

jcsteh:
> IMO the pause makes it easier to miss the additional report

Users who really need this are either more likely to listen to complete messages, more likely to have punctuation set to all or more likely to use faster speech rates, all of which decrease the chance of the message being missed.

jcsteh:
All of that said, I don't care strongly enough to argue any further about this, as long as you're happy to deal with complaints from translators if "with ripple enabled" annoys them. :) As I noted above, paste is already a case where things get pretty messy for translations, so this might just need to be another one of those cases if it really bothers you. I'd just ask you to really consider why you don't like "ripple enabled" without the pause, but you're okay with "with ripple enabled". If it's just about linguistic correctness, you might want to reconsider.

ScottChesworth:
> It's like you're trying to make it grammatical in English because you don't like it being grammatically incorrect, but if it's not grammatical in other languages, tough.
> 
Fair point. Ugh. Gotta learn to keep an eye on that bias. I'll have another fiddle. Will likely end up going with your suggestion.

ScottChesworth:
This uses a new function to notify about ripple now. After some thought it seems kinda daft to make translators do more work for this so I've gone with strings that have already been translated.
@jcsteh, is this more like what you were suggesting?
BTW don't merge yet even if you like it, I still need to get it going inside cmdPaste, and that's still giving me a headache.

jcsteh:
That's sort of what I mean. However, what you're doing there has a couple of problems:

1. The ripple message ends up inside the translate_plural call. This will prevent the .pot file builder from seeing it as a translatable message, but even if that worked, the code wouldn't be able to find the translatable message with the ripple stuff included. It needs to be outside of the translate_plural parentheses.
2. The comma is included in the translatable message, which is probably not what we want since we're appending the ripple message anyway.

Some thoughts to move forward:

1. Let the maybeAddRippleMessage function handle the toggle state check. That is, you would always call maybeAddRippleMessage. It would simply do nothing if ripple is off.
2. Also let maybeAddRippleMessage handle adding the prefixing comma and space.
3. It may be easier and cleaner to have maybeAddRippleMessage take an ostringstream. You would then alter other code to use that. For example:
    ```
    ostringstream s;
    s << format(
    translate_plural("selected area of {} item removed", "selected area of {} items removed", count), count));
    maybeAddRippleMessage(s, command->gaccel.accel.cmd);
    ```
    I suspect this will make cmdPaste easier too.

ScottChesworth:
Aha, look at that Jen, the first problem explains why I was getting those errors from scons that we've never seen before trying to shove this into cmdPaste earlier. Thanks!
All understood, I'll refactor it soon.

jcsteh:
In case it's helpful, take a look at shortenFxName in fxChain.cpp if you need an example of a function taking an ostringstream from a caller and shoving stuff into it.

ScottChesworth:
Thanks, will do.

ScottChesworth:
@jcsteh, is this more what you had in mind? I haven't added anything to cmdPaste yet because Jen is already adding to that function in her PR, guess I'll wait until that's merged.

jcsteh:
Somewhat, yes. However, I'm not quite sure why you've changed the translate_plural calls into separate translate calls. That's going to break non-English languages where the plural forms don't match English; e.g. where there are 3 plural forms instead of 2. Most of that code should remain the same; the only thing you need to do is change outputMessage to output to an ostringstream instead. Note that you can define the ostringstream above whichever block you need to define it above so that it is available when you need it.

Also, please brace the stuff inside the `if` block of maybeAddRippleMessage. This is good style in general, but especially important in this case because there are comments there.

ScottChesworth:
> I'm not quite sure why you've changed the translate_plural calls into separate translate calls.
> 
Hmm Think I did it once just so something I'd messed up would build for a quick test, then stepped away, and that was the first thing I read when I got back so I did it all like that lol. DOH! Will refactor again.

jcsteh:
The joys of copying your own bad code. I've never done that of course. Not me.

ScottChesworth:
Think I've got it this time, ready to go if you're happy @jcsteh.

src/reaper_osara.cpp
```diff
@@ -3503,6 +3512,7 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
 				break;
 			}
 			default: {
+				ostringstream s;
```
jcsteh, line 3515: nit: Nothing really wrong with this being here, but we don't need it until later and it tends to make things easier to understand if variables are defined closer to where they're needed.

```suggestion
```

src/reaper_osara.cpp
```diff
@@ -3512,8 +3522,10 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
 				// Translators: used for  "Item: Cut selected area of items" and "Item:
```
jcsteh, line 3522: Moving it here.

```suggestion
				ostringstream s;
				// Translators: used for  "Item: Cut selected area of items" and "Item:
```

src/reaper_osara.cpp
```diff
@@ -3512,8 +3522,10 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
...
 				// Remove selected area of items".  {} is replaced by the number of items
 				// affected.
-				outputMessage(format(
-					translate_plural("selected area of {} item removed", "selected area of {} items removed", count), count));
+				s << format(
+					translate_plural("selected area of {} item removed", "selected area of {} items removed", count), count);
+				maybeAddRippleMessage(s, command->gaccel.accel.cmd);
+				outputMessage(s.str());
```
jcsteh, line 3528: nit: There's an overload of outputMessage that takes an ostringstream directly, so we can simplify this.

```suggestion
				outputMessage(s);
```

src/reaper_osara.cpp
```diff
@@ -3524,11 +3536,14 @@ void cmdhRemoveItems(int command) {
...
+	ostringstream s;
 	// Translators: Reported when items are removed. {} will be replaced with the
 	// number of items; e.g. "2 items removed".
-	outputMessage(format(
+	s << format(
 		translate_plural("{} item removed", "{} items removed", removed),
-		removed));
+		removed);
+	maybeAddRippleMessage(s, command);
+	outputMessage(s.str());
```
jcsteh, line 3546:
```suggestion
	outputMessage(s);
```

jcsteh (approved):
Perfect, minus a few tiny nits because I'm a pain in the perfectionist.

jcsteh:
If AppVeyor is all happy (pretty sure I didn't screw anything up with my nit fixes, but best to be sure), you can go ahead and merge this. I'm going to bed soon. :)

PR #1081, opened 2024-05-20
Report item move

jennykbrennan:


ScottChesworth:
Can anyone clue us in on how to find out why checks failed? Poking at the AppVeyor log here I only see scons exiting with error code 2, I'm not seeing why that happened.

jcsteh:
I don't know why the Windows build isn't showing the error output, but the Mac output is more useful. (Wow, that's a first.)

```
duplicate symbol 'postMoveEnvelopePoint(int)' in:
    build/reaper_osara.os
    build/envelopeCommands.os
```

I haven't looked at your code, but it looks like postMoveEnvelopePoint somehow got defined in reaper_osara.cpp, when it's already defined in envelopeCommands.cpp.

jennykbrennan:
Doh! Because that branch was a dud. Should have created a new one. Let's see if I can sort this out.

jennykbrennan:
EIn other words. I experimented in reaper--osara cpp before I figured out how and where to do this. Totally slipped my mind. So, what's the easiest way to sort this out, I wonder.

jennykbrennan:
I'll let the more learned have a peak at this for now and clean up as needed. . @jcsteh  I think I might want to try my hands on making this it's own function, as you described, but how opposed are you to keeping it working this way for now?

jcsteh:
I'm really not trying to be difficult, but I don't want to merge this the way it is. There are other actions which use this function which aren't context sensitive, so this might misbehave for those.

ScottChesworth:
We're gonna have a look at this tonight, hopefully will have a cleaner commit soon.

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+					translate_plural("{} item moved down", "{} items moved down", count),
+					count));
+				break;
+			}
+//			default: {
+//				outputMessage("Unknown case ");
+//			}
+		} 
+	}
+	else postMoveEnvelopePoint(command);
```
jcsteh, line 1749: nit: else on same line as closing brace, avoid unbraced else block.
```suggestion
	} else {
		postMoveEnvelopePoint(command);
	}
```

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+void postMoveItemOrEnvelopePoint(int command) {
+	TrackEnvelope* envelope = GetSelectedEnvelope(nullptr);
+	if (!envelope) {
+		int count = CountSelectedMediaItems(0);
+		if(count==0) {
+			outputMessage(translate("no items selected"));
+			return;
+		}
+		// Translators: Reported when moving items{} will be replaced by the
+	// number of items; e.g. "2 items moved".
```
jcsteh, line 1730: Unfortunately, you're going to need to put this translators comment (or a slight variation of it) above each message. Translators comments are only associated with the translatable string immediately following them. So, as this stands, the second message won't get a comment in the .po file.

```suggestion
		// Translators: Reported when moving items{} will be replaced by the
	// number of items; e.g. "2 items moved".
```

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+		switch(command) {
+			case 40117: {
+				outputMessage(format(
```
jcsteh, line 1733:
```suggestion
				// Translators: Reported when moving items up. {} will be replaced by the
				// number of items; e.g. "2 items moved up".
				outputMessage(format(
```

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+	if (!envelope) {
+		int count = CountSelectedMediaItems(0);
+		if(count==0) {
+			outputMessage(translate("no items selected"));
+			return;
+		}
+		// Translators: Reported when moving items{} will be replaced by the
+	// number of items; e.g. "2 items moved".
+		switch(command) {
+			case 40117: {
```
jcsteh, line 1732: Comment action name.

```suggestion
			case 40117: { // Item edit: Move items/envelope points up one track/a bit
```

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+				outputMessage(format(
+					translate_plural("{} item moved up", "{} items moved up", count),
+					count));
+				break;
+			}
+			case 40118: {
+				outputMessage(format(
```
jcsteh, line 1739:
```suggestion
				// Translators: Reported when moving items down. {} will be replaced by the
				// number of items; e.g. "2 items moved".
				outputMessage(format(
```

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+		// Translators: Reported when moving items{} will be replaced by the
+	// number of items; e.g. "2 items moved".
+		switch(command) {
+			case 40117: {
+				outputMessage(format(
+					translate_plural("{} item moved up", "{} items moved up", count),
+					count));
+				break;
+			}
+			case 40118: {
```
jcsteh, line 1738:
```suggestion
			case 40118: { // Item edit: Move items/envelope points down one track/a bit
```

src/reaper_osara.cpp
```diff
@@ -1718,6 +1718,37 @@ void postSelectMultipleItems(int command) {
...
+				outputMessage(format(
+					translate_plural("{} item moved down", "{} items moved down", count),
+					count));
+				break;
+			}
+//			default: {
+//				outputMessage("Unknown case ");
+//			}
```
jcsteh, line 1746: Avoid commented out code. However, for paranoia, we can add an assertion here which will show up in debug builds if this code sees an id it doesn't expect.

```suggestion
			default:
				assert(false);
```

jcsteh (approved):
Thanks. I've made a few small changes here.

src/reaper_osara.cpp
jcsteh:
```suggestion
```

jennykbrennan:
Ah, nice. Thanks. :)

Issue #1088, opened 2024-05-25
Osara move to chord actions generating MIDI editor: Unselect all events undo points


ns-studios:
I don't know if this has been talked about before, or if there's even anything Osara can do about it, but using actions such as
OSARA: Move to next chord
OSARA: Move to previous chord
OSARA: Move to previous note in chord
OSARA: Move to next note in chord
in the midi editor seems to generate
MIDI editor: Unselect all events
undo point every time they're used which then get in the way when trying to undo an actual action like move notes up one semitone, because you have to undo the unselect all events actions first.
Ways to reproduce:
open midi editor, and select piano role
Item: Open in built-in MIDI editor (set default behavior in preferences)
Mode: Piano Roll
left/right/up/down arrow around the notes
OSARA: Move to previous chord
OSARA: Move to next chord
OSARA: Move to previous note in chord
OSARA: Move to next note in chord
then undo, and you should get at least one
MIDI editor: Unselect all events
point wen undoing.
Sometimes only one unselect undo point is made after first using left/right arrow, and then another after editing one of the notes, but sometimes every use of arrows seems to generate an undo point, which can get really annoying to undo.
Would it be possible for Osara not to generate undo points for unselect all unless the action is performed manually?

ScottChesworth:
I'm adding some undo blocks now, giving those points friendlier names that should also work well in other languages (think it'll happen automatically as the action names are already translated).
IMO removing undo points around navigation would need some discussion of the tradeoffs, seeing as navigation is tied to selection. Also... er... I don't know how that would be done yet.

ns-studios:
@ScottChesworth
I've just tried the build with the changes.
I don't think this is much better. I now get an undo point for every osara chord action.
The previous behavior with the unselect all events did exactly that at worst, but sometimes I would only get the unselect all point when using the chord osara action for the first time when opening the midi editor, and then the following actions wouldn't generate any more undo points until I edited the note and then another one would be generated.
I'm even confused why the unselecting would be happening as I'm not manually selecting events, just navigating through them.

pitermach:
In Reaper's general preferences you can exclude various kinds of selection from the undo history and one of the first things I do on a new install is turn most of them off (item navigation in particular gets very spammy). I agree with @ns-studios that ideally for my use case I'd just like the chord navigation to not spam the undo history, I found myself having to spam undos on occasion or review the history.

[...]

PR #1089, opened 2024-05-25
Added friendlier names for undo points when navigating chords and not…

ScottChesworth:
…e in MIDI Editor


ScottChesworth:
Maybe the start of improving suboptimal UX in #1088?

ScottChesworth:
Closing because this approach didn't help.

PR #1093, opened 2024-05-27
Report MIDI ticks correctly in positions and lengths when the time signature denominator is something other than 4; e.g. 6/8.

### LINKED ISSUE #1084: OSARA incorrectly reports ticks if the time signature denominator isn't 4
This confused the hell out of me until I figured out what was going on.

STR:

1. Open a new project.
2. In Project Settings, set the time signature to 6/8.
3. Insert a MIDI item.
4. Set the grid to 16th and insertion length to grid.
5. Insert a note.
6. Press left arrow then down arrow to get OSARA to report the note.
    - Expected: The length should be 240 ticks.
    - Actual: The length is 480 ticks.
7. Press control+f2 to open properties.
8. Check the length. Observe that it is 0.0.240.
9. Press alt+shift+rightArrow to move right by grid.
    - Expected: OSARA should report 240 ticks.
    - Actual: OSARA reports 480 ticks.

I think the problem is that we assume that ppq is ticks per beat, but it's actually ticks per quarter. We need to take the time signature denominator into account. TimeMap2_timeToBeats does provide this info; we just don't use it. We'll probably need to pass this into various time reporting functions.

jcsteh:
Previously, for example, if the time signature was 6/8, we would incorrectly report a 16th note as 480 ticks instead of 240 ticks. To fix this, correctly take the time signature denominator into account when using MIDI PPQ (which is per quarter note, not per beat).

Fixes #1084.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -162,7 +162,7 @@ void CALLBACK CallLater::timerProc(HWND hwnd, UINT msg, UINT_PTR event, DWORD ti
 }
 
 string formatTimeMeasure(int measure, double beat, int measureLength,
-	bool useTicks, bool isLength, bool useCache,
+	int timeDenom, bool useTicks, bool isLength, bool useCache,
 	bool includeZeros, bool includeProjectStartOffset
 ) {
 	int wholeBeat = (int)beat;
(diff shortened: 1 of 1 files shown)


jcsteh:
@RDMurray, I'd appreciate it if you could take a quick look at this if you have time, since you know this tricky time stuff a bit better than I do.

RDMurray:
It looks fine to me.

While testing I found another problem with formatNoteLength; it reports the wrong length if the time signature denominator changes during the note.

RDMurray (approved):

jcsteh:
Yeah, I'm not quite sure what to do where the denominator changes during the note. Here, I only consider the denominator at the start. But how far do we go with that? It's probably possible for the time signature to change multiple times during the same note!

PR #1094, opened 2024-05-27
Paste as new take added to cmd paste

jennykbrennan:


jennykbrennan:
Is there a way to clean up ultimately garbage commits from a PR. Or is there a reason to keep them? I do plenty of unnecessary local commits atm. Is there a strategy to not end up like this? I'm still not all that comfortable with git and all its possibilities, wonderful and horrifying in equal measure. :)

jennykbrennan:
Okay, makes sense. But that does mean I need to resist the temptation to name works in progress with sentiments I truly feel about it in the moment.  Good to know. :)

jcsteh:
For your interest, you can do a thing in git called an interactive rebase (git rebase -i) which lets you squash commits into earlier commits, change commit messages, edit the content of commits, etc. I often have a lot of garbage commits locally, but i do tend to squash them down into logical commits for polish when I submit for review. However, that's largely because I'm a perfectionist and also because I sometimes work on larger stuff where logical commits really help the reviewer to understand what I'm doing. I don't need you to do that here. :)

jcsteh:
Does this strictly operate only on the first selected item? That is, what does REAPER do if you had 5 items selected?

jennykbrennan:
It pastes the extra take/takes on all selected items. Tried it now, but it says only 2 takes added etc.  
Hmm, feels like it should say2 takes added on five items, or similar. but that also feels like too much chatter to my ears as I read it. Thoughts? Because I don't think I have any opinions on that.

jennykbrennan:
I made that more confusing than needed. When testing I copied an item with two takes and pasted that across multiple items, thus 2 takes added.  
Been a long day. :)

jcsteh:
Yeah, the verbosity does hurt a bit, but the fact that I asked this question leads me to think that at least one other person will wonder about it too, possibly only after they've messed something up and can't work out how. That said, fixing this might not be possible because translate_plural can't handle two different plurals; e.g. "2 takes added to 1 item" or "1 take added to 2 items". Drat. Maybe we can only say "2 takes added to selected items"?

jennykbrennan:
Sounds like  a reasonable compromise to me. But, and here I show how little I actually know about what I'm doing, does that mean I can't report on selected items only when there is more than one and go with the short report otherwise?

jcsteh:
You can. You'll just need two separate translate_plural messages instead of one.

jennykbrennan:
Ah, okay.  Thanks. I'll tweak.  :)

jennykbrennan:
Now this says 1 take added, when pasting to one item and 1 take added to selected items when pasting to multiple. This feels right to me.

jcsteh (approved):

src/reaper_osara.cpp
```diff
@@ -3417,6 +3419,20 @@ void cmdPaste(Command* command) {
 		// with the number of points; e.g. "2 points added".
 		outputMessage(format(
 			translate_plural("{} point added", "{} points added", added), added));
+	} else if (item && 
+		(added = CountTakes(item) - oldTakes) 
+		> 0){
```
jcsteh, line 3424: nit: Indentation

```suggestion
			(added = CountTakes(item) - oldTakes)
			> 0) {
```

jennykbrennan:
Thanks Jamie. Ah, I will choose to opt for silly mistake. :) And yeah, the translator comment thought dawned on me when coffee hit my system this morning. Doh!  Sorry about that.

Issue #1095, opened 2024-05-28
"OSARA: Report tempo and time signature at play cursor" action needs a tweak


ScottChesworth:
On the most recent REAPER Made Easy Q&A, we got a request for this action to be tweaked so that it reports at the play cursor position when transport is rolling, but reports at the edit cursor position if transport is stopped or paused. Here are some steps to show why this would be helpful:
1. Add two tempo markers, say 120BPM at the beginning of the project, then 121BPM a bar later.
2. Start playback from bar 2, hit Shift+C while transport is going, you'll hear the correct tempo reported and all is well.
3. Now hit stop, and either go back to the beginning of the timeline with Control+Home or jump to the previous tempo marker with Shift+Semicolon. This movement updates the edit cursor, not the play cursor.
4. Hit Shift+C now, and oh dear, it's still gonna tell you 121BPM, reporting the last position that the play cursor was at.

Think this should be relatively simple to tighten up. Check transport state, add conditionals that get the appropriate cursor position for each. Looks like all of the states we need to account for can be found in the reportTransportState function in reaper-osara.cpp.

@day-garwood, do you reckon you can take a run at this?

jcsteh:
Closed via #1095.

PR #1096, opened 2024-05-28
Fix reportTempoTimeSig to respect transport state

day-garwood:
Now uses the play position when playing/recording, and cursor position when stopped/paused.

In response to issue #1095.


jcsteh (approved):
Thanks. I made a small change here. We had two conditions which used GetCursorPos. Since GetPlayState returns flags and 1 is set for recording as well (since the project is playing back while it is recording), I reduced this to:
`double pos = state & 1 ? GetPlayPosition() : GetCursorPosition();`

PR #1098, opened 2024-06-01
Provide information about nearby markers and regions in OSARA: Report edit/play cursor position and transport state.

### LINKED ISSUE #1062: Include nearest markers in report from "OSARA: Report edit/play cursor position and transport state"
We provide feedback when users navigate to the nearest project marker before or after their cursor position by pressing Semicolon or Apostrophe, running "Markers: Go to previous marker/project start" or "Markers: Go to next marker/project end".
However, these actions move cursor position, it would be useful to be able to hear where the closest markers are without changing that.
Can information about closest markers be appended to what we report when users hit Control+Shift+J to run "OSARA: Report edit/play cursor position and transport state" (ID _OSARA_CURSORPOS), avoiding any extra keystrokes?
- If cursor is before the first project marker, OSARA says "before marker" name of first marker, or its number if unnamed.
- If cursor is after the last project marker, OSARA says "after marker" name of last marker, or its number if unnamed.
- If cursor is between project markers, OSARA says "between markers" name of closest marker before cursor, Comma, name of closest marker after cursor, or numbers if unnamed.
- If there aren't any project markers, OSARA says nothing.
To make these longer reports easy to digest, I'd like to add a Comma between position and transport state. Here are some examples:
"bar 5 beat 2 0%, stopped before marker whistling intro"
"bar 250 beat 4 75%, paused after marker kazoo"
"bar 193 beat 3 27%, recording between markers bongo break, saucepan solo"

I reckon it would shore up the functionality gap discussed recently on RWP in this thread:
https://groups.io/g/rwp/topic/feature_request/105745527

Your thoughts?

jcsteh:
This will report if the cursor is before the first marker in the project, right at a marker, between two markers or after the last marker in the project. It also does this for regions, except that it reports whether the cursor is inside a region rather than right at it. Given the new functionality, this action has been renamed to: OSARA: Report edit/play cursor position, transport state and nearest markers and regions. However, the key map is unchanged and existing mappings will continue to work.

Fixes #1062.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -776,7 +776,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Select from cursor to start of project: Shift+Home
 - OSARA: Select from cursor to end of project: Shift+End
 - OSARA: Remove items/tracks/contents of time selection/markers/envelope points (depending on focus): Delete
-- OSARA: Report edit/play cursor position and transport state: Control+Shift+J
+- OSARA: Report edit/play cursor position, transport state and nearest markers and regions: Control+Shift+J
  - If the ruler unit is set to Measures.Beats / Minutes:Seconds, Pressing this once will report the time in measures.beats, while pressing it twice will report the time in minutes:seconds .
 - OSARA: Delete all time signature markers: Alt+Win+Delete
 - OSARA: Toggle track/take volume envelope visibility (depending on focus): Control+Alt+V
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4174,7 +4174,8 @@ void cmdReportCursorPosition(Command* command) {
 	int state = GetPlayState();
 	double pos = state & 1 ? GetPlayPosition() : GetCursorPosition();
 	ostringstream s;
-	s << formatTime(pos, tf, FT_NO_CACHE) << " ";
+	s << formatTime(pos, tf, FT_NO_CACHE) << ", ";
+
 	if (state & 2) {
 		s << translate("paused");
 	} else if (state & 4) {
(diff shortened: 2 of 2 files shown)


jcsteh:
@ScottChesworth:

1. Is this the kind of thing you're after?
2. Your request didn't discuss the cursor being right at a marker. I included that here anyway.
3. Your request also didn't cover regions. Right now, it only reports when the cursor is within a region. However, this feels a bit biased against regions, since you don't get before, between and after regions, but that seems overkill somehow given that regions have a start and end point. Regions can also overlap, so this gets very tricky very quickly. Thoughts?

jcsteh:
I ended up adding before/between/after regions as well.

ScottChesworth:
This is  brilliant! There's a lil bit of overlap with "OSARA: Report regions, last project marker and items on selected tracks at current position", should I simplify that other action to only report items now this is covered when users hit Control+Shift+J?
Also considering adding to the action name so this is more discoverable in Shortcut Help, something like:
"OSARA: Report edit/play cursor position, transport state, nearest markers and regions".

jcsteh:
Oh crap, I completely forgot about that other action. Now I'm not quite sure why we're extending this one, heh.

Happy to leave it as is or change it or whatever.

ScottChesworth:
Let's keep this. It's nicer reporting than what we had, Control+Shift+J is already widely useed, the context you've added here hits all the points in the thread on RWP where it was requested.
I'll slim down the reporting from the older action, hopefully later today.

jcsteh:
Shall I merge this then? We can handle cleaning up the other action in another PR.

ScottChesworth:
Yeah go for it. Thanks man, this turned out great.

PR #1102, opened 2024-06-05
Make Select from cursor to start/end of project actions more robust

ScottChesworth:
@GianlucaApollaro and @dgl1984 were having problems with the reliability of the OSARA: Select from cursor to start/end of project actions. Couldn't figure out what settings were different to cause unreliability, but explicitly setting focus shored it up.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4818,16 +4818,20 @@ void cmdReportRegionMarkerItems(Command* command) {
 }
 
 void cmdSelectFromCursorToStartOfProject(Command* command) {
+	Main_OnCommand(40769, 0); // Unselect (clear selection of) all tracks/items/envelope points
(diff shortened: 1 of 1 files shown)


ScottChesworth:
Pants, think I meant to take that out. Bit of a tricky one because I couldn't reproduce the issue firsthand. I'll get Ginaluca to test a build with it removed once he's awake and will push an update asap if it doesn't need to be there.

jcsteh:
If it does need to be there, that feels like a big gotcha for a user. I selected to the start of the project, damn, now my complex, non-contiguous track selection that I spent 2 minutes carefully setting up is gone.

ScottChesworth:
Yeah agreed. I'm almost certain it can go, just wanna get a hands on test so I'm not pushing another dodgy snapshot.

PR #1105, opened 2024-06-09
Restrict OSARA: Move to next/previous CC actions to the current (last clicked) CC lane.

### LINKED ISSUE #730: Osara Doesn't Care Which CC Lane Is Selected
I've included a sample project below. It seems that Osara's CC navigation commands don't actually take the selected CC lane into account. Not that the CCs on that lane get ignored, it's rather that Osara always seems to move between all CCs on all lanes. Even if you select a completely empty lane, when using the navigation commands, Osara moves on to the other lanes.
[CC Lane Issue.rpp.txt](https://github.com/jcsteh/osara/files/8322918/CC.Lane.Issue.rpp.txt)

jcsteh:
This just restricts these actions to the current CC lane. Further work will be needed to make this more efficient, such as:

1. Making it easier for the user to choose the CC lane they want. Right now, you have to cycle through many lanes using CC: Next/Previous CC lane or use the Actions list to choose the desired CC lane directly.
2. There are some CC lanes that just don't make sense for OSARA and which OSARA already handles in other ways; e.g. velocity. CC: Next/Previous CC lane will move to these, but you won't be able to navigate in them.
3. OSARA doesn't support 14-bit CC lanes properly. For now, a 14-bit lane will include the appropriate MSB and LSB CCs, but they aren't merged.
4. Sighted users can show multiple CC lanes at once. I'm not sure how to handle that in OSARA, not least because I don't think we can access which lanes are showing without messing around with state chunks.

Still, I think we should take this one step at a time so we don't block on this forever. Unless I'm missing something, this already makes things a lot better than we have now.

This needs thorough testing before it gets merged. I'd appreciate help with that.

Fixes #730.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -925,15 +925,52 @@ class SortedMidiControlChangeIterator {
 	int sortedIndexAtPos = -1;
 };
 
+bool isCCInLane(const MidiControlChange& cc, int lane) {
+	if (lane <= 127) {
+		// CC.
+		return cc.message1 == 0xB0 && cc.message2 == lane;
+	}
+	if (lane & 0x100) {
+		// We don't support 14-bit CC properly yet. For now, at least match both MSB
+		// and LSB CCs.
+		int ccNum = lane - 0x100;
+		return cc.message1 == 0xB0 &&
+			(cc.message2 == ccNum || cc.message2 == ccNum + 32);
+	}
+	if (lane == 0x201) {
+		// Pitch.
+		return cc.message1 == 0xE0;
+	}
+	if (lane == 0x202) {
+		// Program.
+		return cc.message1 == 0xC0;
+	}
+	if (lane == 0x203) {
+		// Channel pressure.
+		return cc.message1 == 0xD0;
+	}
+	return false;
+}
+
 // Finds a single CC at the cursor in a given direction and returns its info.
 // This updates currentCC.
 MidiControlChange findCC(MediaItem_Take* take, int direction) {
+	HWND editor = MIDIEditor_GetActive();
+	int lane = MIDIEditor_GetSetting_int(editor, "last_clicked_cc_lane");
 	SortedMidiControlChangeIterator iter(take);
 	MidiControlChange cc;
 	if (direction == -1) {
-		cc = iter.previous();
+		while ((cc = iter.previous())) {
+			if (isCCInLane(cc, lane)) {
+				break;
+			}
+		}
 	} else if (direction == 1) {
-		cc = iter.next();
+		while ((cc = iter.next())) {
+			if (isCCInLane(cc, lane)) {
+				break;
+			}
+		}
 	} else {
 		cc = iter.current();
 	}
@@ -1430,7 +1467,7 @@ void moveToCC(int direction, bool clearSelection=true, bool select=true) {
 		//Translators: a MIDI program number.  {} will be replaced with the program number; e.g. "Program 5"
 		s << format(translate("Program {}"), cc.message2);
 	} else if (cc.message1 == 0xD0) {
-		// Midi channel pressure. {} will be replaced with the pressure value; e.g. "Channel pressure 64"
+		// Translators: Midi channel pressure. {} will be replaced with the pressure value; e.g. "Channel pressure 64"
 		s << format(translate("Channel pressure {}"), cc.message2);
 	} else if (cc.message1 == 0xE0) {
 		auto pitchBendValue = (cc.message3 << 7) | cc.message2;
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -90,6 +90,10 @@ struct MidiControlChange {
 	bool selected;
 	bool muted;
 
+	operator bool() const {
+		return this->index != -1;
+	}
+
 	struct ReqParams {
 		bool position = true;
 		bool message1 = false;
@@ -793,10 +797,15 @@ vector<MidiNote> getSelectedNotes(MediaItem_Take* take, int offset=-1) {
 
 using MidiControlChangeIterator = MidiEventIterator<MidiControlChange, MediaItem_Take*>;
 
-// Finds a single CC at the cursor in a given direction and returns its info.
-// This updates currentCC.
-MidiControlChange findCC(MediaItem_Take* take, int direction) {
-	MidiControlChangeIterator begin(take, {
+// #434: CC events are ordered arbitrarily and, unlike notes, order can change
+// when interacting with them. Therefore, when we are searching for the next
+// or previous CC, we need to sort CCs at the same position. This class
+// iterates CCs, sorting as necessary. Note that this is not a proper C++
+// iterator.
+class SortedMidiControlChangeIterator {
+	public:
+	SortedMidiControlChangeIterator(MediaItem_Take* take):
+	begin(MidiControlChangeIterator(take, {
 		true,  // position
 		true,  // message1
 		true,  // channel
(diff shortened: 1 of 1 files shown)


jcsteh:
@LeonarddeR, tapping you for review here if you have time because you wrote the original code and perhaps understand the obscurities of CCs better than I do.

MatejGolian:
You rock - seems to work nice so far. 😃 And you are right, this already is a huge improvement. 👍
Your point number 4 sounds intriguing and it could be interesting to be able to select just the lanes you want to see and than to perhaps be able to navigate between all the visible CCs inside them, but this may not be all that crucial. That being said, would it be possible to make Osara navigate between all CCs as long as no particular CC lane is selected? If it's possible to deselect all CC lanes, in conjunction with the MIDI event filter this could work as a kind of alternative/workaround. So use the current public behavior when no lane is selected.
Just my 2 cents - if it doesn't make sense, no problem.

jcsteh:
I don't think you can unselect the CC on the current (AKA last clicked) CC lane, so unfortunately I can't do that.

jcsteh:
I guess the question is whether not being able to walk through all CCs in the piano roll view is a problem for people. If it is, we can't merge this until we figure out a solution to the multiple CC lanes problem.

MatejGolian:
If I had to choose, I'd definitely prefer the behavior in this PR over the old one. Do you think that an Osara toggle action for switching between the new/old mode would make things too complicated/confusing?

ScottChesworth:
I'm confused, why would we toggle back to behaviour that nobody has said they find useful as things stand?

MatejGolian:
I can of course only speak for myself, but perhaps just to have the possibility to see things closer to how sighted users do. That being said, at the end of the day it may not be worth it pursuing that goal, since getting a 1 to 1 relationship may not even be possible or may prove to be a lot of work.
All in all, the behavior in this PR seems like the most logical one and is exactly what I and some others asked for, so from my perspective any thing more would just be a cherry on top.

jcsteh:
To be clear, sighted users don't see all CCs at once. What they can do is choose to create multiple lanes and thus see multiple different CCs at once. In contrast, we only have a single lane at this point.

Lo-lo78:
It doesn't seem like a waste of work to me.
Maybe it should be organized better.
We could think of adding and not replacing the two actions:
"OSARA: Move to next CC"
"OSARA: Move to previous CC"
These two actions are very useful at this time.
Changing its command id and name might help.
The midi editor section doesn't have much traffic in the keymap.
There are many free buttons and simple shortcuts that can be created without problems.

jcsteh:
@Lo-lo78, to be clear, do you have a use case right now that would be harmed by the change from navigating all CCs to navigating only CCs in the current lane? Can you explain that use case? It's helpful for me to understand how things impact real world scenarios when making decisions like this. Thanks.

Lo-lo78:
I can give you an example:
If you want to take a look and quickly edit the piano roll the two official actions are very useful.
Replacing them might be a disadvantage but having two specific actions to move in the single lane makes logical sense.
It wouldn't make sense to replace the two actions we already have.
We would lose a lot of flexibility.
This project should be analyzed step by step.
if we want to ideally solve the problem without following the steps between CC and midi events such as pitch and etc...
I don't even know if it can be done.
A quick idea that comes to mind for moving between lanes is an entry field where you enter the CC number.
While these lanes are not CC:
Velocity Off Velocity Pitch Program Channel Pressure Bank/Program Select Text Events Notation Events & Sysex.
The only events that work with the action ofReaper "Edit: Insert CC event at edit cursor in current lane" I am:
1. Pitch
2. Program
3. Channel Pressure
The others have no effect on Reaper's action.
I don't know how to imagine a simple use.
Let's pretend I want to insert a CC because I have an old hardware synth.
This synth only works with Sysex and has no CC.
We should also think about Sysex.
How to insert a Sysex?
How to select the next or previous Sysex in the piano roll?
And then there are other issues.
So here do we want to move between CCs of the same lane or is the idea different?
There are many variables and arguments.
If we talk about the next and previous CC I agree but without replacing the official actions.
Sorry for the English.

jcsteh:
You note that we would lose a lot of flexibility if we changed the existing CC navigation actions, but you don't provide a real world scenario where that would actually be useful. As I noted above, keep in mind that sighted users do not see all CCs at once.

Note that this doesn't cover inserting SysEx, etc. That is something that should be handled separately.

Lo-lo78:
@jcsteh 
I think I understand what you mean.
If sighted people don't see all the lanes it's difficult for two to work.
Fair observation.
Of course, as far as sysex is concerned, it was just an example because you often mention midi events as if they were also these CCs.
Sorry if I didn't understand.
So your intention is to be able to develop an environment similar to sighted people?
At this point of the situation, why don't you change the command id of the two actions and then let's think about a more coherent environment with Leonard?
I don't think even an entry field to search for a lane can be made.
You can go to the next and previous CC and you can only see the active lanes.
Otherwise Reaper and SWS only have these actions.
The example I gave was just to say that the two official actions are valid for navigation.
If they are replaced you should slow down the workflow a bit and remember the CCs used.
So it occurs to me that maybe going to the previous and next CC isn't enough.
It would take two actions going to the previous and next CC between the active lanes.
This way the user does not have to remember all the lanes used and does not have to search through 127 CCs.
This way the problem could be solved at its root.
Move between active lanes.
I'm writing this to you because sometimes you only use CC once and you forget it.
So moving between active lanes can help.
I don't understand if the problem is English or me writing badly! (rofl)

jcsteh:
Trying to communicate stuff like this across languages is really challenging. I appreciate your efforts.

For now, without a strong real world scenario showing that changing these actions to only use the current lane causes problems, I think we should go ahead with the change. As I noted at the beginning of this thread, this is not the final piece of work that needs to be done here, but we have to start somewhere. If we try to implement all the things at once, we will end up getting nowhere and this issue of CC lanes will just remain unresolved forever. At least this way, you can either move between the CC lanes with the CC: Next/Previous CC lane actions or you can set the CC lane with the CC: Set CC lane actions, and then you can navigate inside those lanes.

Lo-lo78:
It's definitely of no use to you.
"SWS/FNG: Show only used CC lanes"
Command id:
_FNG_ME_SHOW_USED_CC_LANES
With this action you can see the active lanes.
Why not move between these too?

MatejGolian:
I'm not sure if this is what @Lo-lo78 was getting at, but as we all likely agree, it would definitely be useful if instead of cycling through all possible lanes, we could just cycle through the ones that we actually want/need. @jcsteh, I believe you mentioned something around those lines as well. However, as I pointed out earlier this current PR is a huge improvement over the previous behavior and already makes it possible to manipulate CCs in ways that were very difficult/cumbersome before.
Coming as close to the user experience of sighted users as possible is a good idea, but ultimately what matters the most is that it has to make sense for us blind users. For example, sighted users can display several CC lanes at once if they so wish, but IMO we don't want to return to the old behavior where the CC navigation commands jumped between CCs in different lanes. So as far as practical purposes go, what sighted users can or can not see may not always play a role in our case.

jcsteh:
Sure, but my point was that as I understand it, navigating through all CCs at once is not desirable for anyone. Sighted users don't have this UX, and although OSARA users currently do, no one appears to find it useful. This last point is what I was trying to establish for sure: if there are any real world scenarios where navigating through all at once is useful, then we can't merge this change as is. If not, we should merge it and then improve things like lane selection later.

MatejGolian:
@jcsteh, understood. Many (perhaps most) users on the RWP list do not seem to use the MIDI editor that much, let alone the CC features, but maybe we could conduct a kind of poll if needed. That being said, I recall others requesting a behavioral change like in this PR, so hopefully the results would be in favor of the new solution.

ScottChesworth:
> maybe we could conduct a kind of poll if needed.
> 
Nah, use cases over votes for things like this.

MatejGolian:
@ScottChesworth, good by me, of course.
It's a pity we aren't entirely sure of what @Lo-lo78 was trying to drive across.

Lo-lo78:
@MatejGolian 
What did you not understand?
I just wanted to find a simple solutionall'issue.
I inserted the SWS action to see if we can move between the active lanes.
So we would have both the possibility to move to the next and previous CC and to move between the active lanes.
If you read the thread carefully you will understand.
If you want to combine it with the master's it seems a little premature but as they say it's just my two cents.

MatejGolian:
@Lo-lo78, thanks - I think I understand you now.

Lo-lo78:
>  This way the user does not have to remember all the lanes used and does not have to search through 127 CC.

@Lo-lo78 
But what you're saying!
The CC are 128. (rofl)
Is it possible that in all these years you haven't learned anything? :P
What patience it takes with old people!
🤗

MatejGolian:
So to summarize, I think that the essence of the argument outlined here is that the 'old' behavior lets you see what CC types exist in a given MIDI item just because Osara does not respect the chosen lane. With the current PR we lose this ability somewhat, because we have to cycle through all lanes even though most of them might not contain any CC data. To test whether a particular lane actually contains CCs we thus have to use the CC navigation commands.
If this really is the issue, I can understand it, but I still think that restricting navigation to the currently selected lane is the way to go. Moreover it has already been pointed out that this PR is just the first step and that more CC related updates will follow.
@jcsteh, would it perhaps be possible to report the number of CCs in a given lane when cycling through lanes - either before or after reporting the lane name? Maybe that could work as a compromise until something better's in place, although I would personally accept that approach even as a final solution.

jcsteh:
Fair enough. If there are a lot of CCs, this might not be particularly useful because you'll potentially end up seeing many values for a single CC before you give up pressing next CC, rather than seeing what different CCs are in your item.

@Lo-lo78, to clarify, is this behaviour of scanning all CCs something you rely upon on a daily (or even weekly) basis? If it's theoretical and it's not something you actually do, that's not something I'm willing to consider at this point. On the other hand, if this actually harms something you do regularly, then we'll need to hold off on merging this until we come up with a more complete solution. I'll flag that such a solution might be months or years away, depending on what time people have, but it is what it is.

MatejGolian:
@jcsteh, in case this really is the issue to @Lo-lo78, couldn't we really have 2 sets of actions in the meanwhile like he/she originally suggested, or alternatively report the number of CCs when cycling through lanes, like I suggested in my previous comment? I understand in case someone sees some benefit in the old behavior, but on the other hand there are users that would prefer the new behavior.

jcsteh:
I'm not a big fan of having two versions of functionality around where one is basically compensating for real world use cases the other doesn't solve properly. If this doesn't eliminate the need for the old behaviour in real world scenarios, it isn't ready yet.

Reporting the number of CCs when cycling through lanes could cause performance problems in large MIDI items, but aside from that, I don't think it really solves the problem either, given how many CCs you'd have to cycle through to know what types of CCs you have in your item. If you truly don't know what kinds of CCs you have, you could spend a very long time listening to those announcements trying to figure it out.

jcsteh:
To be clear, I'm not saying this needs to solve all the use cases. There is absolutely scope for follow-up work. However, we should not land anything that regresses real world scenarios; i.e. makes them worse. I stress "real world scenarios" here. If it's purely theoretical - "what if I might want to ...", I don't care.

Lo-lo78:
> @Lo-lo78, to clarify, is this behaviour of scanning all CCs something you rely upon on a daily (or even weekly) basis? If it's theoretical and it's not something you actually do, that's not something I'm willing to consider at this point. On the other hand, if this actually harms something you do regularly, then we'll need to hold off on merging this until we come up with a more complete solution. I'll flag that such a solution might be months or years away, depending on what time people have, but it is what it is.

My experience with CC has diminished since using Reaper. With hardware it was necessary.
With Reaper I prefer envelopes.
I happened to use CC but I found more definition in envelopes.
Midi has a definition of 960 ticks while audio has a minimum of 44100 samples.
So
@jcsteh go ahead and take your pick.
If users think it's better, that's super good.
For me
@MatejGolian is right that he could give something more.
I simply know a little about the midi protocol but that doesn't mean anything.
The experience on Reaper is still rewarding even with envelopes.
In piano roll I use the pitch midi event a lot.
I don't use the CC 7. I happen to use the envelope more than the volume.
So
@jcsteh feel free.

ranetto:
Hi everyone, saying that with reaper CCs are not needed is wrong, most synths and software use CCs to do things and move parameters that don't exist at all in parameter automation, let's just think about the whole world of scene synchronization, let's think about the automation of preamplifiers, channel modules, hardware effects that are driven by our MIDI DAW: then we also consider that many synthesizers, especially those that must reproduce the original behavior of the machine from 20 or 30 years ago, must respond perfectly to the midi implementation table: I only mention the software versions of Roland JX8p, Roland JX3p, mini Moog, and many many others. The fact of being able, for example, to map a sound onto a track with its own syx is something invaluable, it avoids having to load an entire bank when a sound does not occupy more than a few K of exclusive system. Being able to control the Lexicon effects is only possible via midi parameters. I also find it very difficult to use them, both because Reaper is poorly conceived for this, and I notice this when working with sighted clients, who often have difficulty understanding how Reaper interprets certain things, unlike other Dows like Cubase or others. , and this is precisely the fact that we too with Osara have difficulty solving this problem. On the other hand, reaper with midi is unfortunately not really a monster in all its parts, we still have a somewhat cumbersome quantization not to mention how the midi items are treated during cutting or overlapping them. I hope that we will somehow resolve all this so as to make it as practical as possible but without giving up on the midi controllers which are still so used.

jcsteh:
No one is suggesting we're getting rid of support for CCs. This is just about whether the next/previous CC actions move through every single CC or just the CCs in the current lane; e.g. only volume, only modulation, etc.

Thanks everyone for your input. I'm locking discussion in this thread to collaborators for now, as I don't think there is any further constructive conversation to be had here. I'm going to proceed with merging this after I've given @LeonarddeR a few days to review the code if he wishes to do so.

Issue #1027, opened 2024-03-14
Cut off configuration window on small monitor


Buy-One:
The max screen resolution is 1366 x 738, Windows 7 x64, REAPER 7.11.

Didn't view it on larger monitor (i'm not blind), so maybe it's not related to the monitor size.

![image](https://github.com/jcsteh/osara/assets/61803004/3f41c3bb-670b-4436-80a8-5397d4b04680)

jcsteh:
The core devs are all blind, so we can't see the screenshot, but I'm guessing you mean it's cut off vertically; i.e. not all the options fit in the dialog? That dialog is only going to get more options over time, so I think we ideally need it to be vertically scrollable. Just to double check, is it not already vertically scrollable? I'm not quite sure how to do that in a .rc file though. Hmm.

jcsteh:
Investigation note: [this article might help in making the dialog scrollable](https://www.codeproject.com/Articles/32925/Scrollable-Dialog-in-Pure-Win32-API)

Buy-One:
It's severely cut off vertically, the last option that's clearly visible is **Report MIDI notes in MIDI editor** and the dialogue isn't scrollable. It's also somewhat cut off on the right as well which affects the caption of the option **Report position when navigating chords in MIDI editor**, the last visible word is **edi** rather than **editor**.

But since i'm not using screen reader and not familiar with how it works i don't know if this really is an issue for blind users in case it's able to read text as long as it's there regardless of the text visibility.

jcsteh:
It's not an issue for users of only speech and braille, but we still want this to appear properly visually, both for sighted testers and also for screen reader users that might have some vision. (Not all screen reader users are totally blind.) So, I do appreciate you reporting this.

How much of your screen isn't being used that could be used both horizontally and vertically? I'm guessing we'd need to make this vertically scrollable in any case, but it seems like the horizontal truncation might just be a matter of the dialog coordinates being a bit wrong.

Buy-One:
You wrote:
> How much of your screen isn't being used that could be used both horizontally and vertically?

Vertically it's 287 px. Horizontally 504 px in both directions i suppose because the distance between screen edges and the dialogue seems symmetric.

You wrote:
> but it seems like the horizontal truncation might just be a matter of the dialog coordinates being a bit wrong.

Upon opening the dialogue's bottom edge is flush with the screen bottom edge, so it doesn't open in the middle, unlike it does horizontally. And of course the dialogue window isn't resizable.

[...]

PR #1106, opened 2024-06-13
Make the OSARA Configuration dialog and its controls bigger so they aren't cut off visually.

jcsteh:
Partially addresses #1027. This increases the size so things should fit better into available space, but it doesn't make that dialog scrollable.

@Buy-One, @mespotine, does this help any? From what I could tell using OCR (I can't see the screen), everything seemed to fit, but that says nothing else about how it appears visually. :)

### DIFF
--- a/site_scons/makeConfigRc.py
+++ b/site_scons/makeConfigRc.py
@@ -15,7 +15,7 @@ def makeConfigRc(target, source, env):
 			break
 	out.write(
 """#include <windows.h>
-{cid} DIALOGEX 250, 125, 185, 212
+{cid} DIALOGEX 250, 125, 500, 500
 	CAPTION "OSARA Configuration"
 BEGIN
 """.format(cid=cid))
@@ -34,7 +34,7 @@ BEGIN
 		m = RE_BOOL_SETTING.match(setting)
 		cid += 1
 		out.write(
-			'\tCONTROL {displayName}, {cid}, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 10, {y}, 200, 14\n'
+			'\tCONTROL {displayName}, {cid}, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 10, {y}, 490, 14\n'
 			.format(displayName=m.group("displayName"), cid=cid, y=y))
 		y += 20
 	out.write('\tDEFPUSHBUTTON "OK", IDOK, 10, {y}, 30, 14\n'


mespotine:
All ui-elements are visible but the window is far too big now. I would suggest to make it x=600 pixels and y=575 pixels in size(seen from non-retina/non hidpi-resolutions), that would be enough to include all of them. -------- Ursprüngliche Nachricht --------Von: James Teh ***@***.***>Datum: Do., 13. Juni 2024, 13:39An: jcsteh/osara ***@***.***>Cc: Meo-Ada Mespotine ***@***.***>, Mention ***@***.***>Betreff: [jcsteh/osara] Make the OSARA Configuration dialog and its controls bigger so they aren't cut off visually. (PR #1106)Partially addresses #1027. This increases the size so things should fit better into available space, but it doesn't make that dialog scrollable.
@Buy-One, @mespotine, does this help any? From what I could tell using OCR (I can't see the screen), everything seemed to fit, but that says nothing else about how it appears visually. :)



  https://github.com/jcsteh/osara/pull/1106

Commit Summary
06fbe2f  Make the OSARA Configuration dialog and its controls bigger so they aren't cut off visually.

File Changes  (1 file)

    M
    site_scons/makeConfigRc.py
    (4)
  

Patch Links:
https://github.com/jcsteh/osara/pull/1106.patchhttps://github.com/jcsteh/osara/pull/1106.diff

ScottChesworth:
Do we know/can anyone make an educated guess on how Cockos handle this sizing/scrolling/scaling stuff? Have I understood it right that they're using the same tools for their UIs? If so, is it likely they'd have some boilerplate we could incorporate? Happy to ask Justin about this stuff, but I also don't want to go in barking up the wrong tree.
Ditto ReaPack or SWS. Is there anything already open we can learn from/adopt to match how their dialogs are presented?

jcsteh:
In terms of sizing, they likely use the visual resource editor, which lets you do all of this using drag and drop and such visual things. Even if we had a regular sighted contributor, I don't want to go down that path because it basically means blind folks can't tweak the GUI at all without sighted help.

As for scrolling, I linked some stuff over in the associated issue which might help with that. But it seems like making it bigger might be enough for now.

jcsteh:
Raw Win32 GUI is kinda awful for doing design in code. It doesn't have inbuilt tools for positioning and sizing things neatly itself, so I think most folks use visual design tools. I hate GUI work in general, but I especially hate it with raw Win32 GUI.

Frameworks like wxWidgets are much nicer in that you can lay things out in terms of boxes, grids, etc. and the framework will figure out how to position and size things itself. But wxWidgets is very bulky and I think it'd be difficult (if not impossible) to get it to play ball running inside REAPER.

Honestly, I wish we could use HTML. It's very easy for a blind person to lay things out in HTML. Even if it's not pretty, it's functionally visible, and a sighted person could even fix up the styling separately from the markup. But there are many reasons that too is infeasible.

PR #1107, opened 2024-06-13
Feedback for more MIDI event selection actions

MatejGolian:
I haven't managed to get the build environment up and running and I'm not entirely sure that this is all that's needed, but I hope that this adds feedback for the following actions:
* Edit: Select all notes
* Select all CC events in last clicked lane
* Edit: Select all CC events in time selection (in last clicked CC lane)
It may not be much, but hey, one has to start somewhere... 🙂

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1227,6 +1227,26 @@ void postMidiSelectEvents(int command) {
 		count));
 }
 
+void cmdMidiSelCC (Command* command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int oldCount = countSelectedEvents (take);
+	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
+	int newCount = countSelectedEvents (take);
+	int count = newCount;
+//	if (count >= 0) {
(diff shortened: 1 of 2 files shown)


MatejGolian:
I'm glad I at least understood the basic principle so that it builds, but as could be expected the CC actions don't give the correct feedback, because I used the postMidiSelectEvents function. Would this require writing a postMidiSelectCCs function?
I guess I won't be able to figure that out, at least not without getting the full build environment up and running. So, sorry if this causes any kind of mess. 🙂

MatejGolian:
@jcsteh, I have finally managed to install my build environment, so in the continuation I can experiment without making a mess like this.
That being said, I'd appreciate some help/guidance.
The feedback for Edit: Select all notes works correctly, but of course that is a simple action. As is likely obvious to you, the same simple approach does not work for adding feedback to the 2 CC selection actions I wanted. The base problem is that Osara also counts the currently selected notes into the resulting number. Looking at it now that should have been obvious to me, but my question is should I close this PR and than create a new one with at least that one working action, or are you able to just include that commit from here and ignore the other two?
If I'll ever be able to add feedback to those other two actions, I don't know yet.
Thanks a bunch! 🙂

MatejGolian:
I'll close this for now and hopefully figure something out.

PR #1109, opened 2024-06-14
Feedback for Select all CC events in last clicked lane

MatejGolian:
@jcsteh, would this be a legit method of reporting the number of CCs in the last clicked lane? It's probably the best thing I'll be able to hack together... 🙂


jcsteh:
Thanks for working on this. Very nice start.

I don't quite understand why you need to unselect all events first and then select them. What happens if you get the old count, run the select action and then get the new count? Obviously, if the events are already selected, those won't get included in that count, but that doesn't seem unreasonable given that they were already selected in the first place and so that didn't change.

MatejGolian:
I thought that it would make sense if Osara always reported the number of CCs in the lane - even on repeated key presses and this is the only way I was able to achieve it. Your suggestion is logical, but on subsequent key presses it would result in 0 being reported since both counts would be equal. This does indeed make sense from a technical point of view since in reality there are no new CCs to select on subsequent key presses, but IMO there is one thing here:
Your suggestion would make things slightly less clear in case 0 got reported IMO, because it would not be possible to tell whether there actually are 0 events in the lane, or if it's just that there are no more CCs to be selected. Hence I think that it's clearer if we only report 0, if there really aren't any CCs in the lane.

ScottChesworth:
> I think that it's clearer if we only report 0, if there really aren't any CCs in the lane.
> 
+1. This behaviour maintains consistency with how we report selecting all of other things, such as all notes with same pitch inside MIDI editor, or selecting all items on selected tracks outside of it.

jcsteh:
If you want to be consistent, in all other cases, we just report the total number of whatever selected. We don't calculate a delta at all. In that case, we only need a post action, not a full hook (cmd function).

jcsteh:
Oh I see, we haven't written a countSelectedCCs function, hence your question at the beginning of whether this is a legit method of doing this. So, unfortunately, my answer is no, as much as I really appreciate the work you've put into this.

jcsteh:
If you do want to tackle this, here's what you'll need to do:

1. Write a new countSelectedCCs function, based on countSelectedEvents. Instead of using MIDI_EnumSelEvts, you'll use MIDI_EnumSellCC. Remove all the stuff about msg; that's not necessary for CCs.
2. Instead of the cmd function you have here, you'll want a postMidiSelectCCs. It will be very similar to postMidiSelectNotes, but it will use your new countSelectedCCs function.

If you don't feel comfortable tackling that, that's totally fine; I'll throw this together when I get some time.

jcsteh:
Btw, doing that should also allow you to easily solve #1112.

MatejGolian:
Thanks for the guidance. I'll have a try and see how it goes. It's not nice to give up without trying at all especially now that I have concrete instructions.

MatejGolian:
This one seems to be done and working nicely. 🙂
Now the question is, how do I add an analogical post command in the PR that's unselecting CCs? I ask because that PR does not contain the 2 new functions added in this PR yet.

jcsteh:
Thanks. I haven't had a chance to review this yet and probably won't until later, but first... that's a great question. :) It's probably easiest to just wait until this one gets merged and then work on the other one. Technically, you could rebase the other branch on this one and work on both simultaneously, but if you're not super familiar with git, you might end up getting tangled and find that more painful than it's worth.

MatejGolian:
Thanks. That other PR is exactly how this one used to be - just using a reversed version of the function I made originally.

src/midiEditorCommands.cpp
```diff
@@ -1311,6 +1311,33 @@ void postMidiSelectNotes(int command) {
...
+		++count;
+	}
+	return count;
+}
+
+void postMidiSelectCCs(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int count = countSelectedCCs (take);
+	if (fakeFocus != FOCUS_NOTE && fakeFocus != FOCUS_CC) {
```
jcsteh, line 1331: I realise that postMidiSelectEvents has this same check. However, I think it probably makes sense to just unconditionally change the focus to CC here. Unless you can think of some reason the focus should stay on notes when selecting CCs?

MatejGolian:
That sounds reasonable. Since this is a CC action, and not a 'generic' MIDI event action, switching to CC mode definitely makes sense. The issue has been addressed in the latest commit.

jcsteh (approved):

PR #1110, opened 2024-06-15
Feedback for Unselect all CC events in last clicked lane

MatejGolian:
This is the reverse of what I did in my previous PR, so if that PR doesn't meet the quality requirements, this one won't either, but it wasn't a lot of work, so no problem should it not get accepted.


MatejGolian:
Once the previous PR is merged, the sole and only line of code this PR should add is
	{40669, postMidiSelectCCs, false}, // Unselect all CC events in last clicked lane

PR #1111, opened 2024-06-15
Feedback for edit insert cc event at edit cursor in current lane

MatejGolian:
I've started working on this based on the cmdMidiInsertNote function, but this one turns to be beyond my abilities.
The problem's somewhere in the if/else block responsible for reporting the CC - as long as it's there the build fails. And I'm pretty sure that there are other problems with my code besides that.
@jcsteh (or anyone competent), could you, please take a look at it?
Further more I think that simpler feedback would be sufficient for this - something like 'CC type inserted' could be enough.
As long as Edit: Insert CC event at edit cursor in current lane succeeds, the inserted CC is automatically selected. Therefore I thought that making the function talk should be fairly easy, but not so for me...
Many thanks. 🙂

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -90,6 +90,10 @@ struct MidiControlChange {
 	bool selected;
 	bool muted;
 
+	operator bool() const {
+		return this->index != -1;
+	}
+
 	struct ReqParams {
 		bool position = true;
 		bool message1 = false;
@@ -793,10 +797,15 @@ vector<MidiNote> getSelectedNotes(MediaItem_Take* take, int offset=-1) {
 
 using MidiControlChangeIterator = MidiEventIterator<MidiControlChange, MediaItem_Take*>;
 
-// Finds a single CC at the cursor in a given direction and returns its info.
-// This updates currentCC.
-MidiControlChange findCC(MediaItem_Take* take, int direction) {
-	MidiControlChangeIterator begin(take, {
+// #434: CC events are ordered arbitrarily and, unlike notes, order can change
+// when interacting with them. Therefore, when we are searching for the next
+// or previous CC, we need to sort CCs at the same position. This class
+// iterates CCs, sorting as necessary. Note that this is not a proper C++
+// iterator.
+class SortedMidiControlChangeIterator {
+	public:
+	SortedMidiControlChangeIterator(MediaItem_Take* take):
+	begin(MidiControlChangeIterator(take, {
 		true,  // position
 		true,  // message1
 		true,  // channel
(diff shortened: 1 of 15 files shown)


jcsteh:
The reason this is failing is that the getMidiControlName function is local to this file (i.e. it isn't declared in a header), but your function occurs earlier than where getMidiControlName is defined. I'd suggest moving your new function to the bottom of the file.

In addition, there's quite a bit of duplicated code here between cmdMidiInsertCC and moveToCC for reporting the CC. That needs to be split out into a separate function which can then be called by both. That function would be of the form:

```
string describeCC(MidiControlChange cc) {
	if (cc.message1 == 0xA0) {
		// Translators: MIDI poly aftertouch. {note} will be replaced with the note
		// name and {value} will be replaced with the aftertouch value; e.g.
		// "Poly Aftertouch c sharp 4  96"
		return format(translate("Poly Aftertouch {note}  {value}"),
			"note"_a=getMidiNoteName(take, cc.message2, cc.channel),
			"value"_a=cc.message3);
	}
	if (cc.message1 == 0xB0) {
		// ... etc. etc.
	}
}
```

Your cmdMidiInsertCC function would then just do:
`outputMessage(describeCC(cc));`
and you would change moveToCC to do:
`s << describeCC(cc);`

Again, happy to take care of this if you're not comfortable with a larger change like this.

MatejGolian:
Thank you. Like I commented under the other PR. I'll give it a shot. You can still take over if I fail although I'm hopeful that I'll manage to do at least something of what needs to be done... 🙂

MatejGolian:
So far I've only moved the function to the bottom and whilst it now does build, it does not report anything.
I've tried changing the line
auto cc = findCC(take, 0);
to
auto cc = findCC(take, 1);
, but the result is the same. Is it supposed to be 0 or 1?
Anyhow, there must be some kind of other error in there.

MatejGolian:
I can confirm one thing that I kinda already touched on. If inserting a CC using that action is successful, Reaper deselects all previously selected events and automatically selects that newly inserted CC. If this is default Reaper behavior, is using the findCC function still necessary in this case?

MatejGolian:
I've tried to create the describeCC function, but of course it won't build. To include all the info provided by moveToCC, I did it a bit differently tham what you suggested, since as far as I understand including all the info in the result would not be possible without using concatenation and additional parameters.
And talking of additional parameters, If we end up doing it like this, I think there's no point in reporting the cursor position when inserting CCs, so that could mean one more extra function parameter to add - if that function should also report the cursor position, that is...
And BTW, are there some kind of debug logs created somewhere on my system I could look at when builds fail?

ScottChesworth:
> are there some kind of debug logs created somewhere on my system I could look at when builds fail?
> 
Build errors are shown near the bottom of the output from SCons. That's the only place I'm aware of.

MatejGolian:
Thanks, Scott, that helped.

MatejGolian:
@ScottChesworth, thanks again - I finally made some progress.
And @jcsteh , any objections against describeCC staying an ostringstream? The code is working the way it is now, but I can have another try at making describeCC a string if you deem it better for whatever reason. And I will add that extra parameter to it for suppressing cursor position reporting, if it's fine with you.
Now with all the above being said, the new cmdMidiInsertCC function still doesn't talk. It would seem that the function does not find the CC.

jcsteh:
It doesn't make sense to me to have the cursor position reporting in describeCC because that functionality is only used in one place, whereas the rest of it is used in both places. That is, I don't see why moveToCC couldn't just handle the cursor position reporting and just call describeCC to describe the CC. Same for the selectCC part: that can just stay in moveToCC.

Also, can a CC ever be inserted muted? If not, the mute stuff could probably stay in moveToCC as well. Then describeCC would just be returning a format(...) for each case, which is why I originally suggested it could be a string. An ostringstream is fine, but I don't think it's necessary if I'm right about the above.

Is there anything I'm missing here?

MatejGolian:
I've checked the mute thing and it doesn't seem like inserting a muted event via that action is possible so you're probably not missing anything.
I refactored the code according to your suggestion. The describeCC function still needs the take at least in one place, so I have passed it as parameter. I thought perhaps it's not worth it to have describeCC retrieve the take again just for itself, since we already have it and we can just pass it along.
That being said, the insert CC command still doesn't speak...

jcsteh:
Thanks, that looks good.

The reason this isn't working is that findCC doesn't know how to find the CC at the cursor if the user hasn't moved to that CC before. findCC(take, 1) might do the trick.

That said, you say that REAPER already deselects all other CCs and selects the newly inserted one. If that's the case:

1. Do we need to call selectCC at all when inserting?
2. If we're certain that the only CC selected is the newly inserted CC, it's probably faster to call `getSelectedCCs(take)`, check its size, then get the first CC from that vector.

MatejGolian:
It should be the way I described. That being said, the function does not yet talk always when it should, so either something's wrong with it, or I don't fully understand Reaper's behavior, or both. This is true for both the findCC technique and the vector one. BTW, is my vector code theoretically correct?

jcsteh:
That looks right to me, yes. I'll try to debug this a bit when I get a chance, but it might be a few days.

MatejGolian:
One error is that the function started counting selected events and not all events (my mistake). I wanted to try and simplify things because it seemed that those lines were part of the issue and so I ended up making a mess. But there might be other things going on.

MatejGolian:
The base problem has been fixed.
Now all this functionality needs is some improved feedback.
1. I would add the word "inserted" (and perhaps even "selected") to the end of the message so that it's clear what's going on.
2. Prepending the message with "CC" could be a good idea. Not all CCs have a name such as modulation or pitch bend and in those cases things could be confusing for people that don't know what's going on.
My question is, do new translation strings have to be added for this?

jcsteh:
Yes. The message is going to be a bit unnatural in some languages because of concatenation, but I think that's probably the best we can do in this case and the message is already a bit unnatural anyway.

jcsteh:
CC possibly doesn't need a translation - it might be one of those terms that people just don't bother translating - but I'm not entirely certain.

jcsteh:
I'd point out that moving to a CC doesn't say "CC", so do we really need that for insertion? Also, saying CC first means the user has to listen to that first before they get to the useful information.

MatejGolian:
I do agree - it doesn't feel quite right to me either. It's just that there probably isn't any perfect solution. Something like CC 3 doesn't have a name at all, so in that case people would hear stuff like "3 127 inserted selected" and that feels odd too.

ScottChesworth:
I reckon for unnamed ones, CC being there makes things clearer, in my brain it makes the numbers easier to separate out if the first one is attached to something. Can live with it being a little slower to hear.

jcsteh:
I agree for unnamed (only numbered) ones. In that case, we should do that in describeCC so that we get it for next and previous CC as well, since I'd argue those suffer from the same problem. And this is why we want reusable functions. :)

MatejGolian:
OK, sorry actually the current feedback is pretty good so far - Osara already reports 'control' at the start of every CC message report.
Osara does so much that sometimes it's possible to forget how exactly some of its aspect work. I reckon it's also because our brains automatically filter away things they deem redundant or self-explanatory.
Regardless, we still have to tweak the feedback when inserting CCs.

jcsteh:
Oh. I'm sorry for any confusion here. When you asked whether you needed to add a translatable string, I thought you just meant a translators comment and a translate() call, which is how we add translatable strings. You don't need to update the .po files; there's a bunch of automation which handles that automatically. Let me know if it's easier for me to remove those changes here; it's pretty quick for me to do given my familiarity with Git.

src/midiEditorCommands.cpp
```diff
@@ -1451,35 +1478,35 @@ void moveToCC(int direction, bool clearSelection=true, bool select=true) {
...
+	vector<MidiControlChange> selectedCCs = getSelectedCCs(take);
+	int count = static_cast<int>(selectedCCs.size());
+	if (count != 1) {
+		return;
+	}
+	auto cc = *selectedCCs.cbegin();
+	fakeFocus = FOCUS_CC;
+	ostringstream s;
+	//Translators: a MIDI CC {} will be replaced with the description of that CC; e.g. "Control 1 (Mod Wheel MSB), 64, inserted and selected"
+	s << format(translate("{}, inserted and selected"), describeCC(cc, take));
```
jcsteh, line 1506: Because you're using format() here for the new string and you're not appending anything else to it, you don't need an ostringstream. You can just do outputMessage(translate(...)).

MatejGolian:
Thanks for the clarification, I think it should be done now. Good to know that there's automation in place for this translation stuff. And no need to apologize - I maybe a bit overactive at times and should perhaps ask before doing certain things instead of assuming. Be as it may, it's good that one can basically undo almost anything on Github one way or another.

src/midiEditorCommands.cpp
```diff
@@ -1424,6 +1424,33 @@ const string getMidiControlName(MediaItem_Take *take, int control, int channel)
...
+		// "Poly Aftertouch c sharp 4  96"
+		return format(translate("Poly Aftertouch {note}  {value}"),
+			"note"_a=getMidiNoteName(take, cc.message2, cc.channel),
+			"value"_a=cc.message3);
+	} else if (cc.message1 == 0xB0) {
+		// Translators: A MIDI CC. {control} will be replaced with the control number and name. {value} will be replaced with the value of the control; e.g. "control 70 (Sound Variation), 64"
+		return format(translate("Control {control}, {value}"),
+		"control"_a=getMidiControlName(take, cc.message2, cc.channel),
+		"value"_a=cc.message3);
+	} else if (cc.message1 == 0xC0) {
```
jcsteh, line 1440:
```suggestion
	}
	if (cc.message1 == 0xC0) {
```

src/midiEditorCommands.cpp
```diff
@@ -1451,35 +1478,33 @@ void moveToCC(int direction, bool clearSelection=true, bool select=true) {
...
+		return; // Not inserted.
+	}
+	vector<MidiControlChange> selectedCCs = getSelectedCCs(take);
+	int count = static_cast<int>(selectedCCs.size());
+	if (count != 1) {
+		return;
+	}
+	auto cc = *selectedCCs.cbegin();
+	fakeFocus = FOCUS_CC;
+	//Translators: a MIDI CC {} will be replaced with the description of that CC; e.g. "Control 1 (Mod Wheel MSB), 64, inserted and selected"
```
jcsteh, line 1504:
```suggestion
	//Translators: Reported when inserting a MIDI CC. {} will be replaced with the description of that CC; e.g. "Control 1 (Mod Wheel MSB), 64, inserted and selected"
```

jcsteh (approved):
Thanks for your perseverance. I know it can be tedious going through multiple rounds of review with a perfectionist reviewer. :) There were a few final minor nits I cleaned up myself.

MatejGolian:
That's OK - I understand that there has to be quality control and that you want the code to be as nice as can be.

jcsteh:
Ah crap, didn't spot there were merge conflicts here. Not sure why that is; will look into it.

jcsteh:
Oh wait, only conflicts for rebase; squash is fine. False alarm.

Lo-lo78:
@MatejGolian 
First of all a big compliment for the work you are doing.
Works well.
I'll do a little reflection.
Sorry if I haven't tried the build versions.
Are we sure of OSARA's message?
"Control 7 (Volume MSB), 0, inserted and selected"
If we were to follow the OSARA language so as not to disorientate the user, the message could be:
"Control 7 (Volume MSB), 0, added"
It seems faster and unifies the OSARA messages.
Generally when you insert a note or midi event it is usually selected automatically.
It's logical.
I think there is no need to include in the message what usually happens when you enter an event.
The operating system also thinks like this.
It should be obvious to the user that every time you enter something it gets selected.
The message can be in my opinion faster.
It works well that way too.
👍

MatejGolian:
@Lo-lo78, good suggestion. I'll create a new PR for the tweak.

Lo-lo78:
@MatejGolian Sorry if they're only replying now.
I agree with James' observation.
When you enter an event you don't need to say that it has been added.
The user already knows this. Just like James says on the notes midi. When they are added there are no unnecessary messages.
Maybe for the cut copy and paste the event count and the consequent removal or copy or paste which tells how many notes and cc are added can be used.
In that case maybe it makes sense.

PR #1114, opened 2024-06-16
Report copy move time selection

ScottChesworth:


ScottChesworth:
I'd appreciate some hands-on testing of this. Set a time selection, move somewhere else in the project and hit X to move the content in time selection to your new cursor position, or hit C to copy it there.
I'm appending the new start position of the time selection in the primary ruler format, and somewhat unusually, reporting it in full. This is because when copying selections multiple times in a row (examples like arranging loops or copying areas of room tone), I found our usual dynamic reporting got confusing.
Thoughts?

ScottChesworth:
Note: X and C are what these actions will be mapped to in the next key map update.

src/reaper_osara.cpp
```diff
@@ -1595,6 +1595,26 @@ void postCopy(int command) {
 	}
 }
 
+void postCopyMoveTimeSelection(int command) {
+	ostringstream s;
+	double start, end;
+	GetSet_LoopTimeRange(false, true, &start, &end, false);
+	fakeFocus = FOCUS_RULER;
+	if (start != end && command == 40397) {
```
jcsteh, line 1603: This can be made a bit cleaner. You can check if start == end, then outputMessage no time selection and return. That avoids the repeated start != end checks and also the s.tellp check.

jcsteh (changes requested):

ScottChesworth:
Like that? Don't merge yet btw even if you're ok with the code, I still wanna get some feedback on how this reports.

jcsteh:
Almost. Pull your branch and take a look now; I've cleaned up a tiny bit more.

jcsteh (approved):
Approving for merge if user feedback is solid.

ScottChesworth:
There are some unusual behaviours with these actions that make me want to try this with details of the new time selection appended. I think having that info spoken would provide a clearer picture of what's changed on the timeline, it'll be easily interrupted for anyone who doesn't need that amount of detail.
But, how do I do it? I know how to call a REAPER action using Main_OnCommand(), is there an equivalent I could use to call OSARA_REPORTSEL?

jcsteh:
You'd need to split the time selection reporting stuff you want out of cmdReportSelection into its own utility function, probably taking an `ostringstream&`, and then call that from both places.

src/reaper_osara.cpp
```diff
@@ -2408,6 +2427,9 @@ PostCommand POST_COMMANDS[] = {
 	{40126, postSwitchToTake}, // Take: Switch items to previous take
 	{40057, postCopy}, // Edit: Copy items/tracks/envelope points (depending on focus) ignoring time selection
 	{41383, postCopy}, // Edit: Copy items/tracks/envelope points (depending on focus) within time selection, if any (smart copy)
+
```
jcsteh, line 2430: Extraneous blank line.

```suggestion
```

ScottChesworth:
Forgot this was open, it's already included in the June key map update so closing the PR.

PR #1116, opened 2024-06-19
Tweak feedback when inserting CCs

MatejGolian:
Per @Lo-lo78's note/observation, change the feedback once a CC's inserted.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1504,8 +1504,8 @@ void cmdMidiInsertCC(Command* command) {
 	}
 	auto cc = *selectedCCs.cbegin();
 	fakeFocus = FOCUS_CC;
-	//Translators: Reported when inserting a MIDI CC. {} will be replaced with the description of that CC; e.g. "Control 1 (Mod Wheel MSB), 64, inserted and selected"
-	outputMessage(format(translate("{}, inserted and selected"), describeCC(cc, take)));
+	//Translators: Reported when inserting a MIDI CC. {} will be replaced with the description of that CC; e.g. "Control 1 (Mod Wheel MSB), 64, added"
+	outputMessage(format(translate("{}, added"), describeCC(cc, take)));
 }
 
 void cmdMidiMoveToNextCC(Command* command) {
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1504,8 +1504,7 @@ void cmdMidiInsertCC(Command* command) {
 	}
 	auto cc = *selectedCCs.cbegin();
 	fakeFocus = FOCUS_CC;
-	//Translators: Reported when inserting a MIDI CC. {} will be replaced with the description of that CC; e.g. "Control 1 (Mod Wheel MSB), 64, added"
-	outputMessage(format(translate("{}, added"), describeCC(cc, take)));
+	outputMessage(describeCC(cc, take));
 }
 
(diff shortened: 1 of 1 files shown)


jcsteh:
Thinking about this more, we don't currently report "added", "inserted", etc. when inserting notes. Why should CCs be different?

MatejGolian:
That's a good point. We do report those kind of things in case of items tho, so probably that's where it all came from. I don't mind the last bit of information being removed. @Lo-lo78, what about you?

jcsteh:
True, but for items, this is normally due to pasting, etc., rather than deliberately inserting a single item you choose. We don't have anything better to report in that case other than "1 item added". In contrast, with CCs, we are already reporting something, specifically the thing you just inserted.

jcsteh:
To be clear, I don't object to saying "added" if that's what people prefer - users can always ignore it since it's at the end - but if we do, it does raise the question of whether we need to do this for notes as well.

MatejGolian:
I see. For what it's worth, I'm considering giving a shot at adding feedback when copying/pasting MIDI events. In case of pasting we would definitely have to report the number of events pasted. Of course that doesn't necessarily mean that we have to alter the feedback of other things already in place - just saying in case it provokes any thought. Either way, it would likely be useful to hear more opinions on this.
And BTW, if we do remove the last bit of the report when inserting CC messages, the message should no longer be translatable, right?

jcsteh:
Pasting would of course need to report a count, but that's consistent with what we do elsewhere already. The difference here is that we already report specific information about the particular thing that's being added, not just a count.

If we did remove this extra feedback, yes, the translatable message would go away too.

MatejGolian:
Thanks, I thought so.

MatejGolian:
One more question for now.
Can I try incorporating feedback for MIDI copy/paste into the functions in reaper_osara.cpp (postCopy, paste), or should I make new MIDI specific functions like postMidiCopy or MidiPaste?
I guess the former as there already seems to be some MIDI related code in the main cpp file, but that's just a guess.

jcsteh:
While not ideal, there are really only two cases where MIDI code goes in the main cpp file:

1. An action that is in the main section but is also used in the MIDI editor. A good example of this is cmdReportSelection, which has to report selection globally.
2. A utility function that is used across the code base where it would be a lot uglier to split out the MIDI specific functionality. The various functions for formatting times are a good example. The stuff about ticks is only relevant in the MIDI editor, but trying to split that out without a lot of duplication or indirection would be really nasty.

For MIDI copy and paste, I think those have dedicated actions in the MIDI editor section, so you should put that code in midiEditorCommands.cpp.

MatejGolian:
Cool, should have that done basically in no time then.

MatejGolian:
After playing more with it my vote is on only keeping the return from describeCC.

jcsteh:
I think we can just go ahead with that. It's consistent with note insertion. If it turns out users have a problem with this, we'll need to tweak both at the same time anyway.

Lo-lo78:
> That's a good point. We do report those kind of things in case of items tho, so probably that's where it all came from. I don't mind the last bit of information being removed. @Lo-lo78, what about you?

Perfect. I agree with you.
Inserting a midi event does not require messagesaggiuntivi.
All you need is the name number of the cc and the value.
The rest is obvious to the user. In fact, an additional message is not needed for midi notes.
For items it's cut copy and paste. An OSARA message makes sense. If midi events are cut copied and pasted then the event count can be useful.

PR #1117, opened 2024-06-19
Feedback for MIDI copy, cut and paste

MatejGolian:
Seems to work fine here.
For the cut operation I used the existing cmdMidiDeleteEvents function, since the feedback is in line with what we're doing outside of the MIDI editor. That is to say, we don't always verbally distinguish between cutting and deleting - instead we say that something has been 'removed'.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1266,14 +1266,15 @@ void cmdMidiPasteEvents(Command* command) {
 	int oldCount = MIDI_CountEvts(take, nullptr, nullptr, nullptr);
 	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
 	int newCount = MIDI_CountEvts(take, nullptr, nullptr, nullptr);
-	int count = 0;
-	if (newCount > oldCount) {
-		count = newCount - oldCount;
+	int added = newCount - oldCount;
+if (added <= 0) {
+		outputMessage(translate("nothing pasted"));
+		return;
 	}
 	// Translators: Reported when pasting events in the MIDI editor. {} will be replaced with
 	// the number of events; e.g. "2 events pasted".
 	outputMessage(format(
-		translate_plural("{} event added", "{} events added", count), count));
+		translate_plural("{} event added", "{} events added", added), added));
 }
 
 void cmdMidiDeleteEvents(Command* command) {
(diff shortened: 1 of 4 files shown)


src/midiEditorCommands.cpp
```diff
@@ -1260,6 +1260,22 @@ void cmdMidiInsertNote(Command* command) {
...
 }
 
+void cmdMidiPasteEvents(Command* command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int oldCount = MIDI_CountEvts(take, nullptr, nullptr, nullptr);
+	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
+	int newCount = MIDI_CountEvts(take, nullptr, nullptr, nullptr);
+	int count = 0;
+	if (newCount > oldCount) {
```
jcsteh, line 1270: This will mean we report "0 events added" when pasting nothing. I wonder whether we should report "nothing pasted" here like we do for cmdPaste for consistency. That also simplifies things a bit:

```
int added = newCount - oldCount;
if (added <= 0) {
	outputMessage(translate("nothing pasted"));
	return;
}
```

src/midiEditorCommands.cpp
```diff
@@ -1272,6 +1288,19 @@ void cmdMidiDeleteEvents(Command* command) {
 		translate_plural("{} event removed", "{} events removed", removed), removed));
 }
 
+void postMidiCopyEvents(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int count = countSelectedEvents (take);
+	if (fakeFocus != FOCUS_NOTE && fakeFocus != FOCUS_CC) {
+		fakeFocus = FOCUS_NOTE;
```
jcsteh, line 1296: Why do we want to reset fakeFocus to note here? That's inconsistent with a normal paste and I don't really follow the logic from a user perspective.

jcsteh (changes requested):
Thanks.

MatejGolian:
BTW, can I accept your changes directly or do I have to submit a commit? The latter does not seem possible over here.

MatejGolian:
Be as it may, I submitted a commit.

jcsteh:
I didn't submit those as actual suggested changes that time, just a comment, which is why you didn't have an option to commit it.

MatejGolian:
Yes, these are the standard, already mapped, actions.

Lo-lo78:
I don't know if it's interesting for the community.
It is also possible to distinguish midi events if needed.
With lua scripts I was able to distinguish midi notes from CCs.
The OSARA message with the lua scripts tells me:
"10 notes selected, 2 CC selected"
"10 notes copied, 2 CC copied"
"10 notes removed, 2 CC removed"
"10 notes added, 2 CCs added"
And when I delete the events they announce the same message as the cut:
"10 notes removed, 2 CC removed"
Maybe you don't need this job. It was just a gpt chat game.
But it worked well.
@MatejGolian 
I don't think you need the scripts but if you need the scripts I'll leave them here on Github

MatejGolian:
@jcsteh, task complete. I've added the new lines a bit elsewhere, but IMO having them there near the delete events command makes more sense.

jcsteh (approved):

Issue #1121, opened 2024-06-20
有关cc 动作的报告反馈


cn-peter:
hello everyone
I have a few friends who arrange music. They have a need to edit CC
Can the following actions be reported?

Edit: Move CC events right 1 pixel; ID:40673
 Edit: Move CC events left 1 pixel; ID:40672
Edit: Move CC events right by grid; ID:40675
Edit: Move CC events left by grid; ID:40674

I am speaking here using a translator
Thank you

ScottChesworth:
I think we have a guy who can do that. :)

MatejGolian:
Indeed... 🙂
From my understanding we'd have to write a function like postMidiMoveStart, just for CCs. I'm sure that Jamie would be able to get all MIDI event movement feedback working even from a single function, but perhaps having a dedicated function for CCs could be better for function separation/clarity?

PR #1122, opened 2024-06-20
Added feedback for CC movement commands

MatejGolian:
I tried my hand at #1121, but I'm afraid that I have reached the end of my abilities.
It does partially work, but with just one CC selected, triggering those actions repeatedly after one another always results in the original start position being spoken. Apart from that it's as if the Osara CC navigation commands needed time to catch on to the new positions. I'm sure it's just my code though - it's probably missing a lot of crucial stuff.
Another thing is that outputting the full CC description in this case, like I've done, may be overkill. Maybe we could do without reporting the CC values similarly how we don't report note lengths when moving them either.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -2067,6 +2067,62 @@ void postMidiChangePitch(int command) {
 	outputMessage(s);
 }
 
+void postMidiMoveCCs(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	// Get selected CCs.
+	vector<MidiControlChange> selectedCCs = getSelectedCCs(take);
+	int count = static_cast<int>(selectedCCs.size());
+	if (count == 0) {
+		return;
+	}
+	ostringstream s;
+	if (count > 1) {
+		switch (command) {
+			case 40672:
+				// Translators: Used when moving CCs in the MIDI
+				// editor. {} is replaced by the number of CCs, e.g. 
+				// "3 CC events pixel left"
+				s << format(
+					translate_plural("{} CC event pixel left", "{} CC events pixel left", count), count);
+				break;
+			case 40673:
+			// Translators: Used when moving CCs in the MIDI
+				// editor. {} is replaced by the number of CCs, e.g. 
+				// "3 CC events pixel right"
+				s << format(
+					translate_plural("{} CC event pixel right", "{} CC events pixel right", count), count);
+				break;
+			case 40674:
+				// Translators: Used when moving CCs in the MIDI
+				// editor. {} is replaced by the number of CCs, e.g. 
+				// "3 CC events grid unit left"
+				s << format(
+					translate_plural("{} CC event grid unit left", "{} CC events grid unit left", count), count);
+				break;
+			case 40675:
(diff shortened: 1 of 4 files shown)


MatejGolian:
I've managed to derive a full CC function based on postMidiMoveStart after all, but the issues I mentioned in my previous comment still persist.
1. When moving multiple CCs on the same position, Osara keeps reporting the original position and not the new one.
2. When using Osara CC navigation commands after moving CCs, Osara doesn't register if those CCs have been moved to different positions. Unselecting everything and navigating CCs again helps.

MatejGolian:
@jcsteh, I think this is really as far as I'm able to take this. I've added a 3rd parameter to describeCC to optionally disable value reporting, which makes feedback when moving CCs analogical to that when moving notes.
Unfortunately I'm not able to solve the problem where Osara keeps reporting the old positions unless one performs an unselect all operation.
Any idea what to do about it? I really don't get why it behaves the way it does.

jcsteh:
I don't really understand why this isn't working. I did look at your code briefly and can't see the problem. I'll do some debugging when I get a chance.

jcsteh:
I think we might be seeing a REAPER bug here. When you use those actions, the properties dialog also fails to update until you select the same CC again. I'll see what Cockos have to say.

jcsteh:
Cockos agrees this is a bug and is fixing it.

ScottChesworth:
@MatejGolian , try 7.17RC5, do things work as expected now?

MatejGolian:
@ScottChesworth, thanks for the notification. Yes, looks good tome now.

jcsteh:
> I've added a 3rd parameter to describeCC to optionally disable value reporting, which makes feedback when moving CCs analogical to that when moving notes.

Hmm, I'm not sure about this. I know we don't report velocity for notes when moving them, but velocity is more of a secondary value, rather than a primary value. In contrast, moving hold 0 vs hold 127 is a very different thing and those might be in close proximity.

MatejGolian:
In case of notes we report nor length nor velocity when moving, but you're right that notes and CCs are different and therefore could/should be treated differently as well. If you think that reporting the full description makes more sense, fine with me. So feel to make any adjustments you deem fit.

jcsteh:
Any thoughts on this @ScottChesworth? I don't want to undo work here if it really does provide a better experience.

MatejGolian:
I figured I'd remove the reportVal parameter from describeCC just to get the work done. As far as I understand we can always revert the change should we come to the conclussion that having it is useful.

src/midiEditorCommands.cpp
```diff
@@ -1485,18 +1485,18 @@ string describeCC(MidiControlChange cc, MediaItem_Take* take) {
 			"note"_a=getMidiNoteName(take, cc.message2, cc.channel),
 			"value"_a=cc.message3);
 	}
-	if (cc.message1 == 0xB0) {
+	if (cc.message1 == 0xB0 ) {
```
jcsteh, line 1488: nit: Extraneous white space.

```suggestion
	if (cc.message1 == 0xB0) {
```

jcsteh (approved):

jcsteh:
In the absence of additional thoughts, I'm making an executive decision here that we should report the value.

PR #1124, opened 2024-06-21
Report heal splits in items

jennykbrennan:
This reports number of splits healed,(nothing healed if none), and number of items selected after healing.


jennykbrennan:
Ah, gotcha. 
Thanks for sorting this out. The word conflict gives me hives. :)

PR #1126, opened 2024-06-25
Added feedback for 'Select' and 'Unselect all CC events'

MatejGolian:
Provides feedback for actions to quickly select/unselect all CC events in a MIDI item.
These actions are not mapped to any keyboard shortcuts at the moment.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -915,11 +915,12 @@ This list is worth referencing when making your own key map additions, assigning
 - Reset all MIDI control surface devices
 
 #### Unmapped in MIDI Editor section
-- Edit: Select all CC events in time selection (even if CC lane is hidden)
 - Edit: Select all events in time selection (even if CC lane is hidden)
 - Edit: Select all notes in time selection
 - Options: MIDI inputs as step input mode
 - Options: F1-F12 as step input mode
+- Select all CC events
+- Unselect all CC events
 
 #### Unmapped OSARA actions
 - OSARA: Pause/resume Peak Watcher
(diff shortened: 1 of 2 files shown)


MatejGolian:
Since we voice selecting all events, all notes and all CCs in the current lane, voicing these 2 actions as well makes sense to have more hierarchical levels covered, at least IMO. That being said, it's not essential, so you don't have to accept this PR if you think that that's not a valid enough reason.

MatejGolian:
If I understand and remember correctly, yes that was intentional, since I had that line doubled in the readme. I think it's still doubled and I've opened a new separate PR for it, so don't know how to fix this PR now since merge conflicts have arisen.

Lo-lo78:
Hi maybe when you deselect all the CCs of a lane it should say:
"unselected".
For example:
"2 CC unselected". 
It shouldn't say how many CCs are left.selected.
Perhaps the first few times it could be misinterpreted by users.
As regards all lanes the count could be for:
for example:
"3 CC unselected" or "All CC unselected.".
For me, however, that's fine too.
It's just for message consistency.
But it's understandable.

MatejGolian:
We have discussed this with @jcsteh previously and we came to the conclusion that as far as Osara behavior is concerned this is the more standard way of doing it.
I agree, it is a bit unintuitive at first, but all things considered, hearing how many selected CCs there are at a given time is much more useful than just hearing how many of them have been unselected.

Lo-lo78:
Ok!
Perfect.
The choice is logical.
👍

Issue #1076, opened 2024-05-11
Key map update for June 2024


ScottChesworth:
This is where we'll be discussing and tracking work on the next OSARA key map update, aiming to merge at the end of June 2024. Please check out the proposals below, raising any concerns or alternative suggestions in the comments. If you want to remind us about an open issue that could be included, put its issue number in your comment so GitHub can link stuff up, it helps us track work. I'm also adding a list of supporting dev tasks below, mainly adding/improving speech feedback this month. All help appreciated. Test builds will be added here in due course, probably around the start of June.
## Main Section Additions and Changes:
- Remove SWS/PADRE: Envelope LFO generator from the O key, as it causes REAPER to crash in a handful of easily reproduceable situations and doesn't seem to be getting any TLC from SWS yet.
- Add a new OSARA action to the O key that manages tempo markers, one press to report tempo/time signature at play cursor, two presses to insert/edit them (contributed by @day-garwood in PR #1073).
- Add Time selection: Move contents of time selection to edit cursor (moving later items) to unmodified X, replacing Item: Crossfade items within time selection which we've never had feedback for and doesn't seem to be widely adopted.
- Add Time selection: Copy contents of time selection to edit cursor (moving later items) bound to unmodified C, thus removing a tempo marker action with a name that's too vague and is now redundant.
Note: Cockos specifically designed these two copy/move actions to opperate with a single click, hence my rationale for assigning them to single letter shortcuts.
- Item: Copy selected area of items needs to move from Control+Windows+C as that keystroke has long-since been commandeered by the OS for inverting contrast, functionality that is difficult to disable and runs without providing feedback to any screen reader.
- Ideally I'd like to keep all of these "area of item" keystrokes unified under one set of modifiers, possibly Control+Shift? With that goal in mind...
- Edit: Copy items/tracks/envelope points (depending on focus) within time selection, if any (smart copy) would move to Shift+C, replacing a tempo marker related action that's now called by a double-press of O, has stronger phenetic association between the word tempo and the keystroke.
- Accompanying smart cut would move to Shift+X, replacing Item: Remove content (trim) behind items, I reckon that's too niche to need a default binding.
- Item: Trim items to selected area moves to Control+Shift+T, replacing Take: Delete active take from items (prompt to confirm) which moves to Control+Windows+Delete, seeing as it deletes stuff.
- Item: Remove selected area of items moves to Control+Shift+Delete to be consistent with modifiers above.
[...]

PR #1127, opened 2024-06-25
Keymap june24

ScottChesworth:
Includes all changes proposed in #1076 for wider testing


jcsteh:
Minor points for future reference/thought:

1. I'd really prefer to have the summary of key map changes included in the commit message. Having it separate means we lose track of it in the commit history and it's unclear what changed in a given commit.
2. For big key map changes, I don't think code changes should be included in the same commit. I think it's fine to include a single key map change in the same commit as a single code change. However, when it's a larger key map commit, again it gets blurry what is changing and why.

ScottChesworth:
Ah sorry about that, will keep those points in mind next time. I only included code here because I goofed, thought some feedback I implemented in another PR had already been merged. You'd already reviewed the other PR, so the code itself should be up to snuff.

jcsteh:
Yeah, the code is fine, just purely about organisation and history. It's no biggy, just things for future reference. Thanks.

PR #1128, opened 2024-06-28
Add feedback when quantizing items in TCP

ScottChesworth:


ScottChesworth:
Anyone know why that check failed instantly? It was saying the branches could be merged before I opened the PR.

jcsteh:
I have absolutely no idea. I think AppVeyor is on acid.

ScottChesworth:
Where does it get that from? Asking for a friend...

jcsteh:
That said, I'm surprised this built for you locally. It won't build here because of the declaration in the header file.

ScottChesworth:
What did I do wrong in there? Weird, it builds fine here.

jcsteh:
In header files, we don't (and can't) have "using namespace std", so you have to prefix anything in the std namespace; e.g. std::string. I really don't understand why this built for you and not me. Very odd.

PR #1133, opened 2024-07-06
Add project and take markers action

ScottChesworth:
This improves support for take markers, by making adding them context sensitive. Navigation of them will be added to the key map in the next commit.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4793,8 +4793,33 @@ void cmdInsertMarker(Command* command) {
 		Main_OnCommand(command->gaccel.accel.cmd, 0);
 		return;
 	}
+	if(CountSelectedMediaItems(0)>0) {
+		vector<int> preMarkers(CountSelectedMediaItems(nullptr));
+		for(int i = 0; i < preMarkers.size(); ++ i) {
+			MediaItem* item = GetSelectedMediaItem(nullptr, i);
+			MediaItem_Take* take = GetActiveTake(item);
+			preMarkers[i] = GetNumTakeMarkers(take);
+		}
+		Main_OnCommand(42390, 0); // Item: Quick add take marker at play position or edit cursor
+		int addedMarkers = 0;
+		for(int i = 0; i < preMarkers.size(); ++ i) {
+			MediaItem* item = GetSelectedMediaItem(nullptr, i);
+			MediaItem_Take* take = GetActiveTake(item);
+			if (preMarkers[i] < GetNumTakeMarkers(take)) {
+				++ addedMarkers;
+			}
+		}
+		if (addedMarkers == 0) {
+			return; // not inserted
+		} else {
+			// Translators: Reported when using REAPER's quick add take marker action. [] will be replaced with the number of take markers that have been inserted. E.G. "2 take markers inserted".
+			outputMessage(format(
+				translate_plural("{} take marker inserted", "{} take markers inserted", addedMarkers), addedMarkers));
+			return;
+		}
+	}
 	int count = CountProjectMarkers(nullptr, nullptr, nullptr);
-	Main_OnCommand(command->gaccel.accel.cmd, 0);
+	Main_OnCommand(40157, 0);
 	if (CountProjectMarkers(nullptr, nullptr, nullptr) == count) {
 		return; // Not inserted.
 	}
@@ -5252,6 +5277,7 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Check for update")}, "OSARA_UPDATE", cmdCheckForUpdate},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Open online documentation")}, "OSARA_OPENDOC", cmdOpenDoc},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report tempo and time signature at play cursor; press twice to add/edit tempo markers")}, "OSARA_MANAGETEMPOTIMESIGMARKERS", cmdManageTempoTimeSigMarkers},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Add project or take marker at cursor (depending on focus)")}, "OSARA_ADDPROJTAKEMARKER", cmdInsertMarker},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Enable noncontiguous selection/toggle selection of current chord/note")}, "OSARA_MIDITOGGLESEL", cmdMidiToggleSelection},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to next chord")}, "OSARA_NEXTCHORD", cmdMidiMoveToNextChord},
 	{MIDI_EDITOR_SECTION, {DEFACCEL, _t("OSARA: Move to previous chord")}, "OSARA_PREVCHORD", cmdMidiMoveToPreviousChord},
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4812,9 +4812,9 @@ void cmdInsertMarker(Command* command) {
 		if (addedMarkers == 0) {
 			return; // not inserted
 		} else {
-			// Translators: Reported when using REAPER's quick add take marker action. [] will be replaced with the number of take markers that have been inserted. E.G. "2 take markers inserted".
+			// Translators: Reported when using REAPER's quick add take marker action. If more than one take marker is inserted, [] will be replaced with the number of markers. E.G. "2 take markers inserted".
 			outputMessage(format(
-				translate_plural("{} take marker inserted", "{} take markers inserted", addedMarkers), addedMarkers));
+				translate_plural("take marker inserted", "{} take markers inserted", addedMarkers), addedMarkers));
 			return;
 		}
 	}
@@ -4833,6 +4833,20 @@ void cmdInsertMarker(Command* command) {
 	// Translators: Reported when inserting a marker. {} will be replaced with the
 	// number of the new marker; e.g. "marker 2 inserted".
 	outputMessage(format(translate("marker {} inserted"), number));
+}
+
+void cmdInsertOrEditMarker(Command* command) {
+	double start, end;
+	GetSet_LoopTimeRange(false, true, &start, &end, false);
+	if (start != end && CountSelectedMediaItems(nullptr)>0) {
+		Main_OnCommand(43181, 0); // Item: Add/edit take marker at time selection
+		return;
+	} else if(CountSelectedMediaItems(nullptr) > 0) {
+		Main_OnCommand(42385, 0); // Item: Add/edit take marker at play position or edit cursor
+	} else {
+		Main_OnCommand(40171, 0); // Markers: Insert and/or edit marker at current position
+		return;
+	}
 }
 
 void cmdInsertRegion(Command* command) {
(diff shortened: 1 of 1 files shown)


jcsteh (changes requested):
There are a few problems here:

1. We shouldn't be using the same function (cmdInsertMarker) to both intercept a REAPER command and implement a custom OSARA command. Otherwise, the custom OSARA command might end up overriding functionality provided by an intercepted REAPER command. It's okay for a custom OSARA command to call the cmd function for an intercepted REAPER command, though. So, I think cmdInsertMarker should be renamed to cmdInsertProjectMarker and a new function cmdInsertMarker should be added.
2. We add a custom OSARA command which calls Item: Quick add take marker at play position or edit cursor, but we don't support that action itself. Some users may wish to use it directly. So, we need cmdInsertTakeMarker as well, and then cmdInsertMarker would call either that or cmdInsertProjectMarker depending on focus.
3. You define item focus here as 1 or more selected items. But that's almost always true as soon as you've moved to an item, even if you've since focused the track or the ruler. I haven't tested this, but I'll wager that if you select an item, move to a new track, scrub and move by bar, you'll still end up adding a take marker instead of a project marker. I think this needs to use fakeFocus.
4. A minor nit: strictly speaking, the new OSARA commands should probably be in the Miscellaneous Actions section of the readme because they're custom OSARA commands. The earlier sections are under "Supported REAPER and Extension Actions" and this is not a REAPER action.

ScottChesworth:
Gotcha, I can refactor that. Did you have specific fakeFocus states in mind? I started off trying to use FOCUS_ITEM and FOCUS_RULER because I figured it would be nice to Page through long takes dropping markers, but it felt like the lesser used marker type was taking precedence too often that way, so I was gonna tell people to clear selection with Shift+Escape instead. Maybe I should try removing item selection from the conditionals, rely entirely on fakeFocus instead? Hmm.

jcsteh:
I was thinking take markers would only be FOCUS_ITEM, but I guess I can see why that would make this action far less useful, precisely because you can't drop take markers while moving through the project. I guess the guidance could be that you need to select the item under the cursor to force focus back to the item when you want to drop a take marker, but that seems quite annoying too. Ug.

I'm not really sure what to suggest here. I'm not sure we can really make this context sensitive without encountering problems like this. I don't love the idea of having to clear selection to drop a project marker; it feels a bit unintuitive. I'm probably a bit biased though because I use project markers all the time, but I've never used take markers and I don't see that changing any time soon. That said, I can just keep using the existing actions, so if I'm the odd one out here, that's fine.

ScottChesworth:
I don't think you'll be the odd one out, using it more tonight, I'm not liking it yet either. Will keep trying stuff. I'm exploring the context sensitive approach as there's nowhere logical left to put adding and editing take markers. T, A, K, E, M and R are all crammed.

ScottChesworth:
Refactored with a lot of help from @TimTam, changed to using fakeFocus instead of item selection because while not ideal, it makes the context sensitivity less weighted toward take  which feels better in situ. I'll come back tomorrow and fix the key map conflict.

ScottChesworth:
Wew, resolved conflicts. Can you take a gander at this approach when you get time @jcsteh?

src/reaper_osara.cpp
```diff
@@ -4784,17 +4784,18 @@ void cmdTransientDetectionSettings(Command* command) {
 	// We must register the hook before the dialog appears or it won't work.
 	plugin_register("accelerator", &transDetect_accelReg);
 	// Open the dialog.
-	Main_OnCommand(command->gaccel.accel.cmd, 0);
+	
+Main_OnCommand(command->gaccel.accel.cmd, 0);
```
jcsteh, line 4788: I think this line accidentally got split.

```suggestion
	Main_OnCommand(command->gaccel.accel.cmd, 0);
```

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
 	outputMessage(format(translate("marker {} inserted"), number));
 }
 
+void cmdInsertProjectMarker(Command* command) {
+	cmdhInsertProjectMarker(command->gaccel.accel.cmd);
+}
+
+void cmdhInsertTakeMarker(int command) {
+	if(CountSelectedMediaItems(0)>0) {
```
jcsteh, line 4819: If there are no selected items, this means we will never pass the command to REAPER. Practically, this should be fine, since REAPER will probably do nothing in the case of 0 selected items. However, when intercepting commands, I do think we should always pass them to REAPER, on the very slight chance that REAPER does something we don't expect, either now or in future. Otherwise, the action will just do nothing, even if it would have done something without OSARA installed.

You could either:

1. Check if the item count is 0, call Main_OnCommand and return; or
2. Don't check at all and rely on the fact that 0 items means the loops will do nothing.

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+		vector<int> preMarkers(CountSelectedMediaItems(nullptr));
```
jcsteh, line 4820: Do we need the vector here? Can we just add up all the take markers across all selected items now and then after and compare them, rather than comparing the counts for specific items? Am I missing something?

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+		for(int i = 0; i < preMarkers.size(); ++ i) {
+			MediaItem* item = GetSelectedMediaItem(nullptr, i);
+			MediaItem_Take* take = GetActiveTake(item);
+			if (preMarkers[i] < GetNumTakeMarkers(take)) {
+				++ addedMarkers;
+			}
+		}
+		if (addedMarkers == 0) {
+			return; // not inserted
+		} else {
```
jcsteh, line 4837: There's no need for an else block if you always return in the if above it. This also simplifies things by avoiding a level of nesting.

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+			// Translators: Reported when using REAPER's quick add take marker action. If more than one take marker is inserted, [] will be replaced with the number of markers. E.G. "2 take markers inserted".
```
jcsteh, line 4838: `[]` should be `{}`

```suggestion
			// Translators: Reported when using REAPER's quick add take marker action. If more than one take marker is inserted, {} will be replaced with the number of markers. E.G. "2 take markers inserted".
```

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+			outputMessage(format(
+				translate_plural("take marker inserted", "{} take markers inserted", addedMarkers), addedMarkers));
+			return;
```
jcsteh, line 4841: There's no need for a return if it's the last thing in the function and the function returns nothing.

```suggestion
```

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+		}
+	}
+}
+
+void cmdInsertTakeMarker(Command* command) {
+	cmdhInsertTakeMarker(command->gaccel.accel.cmd);
+}
+
+void cmdInsertProjectOrTakeMarker(Command* command) {
+	if  (fakeFocus == FOCUS_ITEM && CountSelectedMediaItems(nullptr) > 0)
```
jcsteh, line 4851: How often will fakeFocus be item with 0 items selected? I'm just wondering whether we should only look at fakeFocus and not the selected item count. That will mean we do nothing if no items are selected, but it's consistent with other commands I think? Also, we don't check the selection count for edit marker below.

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+		cmdhInsertTakeMarker(42390); // Item: Quick add take marker at play position or edit cursor
+	else
+		cmdhInsertProjectMarker(40157); // Markers: Insert marker at current position
+}
+
+void cmdInsertOrEditMarker(Command* command) {
+	double start, end;
+	GetSet_LoopTimeRange(false, true, &start, &end, false);
+	if (start != end && fakeFocus == FOCUS_ITEM) {
+		Main_OnCommand(43181, 0); // Item: Add/edit take marker at time selection
```
jcsteh, line 4861: The action is called "OSARA: Add project or take marker at cursor (depending on focus)". It doesn't mention time selection. Should it?

src/reaper_osara.cpp
```diff
@@ -4810,6 +4811,62 @@ void cmdInsertMarker(Command* command) {
...
+	} else if (fakeFocus == FOCUS_ITEM) {
+		Main_OnCommand(42385, 0); // Item: Add/edit take marker at play position or edit cursor
+	} else {
+		Main_OnCommand(40171, 0); // Markers: Insert and/or edit marker at current position
+		return;
```
jcsteh, line 4866: Unnecessary return.

```suggestion
```

jcsteh (changes requested):
Thanks for working on this. Sorry it's taken me so long to get to a second round of review.

src/reaper_osara.cpp
ScottChesworth: I've removed the vector, you're right that it wasn't needed with this approach. I went in more on checking for selected items alongside fakeFocus because this approach keeps the project markers feeling like the primary function of the keystroke, important because those have higher adoption.

src/reaper_osara.cpp
ScottChesworth: Hmm well it isn't difficult to get that fakeFocus item with 0 items selected situation, it can often be the case after an item is deleted. Now in that situation we go back to the action inserting project markers until the user explicitly selects another item.

src/reaper_osara.cpp
ScottChesworth: Urgh, just realised that I renamed the wrong action. Time selection is only relevant when using the add/edit one. Another commit coming up shortly.

src/reaper_osara.cpp
```diff
@@ -4810,6 +4810,56 @@ void cmdInsertMarker(Command* command) {
...
+void cmdInsertProjectMarker(Command* command) {
+	cmdhInsertProjectMarker(command->gaccel.accel.cmd);
+}
+
+void cmdhInsertTakeMarker(int command) {
+	int preMarkers = 0;
+	for(int i = 0; i < CountSelectedMediaItems(nullptr); ++ i) {
+		MediaItem* item = GetSelectedMediaItem(nullptr, i);
+		MediaItem_Take* take = GetActiveTake(item);
+		preMarkers += GetNumTakeMarkers(take);
```
jcsteh, line 4822: Sorry, I missed this in the prior review. It's possible for an item to be empty and contain no takes, so you need to only do this if take is valid. Otherwise, we will possibly crash.

```suggestion
		if (take) {
			preMarkers += GetNumTakeMarkers(take);
		}
```

src/reaper_osara.cpp
```diff
@@ -4810,6 +4810,56 @@ void cmdInsertMarker(Command* command) {
...
+	}
+	Main_OnCommand(command, 0);
+	int postMarkers = 0;
+	for(int i = 0; i < CountSelectedMediaItems(nullptr); ++ i) {
+		MediaItem* item = GetSelectedMediaItem(nullptr, i);
+		MediaItem_Take* take = GetActiveTake(item);
+		postMarkers += GetNumTakeMarkers(take);
```
jcsteh, line 4829: Ditto.

src/reaper_osara.cpp
```diff
@@ -5252,6 +5303,8 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Check for update")}, "OSARA_UPDATE", cmdCheckForUpdate},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Open online documentation")}, "OSARA_OPENDOC", cmdOpenDoc},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report tempo and time signature at play cursor; press twice to add/edit tempo markers")}, "OSARA_MANAGETEMPOTIMESIGMARKERS", cmdManageTempoTimeSigMarkers},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Add project or take marker at cursor (depending on focus)")}, "OSARA_ADDPROJTAKEMARKER", cmdInsertProjectOrTakeMarker},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Add/edit project or take marker at cursor (depending on focus) sets time selection as take marker length")}, "OSARA_ADDEDITPROJTAKEMARKER", cmdInsertOrEditMarker},
```
jcsteh, line 5307: This is bothering me slightly linguistically. Perhaps this:
OSARA: Add/edit project or take marker at cursor (depending on focus), sets time selection as take marker length
or this:
OSARA: Add/edit project or take marker at cursor (depending on focus) and set time selection as take marker length

jcsteh (commented):
Thanks. Approving this for merge once these comments are addressed.

Timtam:
Just FYI @jcsteh, we tested explicitly with items that don't have takes and didn't cause a crash, probably because REAPER is smart enough to just return 0 if you hand it an invalid take. Its fair to ensure we don't cause unexpected behaviour, but REAPER is smart enough to don't crash too. An extra safety layer doesn't hurt though.

jcsteh:
Yeah, I did wonder whether REAPER might be smart enough to handle that. I'm fairly sure I've seen a crash by my failure to null check takes in the past, but maybe that wasn't a REAPEr function I was calling; I can't recall.

If you've explicitly tested it and it doesn't crash, I can live without that change.

jcsteh:
The commit I'm talking about is 8944a12bc3be0d16ba3d68c1a4fb89c1daa30b05, but maybe that was caused by something else I was doing with a null take there. My commit comment is a bit vague on the details of the actual crash. :)

jcsteh (approved):
Durp, meant to actually hit approve on this. Feel free to merge once you've addressed the comments one way or another, even if that's an intentional decision to ignore the null check guidance given that you tested it and it works fine.

RDMurray:
@ScottChesworth it's been a while, is it still planned to merge this?

LeonarddeR:
@ScottChesworth Any chance you can fix this up?

PR #1141, opened 2024-07-12
Fix Mac key map regression.

ScottChesworth:
An older version was accidentally included in the prior snapshot.


ScottChesworth:
Sorry @jcsteh, the Mac key map got messed up in yesterday's snapshot and somehow I didn't notice the diff was weird. Have checked through everything recent with Jen on Mac to confirm for certain that this PR restores parity.

PR #1144, opened 2024-08-05
Make the cmdMaybeAddRippleMessage function report actual ripple state…

ScottChesworth:
… instead of a generic message


ScottChesworth:
I've added specific reporting for each ripple mode being on because it was pointed out on a Q&A that although the ripple toolbar button doesn't make any visual distinction between ripple per-track or ripple all tracks being enabled, sighted users get confirmation of which ripple mode was active by seeing which items move. Our previous catchall "ripple is on" report didn't provide an equivalent for the visual feedback.

PR #1145, opened 2024-08-08
Add feedback for actions that invert selection of envelope points, it…

ScottChesworth:
…ems and tracks


ptorpey:
Man, you are quicker than greased lightening!  Thanks for this.  Will use it tomorrow.

 

--Pete

PR #1150, opened 2024-08-27
Improve feedback stretch markers

ScottChesworth:
Make OSARA report the addition of stretch markers when Report time movement during playback/recording is enabled in Config. Requested on RWP.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3835,11 +3835,13 @@ void cmdhAddOrRemoveStretch(int command) {
 	int difference = newCount - oldCount;
 	if (newCount >= oldCount) {
 		if (shouldReportTimeMovement()) {
+		// Translators: Reported when one or more stretch markers are added. {} will be replaced by the number of stretch markers, EG "2 stretch markers added".
 			outputMessage(format(
 				translate_plural("{} stretch marker added", "{} stretch markers added", difference),
 				difference));
 		}
 	} else {
+		// Translators: Reported when one or more stretch markers are removed. {} will be replaced by the number of stretch markers, EG "2 stretch markers removed".
 		outputMessage(format(
 			translate_plural("{} stretch marker removed", "{} stretch markers removed", -difference),
 			-difference));
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3809,17 +3809,45 @@ void cmdDeleteTimeSig(Command* command) {
 		outputMessage(translate("time signature deleted"));
 }
 
-void cmdRemoveStretch(Command* command) {
-	MediaItem* item = GetSelectedMediaItem(0, 0);
-	if (!item)
+void cmdhAddOrRemoveStretch(int command) {
+	const int itemCount = CountSelectedMediaItems(nullptr);
+	if (itemCount == 0) {
+		outputMessage(translate("no selected items"));
 		return;
-	MediaItem_Take* take = GetActiveTake(item);
-	if (!take)
-		return;
-	int count = GetTakeNumStretchMarkers(take);
-	Main_OnCommand(41859, 0); // Item: remove stretch marker at current position
-	if (GetTakeNumStretchMarkers(take) != count)
-		outputMessage(translate("stretch marker deleted"));
+	}
+	int oldCount = 0;
+	for (int i = 0; i < itemCount; ++i) {
+		MediaItem* item = GetSelectedMediaItem(nullptr, i);
+		MediaItem_Take* take = GetActiveTake(item);
+		if (!take)
+			continue;
+		oldCount += GetTakeNumStretchMarkers(take);
+	}
+	Main_OnCommand(command, 0);
+	int newCount = 0;
+	for (int i = 0; i < itemCount; ++i) {
+		MediaItem* item = GetSelectedMediaItem(nullptr, i);
+		MediaItem_Take* take = GetActiveTake(item);
+		if (!take)
+			continue;
+		newCount += GetTakeNumStretchMarkers(take);
+	}
+	int difference = newCount - oldCount;
+	if (newCount >= oldCount) {
+		if (shouldReportTimeMovement()) {
+			outputMessage(format(
+				translate_plural("{} stretch marker added", "{} stretch markers added", difference),
+				difference));
+		}
+	} else {
+		outputMessage(format(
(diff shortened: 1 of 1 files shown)


src/reaper_osara.cpp
```diff
@@ -3809,17 +3809,47 @@ void cmdDeleteTimeSig(Command* command) {
...
+		outputMessage(translate("no selected items"));
 		return;
-	int count = GetTakeNumStretchMarkers(take);
-	Main_OnCommand(41859, 0); // Item: remove stretch marker at current position
-	if (GetTakeNumStretchMarkers(take) != count)
-		outputMessage(translate("stretch marker deleted"));
+	}
+	int oldCount = 0;
+	for (int i = 0; i < itemCount; ++i) {
+		MediaItem* item = GetSelectedMediaItem(nullptr, i);
```
jcsteh, line 3820: Just to clarify, these commands definitely insert stretch markers across all selected items? I guess that makes sense and we just never supported this properly, but I just wanted to check.

src/reaper_osara.cpp
```diff
@@ -3809,17 +3809,47 @@ void cmdDeleteTimeSig(Command* command) {
...
+	if (newCount >= oldCount) {
+		if (shouldReportTimeMovement()) {
+		// Translators: Reported when one or more stretch markers are added. {} will be replaced by the number of stretch markers, EG "2 stretch markers added".
+			outputMessage(format(
+				translate_plural("{} stretch marker added", "{} stretch markers added", difference),
+				difference));
+		}
+	} else {
+		// Translators: Reported when one or more stretch markers are removed. {} will be replaced by the number of stretch markers, EG "2 stretch markers removed".
+		outputMessage(format(
```
jcsteh, line 3845: Is it intentional that we respect shouldReportTimeMovement for addition, but not removal? I guess I kinda get this - you might well add them while playing, but you're less likely to remove them while playing, since you have to move to the marker first anyway - but I wanted to check because it does feel asymmetric.

src/reaper_osara.cpp
```diff
@@ -4273,6 +4303,7 @@ void cmdRemoveFocus(Command* command) {
 			cmdhRemoveItems(40006); // Item: Remove items
 			break;
 		case FOCUS_MARKER:
+
```
jcsteh, line 4306: Extraneous blank line.

src/reaper_osara.cpp
```diff
@@ -4282,7 +4313,7 @@ void cmdRemoveFocus(Command* command) {
 			cmdDeleteTimeSig(nullptr);
 			break;
 		case FOCUS_STRETCH:
-			cmdRemoveStretch(nullptr);
+			cmdhAddOrRemoveStretch(41859);
```
jcsteh, line 4316: Please comment this command id.

src/reaper_osara.cpp
```diff
@@ -5107,7 +5138,10 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {{0, 0, 40613}, nullptr}, nullptr, cmdDeleteMarker}, // Markers: Delete marker near cursor
 	{MAIN_SECTION, {{0, 0, 40615}, nullptr}, nullptr, cmdDeleteRegion}, // Markers: Delete region near cursor
 	{MAIN_SECTION, {{0, 0, 40617}, nullptr}, nullptr, cmdDeleteTimeSig}, // Markers: Delete time signature marker near cursor
-	{MAIN_SECTION, {{0, 0, 41859}, nullptr}, nullptr, cmdRemoveStretch}, // Item: remove stretch marker at current position
+	{MAIN_SECTION, {{0, 0, 41842}, nullptr}, nullptr, cmdAddOrRemoveStretch}, // Item: Add stretch marker at cursor
+	{MAIN_SECTION, {{0, 0, 41859}, nullptr}, nullptr, cmdAddOrRemoveStretch}, // Item: remove stretch marker at current position
+	{MAIN_SECTION, {{0, 0, 41844}, nullptr}, nullptr, cmdAddOrRemoveStretch}, // Item: Remove all stretch markers
```
jcsteh, line 5143: Do either of these need a readme update?

jcsteh (changes requested):

src/reaper_osara.cpp
ScottChesworth: Yep REAPER always did that, our reporting didn't used to include the number of selected items or speak when no item was selected.

src/reaper_osara.cpp
ScottChesworth: I left it asymmetric because I think it makes sense for deletion to always speak. We do that with items already.

src/reaper_osara.cpp
ScottChesworth: Yep they did and I've done it

ScottChesworth:
Bringing this in line with um... 2 years of hindsight lol... It seems strange that reporting of stretch marker, project marker and region insertion is gated with the time movement during playback cconfig. They're not time movement. They are all a type of marker or region, so why not tie those to "Report markers and regions during playback" instead?

jcsteh:
I could swear we had this conversation fairly recently, but maybe not?

I'm totally willing to grant that maybe the settings are poorly named. However, I do think that tying marker insertion to the current "Report markers and regions during playback" setting is problematic.

As it currently stands, "Report markers and regions during playback" controls whether markers automatically report during playback. That is, without any active interaction at all, just letting the audio play, you will get notified about markers as you pass them. In contrast, "Report time movement during playback/recording" is about whether active interaction causes certain things to be reported.

I can see someone wanting to know about markers they are actively inserting during playback, but not wanting to have chatter every time they passively pass a marker when they didn't interact. I think the flip side is also plausible.

The fact that this has come up again (I think) suggests we do need to rename these somehow.

jcsteh:
Alternative names (not necessarily good ones, but at least this should illustrate my point):

1. Report markers and regions during playback -> Report markers and regions as the play cursor passes them
2. Report time movement during playback/recording -> Report time movement and marker insertion during playback/recording

I'm still not quite happy with the second one. I'm trying to explain why I conceptually see time movement and marker insertion as similar. The point is that these are things you commonly due during playback, and because it's so common and natural to do them during playback, you don't want action feedback to distract you from playback.

jcsteh:
Or "Report cursor movement and marker insertion during playback/recording"?

PR #1151, opened 2024-08-29
speak passing start and end of time selection when Report time moveme…

ScottChesworth:
…nt during playback/recording is checked in OSARA config

### DIFF
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -91,7 +91,7 @@ class Surface: public IReaperControlSurface {
 			if (settings::reportMarkersWhilePlaying) {
 				this->reportMarker(playPos);
 			}
-			if (settings::reportTimeMovementWhilePlaying) {
+			if (settings::reportTimeSelectionWhilePlaying) {
 				this->reportTimeSelectionWhilePlaying(playPos);
 			}
 		}
@@ -423,7 +423,7 @@ class Surface: public IReaperControlSurface {
 	double lastPlayPos = 0;
 	int lastMarker = -1;
 	int lastRegion = -1;
-bool hasReportedTimeSelection = false;
+	bool hasReportedTimeSelection = false;
 };
 
 IReaperControlSurface* createSurface() {
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -362,10 +362,11 @@ class Surface: public IReaperControlSurface {
 	void reportTimeSelectionWhilePlaying(double playPos) {
 		double start, end;
 		GetSet_LoopTimeRange(false, false, &start, &end, false);
+		if (start == end) {
+			return;
+		}
 		double startDiff = playPos - start;
 		double endDiff = playPos - end;
-		if (start == end)
-			return;
 		if (startDiff >= 0 && startDiff <= 0.1) {
 			if (this -> hasReportedTimeSelection)
 				return;
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -82,13 +82,18 @@ class Surface: public IReaperControlSurface {
 	}
 
 	void Run() final {
-		if (settings::reportMarkersWhilePlaying && GetPlayState() & 1) {
+		if (GetPlayState() & 1) {
 			double playPos = GetPlayPosition();
 			if (playPos == this->lastPlayPos) {
 				return;
 			}
 			this->lastPlayPos = playPos;
-			this->reportMarker(playPos);
+			if (settings::reportMarkersWhilePlaying) {
+				this->reportMarker(playPos);
+			}
+			if (settings::reportTimeMovementWhilePlaying) {
+				this->reportTimeSelectionWhilePlaying(playPos);
+			}
 		}
 		this->reportInputMidiNote();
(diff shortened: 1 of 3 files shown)


ScottChesworth:
Implemented this request that came in via WhatsApp. The steps to have REAPER automatically stop at the end of a time selection aren't obvious, so when the appropriate option in OSARA config is checked (which it is by default) users will be told when they pass the start and end. I used existing strings so no additional work needed from translators.

ScottChesworth:
Also dabbled in a file I've never been in before and I would've been lost without @TimTam holding my hand hahaha, hence the review.

src/controlSurface.cpp
```diff
@@ -82,13 +82,18 @@ class Surface: public IReaperControlSurface {
...
 			double playPos = GetPlayPosition();
 			if (playPos == this->lastPlayPos) {
 				return;
 			}
 			this->lastPlayPos = playPos;
-			this->reportMarker(playPos);
+			if (settings::reportMarkersWhilePlaying) {
+				this->reportMarker(playPos);
+			}
+			if (settings::reportTimeMovementWhilePlaying) {
```
jcsteh, line 94: I'm not sure about tying this to this particular setting. This setting is currently about on-demand speech; i.e. it determines whether certain explicit actions report when playing. In contrast, this check is about whether we report something automatically even without an explicit action. In that sense, this is more like the report markers during playback option. Those two things are quite different.

Happy to be convinced otherwise, but I do think it needs convincing. I'd personally be more inclined to tie this to the markers option and rename that option to cover time selection.

src/controlSurface.cpp
```diff
@@ -396,6 +423,7 @@ class Surface: public IReaperControlSurface {
 	double lastPlayPos = 0;
 	int lastMarker = -1;
 	int lastRegion = -1;
+bool hasReportedTimeSelection = false;
```
jcsteh, line 426: Missing indent.

jcsteh (changes requested):

src/controlSurface.cpp
ScottChesworth: Ah ok, hadn't considered the difference between on demand and automatic reporting. I could create a new toggle just for time selection? Feels like users would be opting in for what could be quite a lot of additional speech if we tie it in with markers/regions, also given how often markers are used for navigation, I think there's a high likelihood of a marker/region and one edge of a time selection to be in the same position.

jcsteh: I'm not dead against it, but I would like to understand how users will perceive these two things as inherently related. My concern is that some folks might want to know where they are if they move during playback, but not have their speech nattering at them if they don't particularly care about their selection right now. If you really think this isn't going to bother users - you'd probably know more than I would given you deal with users more often :) - perhaps we could rename this option to "Report time movement/selection during playback/recording". Otherwise, I guess we could add a new setting, though it's unfortunate that we'd have an entirely new setting covering a very narrow, single use case.

ScottChesworth: Using it in situ today, I'm not convinced about it being tied to either of our existing options. Reckon I should make a new one called "Report time selection/loop points during playback/recording". We only report time selection if those are linked. Time selection is fundamental functionality, so to me it doesn't seem problematic that it has its own option. Any objections?

jcsteh: No real objections, aside from a general reluctance to add even more settings, but that's not a useful objection. :)

I'm not following the bit about loop points though. If I'm reading the code correctly, this only looks at time selection. Loop points don't come into it at all.

ScottChesworth: Yeah I haven't written the loop points stuff yet. It's an easy addition though. Figured they're so closely related, the new option may as well cover those too.

ScottChesworth:
Have added a new setting. I know the narrow scope isn't ideal, adding loop points in felt too verbose, I'll revisit if that's requested with good use cases.
Question: is there anything I need to do to make the new setting in OSARA Config translatable or does that happen automatically?

src/controlSurface.cpp
```diff
@@ -354,6 +359,28 @@ class Surface: public IReaperControlSurface {
 		}
 	}
 
+	void reportTimeSelectionWhilePlaying(double playPos) {
+		double start, end;
+		GetSet_LoopTimeRange(false, false, &start, &end, false);
+		double startDiff = playPos - start;
+		double endDiff = playPos - end;
+		if (start == end)
+			return;
```
jcsteh, line 368: nit: Avoid unnecessary work if we return early.

```suggestion
		if (start == end) {
			return;
		}
		double startDiff = playPos - start;
		double endDiff = playPos - end;
```

jcsteh (approved):
Thanks. Sorry it's taking me an age to get to reviews at the moment. My brain is fried and I seem to be barely able to do my day job at the moment, let alone anything else.

You don't need to do anything special to make the setting display strings translatable. The build system has magic which processes settings.h and builds it into a GUI resource file (site_scons/makeConfigRc.py). The GUI resource file is then processed by the translation template magic (site_scons/makePot.py), which looks for translatable strings and adds them to the translation template.

PR #1162, opened 2024-10-18
Add OSARA: Report length of last touched item, include item length wh…

ScottChesworth:
…en reporting selection with Control_Shift+Space

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -825,6 +825,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Toggle global automation override between latch preview and off: Control+Alt+Shift+L
 - OSARA: Cycle through midi recording modes of selected tracks: Alt+Shift+\
 - OSARA: Report groups for current track
+- OSARA: Report length of last touched item
 - OSARA: Report regions, last project marker and items on selected tracks at current position: Control+Shift+R
  - Pressing this twice will display the information in a dialog with a text box for easy review.
 - OSARA: About: Control+F1
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4516,6 +4516,18 @@ void cmdReportCursorPosition(Command* command) {
 	outputMessage(s);
 }
 
+void cmdReportItemLength(Command* command) {
+	MediaItem* item = getItemWithFocus();
+	if (!item) {
+		outputMessage(translate("no items selected"));
+		return;
+	}
+	double start = GetMediaItemInfo_Value(item, "D_POSITION");
+	double length = GetMediaItemInfo_Value(item, "D_LENGTH");
+	double end = start + length;
+	outputMessage(formatLength(start, end, TF_RULER, FT_NO_CACHE, false));
+}
+
 void reportCursorPositionPrimaryFormat() {
 	// This function can be called when only reporting the primary ruler format is needed.
 	TimeFormat tf = TF_RULER;
@@ -5278,6 +5290,7 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Remove items/tracks/contents of time selection/markers/envelope points (depending on focus)")}, "OSARA_REMOVE", cmdRemoveFocus},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Toggle shortcut help")}, "OSARA_SHORTCUTHELP", cmdShortcutHelp},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report edit/play cursor position, transport state and nearest markers and regions")}, "OSARA_CURSORPOS", cmdReportCursorPosition},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report length of last touched item")}, "OSARA_REPORTLENGTHSELITEM", cmdReportItemLength},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Enable noncontiguous selection/toggle selection of current track/item (depending on focus)")}, "OSARA_TOGGLESEL", cmdToggleSelection},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Move last focused stretch marker to current edit cursor position")}, "OSARA_MOVESTRETCH", cmdMoveStretch},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report level in peak dB at play cursor for channel 1 of current track (reports input level instead when track is armed)")}, "OSARA_REPORTPEAKCURRENTC1", cmdReportPeakCurrentC1},


jcsteh:
This seems like it significantly increases verbosity when reporting selected items, particularly if there are a lot of them. Why is length so important as to justify that? And why length vs start position?

ScottChesworth:
I added the action to report length because we get a ton of support requests asking how to identify stray item tails on areas of the timeline where users have done multiple recordings without a strategy for keeping them tidy.
Added it to the selection reports just because that seemed consistent with reporting ruler selection, but that can be taken out easy enough if you don't dig it. My thinking was that if there are a lot of items selected, you'd be in the multiline, where it's easy to interrupt length if you don't need to hear about it.

jcsteh:
With ruler selection, we also report the start and end. We're not reporting that for items. Should we? Otherwise, I don't follow the consistency argument.

I wonder if it's worth reporting item length for multi line (double press), but not for single press? That's inconsistent with other things we've done, but it does make a certain amount of sense from a quick vs interactive perspective.

ScottChesworth:
> With ruler selection, we also report the start and end. We're not reporting that for items. Should we?
> 
Considered that and decided not to. Reporting ruler selection we jump straight into start end and length, with item selection there's the take name first. I wouldn't sit through take name, start and end just to hear length, could probably go to the end of the item and do the maths quicker in my noggin.

> I wonder if it's worth reporting item length for multi line (double press), but not for single press? That's inconsistent with other things we've done, but it does make a certain amount of sense from a quick vs interactive perspective.
> 
Sure, that could work. *gulp* don't think I know how...

jcsteh:
> I wouldn't sit through take name, start and end just to hear length, could probably go to the end of the item and do the maths quicker in my noggin.

I'm really not getting this. Why is length important in the context of selection? For ruler, it's kinda important because, well, you're dealing with time. But items aren't primarily about time for selection. Or if they were, the start and end would also be important. What am I missing here that makes length special but start and end irrelevant?

jcsteh:
Let me ask this another way: can you give me a real world scenario where hearing length for item selection is useful, but you don't care about start and end?

ScottChesworth:
> can you give me a real world scenario where hearing length for item selection is useful, but you don't care about start and end?
> 
Yeah, like I said in the first reply, length can identify stray item tails on areas of the timeline where users have done multiple recordings before they know about strategies to keep items tidy. I often see knew users getting themselves tied up in knots because the more recordings they're making, the more fiddly diddly short tails they're ending up with. On RWP, we recently had someone who could hear those overlapping tails and thought the default crossfading was a problem with their gear.

jcsteh:
Right, I saw that, but I thought that was a scenario for reporting item length, not for reporting length in selection. Do you think the same still applies for selection? I'm not quite following that workflow.

ScottChesworth:
How about this to keep verbosity under control. I can add a Config option, "When moving through items, report their length instead of name". That way you can choose what's important to you at the time and you've got a reasonable amount of verbeage with each of the choices.

jcsteh:
Wow. We're... really on very different planets here, haha.

This PR adds length reporting in two cases:

1. A dedicated command to report length. I don't fully understand the use case here, but I follow that this command solves that scenario and that seems reasonable enough.
2. Item length when reporting selection with control+shift+space. I don't understand how the same use case fits here. If someone is trying to tidy up item tails or whatever, why would they be repeatedly using report selection to do that?

jcsteh:
> I can add a Config option, "When moving through items, report their length instead of name".

And this is a third place we'd report length which isn't covered by this PR: when moving through items with control+left/rightArrow. I'm not following how all of this ties together, nor why we need to do this in three separate places.

jennykbrennan:
I can only comment from my own workflow obviously, but I see how this could have saved me much time. With editing projects with close to three hours of spoken word across five tracks, mostly editing with ripple all tracks enabled, with item reported as 1162 20231014-201016 #0 ScottChesworth.flac 144 min 3.653, and 1100+ items on that track alone, an option to hear the length of that item instead of, or possibly before,  take name, would make editing faster  and easier for me. 
Certainly, i would be quite happy with the report action alone.  It would speed up my work in its own right, but that is another keystroke to contend with.
I think this is all about users having choices in how they work. For me, length of item is critical for how I work and start and end of those items don't matter. I frequently have to figure out what stays and what goes in a project like this and I often find myself checking selection, while being careful to maintain my position in the project, wishing it would tell me more about what i have selected, a minute of speech or half a second of silence that can be removed. Yes, I can do as Scott says, go to the end and do the math in my noggen, but honestly, if I have to focus on doing math, simple as it may be, I'm not focusing on what I should be doing, making Scott sound like he knows what he's talking about. :D 
The double press selection report is not something I use, so I can't comment on whether that is useful or not.

ScottChesworth:
> making Scott sound like he knows what he's talking about.
> 
Dirty work, but someone's gotta do it...

ScottChesworth:
I'm taking length out of the selection reports. Have thought of a way to custom action the reporting that people have requested, that'll do for now.

jcsteh:
Okay. Let's get this merged and we can then figure out other options. I'm not 100% against the option of reporting it with selection; I just need to get a better handle on the use cases.

jcsteh:
> I think this is all about users having choices in how they work.

Sure, but this is a very slippery slope and IMO it's not enough to justify a new thing in and of itself. Each time we add an option like this, we need to be able to justify that it isn't specific to one user's very specific workflow. We could use this argument to say that we should also have an option to report the item end time, or the number of stretch markers in the take, or the number of envelopes on a track, or the number of MIDI notes in an item, etc.; it goes on forever.

> For me, length of item is critical for how I work and start and end of those items don't matter.

Regardless of whether I personally get it (I'm still struggling to follow the workflow, haha), I'm starting to get the impression that this is important to quite a few people, so that's great and gets us into "this is probably a good thing to have". The question then is: where is it most useful? Based on the use cases described here, it sounds to me like it'd be most useful when moving to items, rather than when reporting selection.

> The double press selection report is not something I use,

Ah, so that verbosity reduction trick wouldn't have worked for you, since the idea was to not report the length for a single press.

GarthHum:
I would like to have this config option to report length rather than item name when moving through items. 
For my own workflows with podcast editing the ideal would be item number and length, reported when navigating through items. It is very rare that I want to hear the item name, and would prefer to just use the report selection to get this if required.
I’ve just been playing with the dedicated action to report last touched item length, and it works well enough , when put in to a custom action with the Select and move to next item, action. 
Thanks for the work an thought put in to everything you all do.

jennykbrennan:
Tricky one. We do different kinds of work in Reaper so it makes sense the perspectives are different.  I have a difficult time understanding how it's not important. lol. Different planets and all that.
My issue with the double press is that I have to scroll past information I don't need. I already know I have no time selection or track selection when i do what I do.  I would like to have the single press give me the item selected and the time of that item. I don't expect to persuade anyone that it's something they should want.  The argument that changing selection should be enough doesn't quite do it for me. As I mentioned, I want to maintain position in the project and quickly find out what I have going on, without changing selection.
Again, it's what I have found I'm missing when working.  Now, with this new action it's already so much easier. I'm using the custom action to speak length instead of item name for now. 
An option would be nicer. Having it added to single press report selection would be the cherry on top.

The number of emails we get these days from people who are starting podcasts and recording audio books makes me think that there might be a bigger use case than hunting down tiny stray items for beginners. 
Again, my perspective :).

MatejGolian:
Don't know if it's related to the things discussed here in any way, but when looking at reaperaccessibility.com, the latest 'public' build for MAC is from Thursday while the last one for windows is from Friday.

jcsteh:
> The argument that changing selection should be enough doesn't quite do it for me. As I mentioned, I want to maintain position in the project and quickly find out what I have going on, without changing selection.

Wait... so you only want it reported when reporting selection, not when moving to the item? Or you want both?

> I'm using the custom action to speak length instead of item name for now.

Can you clarify what the custom action does?

jennykbrennan:
I would want both. Greedy as I am, I want to hear the length in either case. I suspect that I can only find out with time if it's as useful as I think. A sample size of 1 isn't convincing, even to the sample.  

the custom action does
OSARA: Mute next message from OSARA  1 of 3
Item navigation: Select and move to next item  2 of 3
OSARA: Report length of last touched item  3 of 3

GarthHum:
That is basically the same custom action I setup however found that the first action, mute next action from osara,  isn’t required as the feedback from report last item length action immediately interrupts the feedback from  move to next item - at least on my system. 
I tried adding a wait 0.5 sec in-between move to next item and report last touched length so that there was time for the item number to read but not the whole name. So very klugey 
I only mention as it was an attempt to get what I’d really like which ist to be able to optionally swap out the item name reporting with item length, or even better, neither so just the item number reads.  
Two tick boxes to customise when moving to next/prev item, report item name, and report item length, so both, neither or either could be reported, would be amazing.  
For my workflows, I would swap between just item number and item  number + length, depending on the phase of editing. 
For me, the item name is almost never of interest when editing long form dialog.
I get that the current reporting has the item number first and I should be able to ignore the name or move on quickly to the next item, if the rest of the name isn’t of interest, but especially when the item name also includes lots of numbers, it just becomes more verbiage that I’d prefer not to hear.

jcsteh:
Just spitballing here and not making any promises or commitments whatsoever, but...

How well do you think folks would cope if, rather than trying to design some complex UI with movable check boxes or whatever, we had a template string for track and item reports? For example, for items, the default would look something like this:

{number} {muted} {locked} {takeName} {start} {fx}

You could customise that to:

{number} {muted} {locked} {length} {start} {fx}

or even just:

{number} {muted} {locked}

{fx} is a bit of a problem, since we really need to separate that with a semi-colon and an "FX: " prefix, but we don't want to do that if there are no effects. And I really don't want some wild templating language with ${if} or something like that; it's just too hard to maintain and hard for users to grasp. But I think that's the only real edge case here and we can probably just have {fx} include "; FX: " at the start.

jcsteh:
One caveat with the template string approach is that it conflicts somewhat with two check boxes we have: Report track numbers, Report FX when moving to tracks/takes. It'd be best to get rid of those, but if we do, that's going to break users who rely on them.

jcsteh:
The template string approach also doesn't solve the selection reporting use case requested here. Hmm.

GarthHum:
No it may not but these seem like two separate concepts entirely to me - sorry I may not be understanding. 
I just mean that I’m moving between items all the time and only occasionally getting the report for selection when I want to find out additional info.
I wouldn’t really want the verbosity for navigation to be tied to that of querying a selection. In my little corner of the world navigation reporting should be minimal and querying a selection can be as verbose as you like :)

jcsteh:
Right. I was responding to the request that the length should also be included with the report selection command, but since that won't follow the navigation template, length wouldn't be included there, and including length but not start time or whatever else feels harder to rationalise since it couldn't be configurable with the navigation template.

PR #1166, opened 2024-11-09
Add report project length action

ScottChesworth:
Added "OSARA: Report project length" action which was requested on RWP, unmapped at the moment. Calling once reports primary ruler format, calling twice quickly reports secondary. Note that during implementation we added an extra parameter to formatLength function in order to support REAPER's project start time offset feature, but it turned out that REAPER itself doesn't respect that when rendering yet. There were circumstances where OSARA would report a different project length to a render with Bounds set to Entire project, so for now I'm not using the extra param because it lead to a confusing UX. The code is still included though, it will be useful when Cockos improve the render behaviour. When that happens, change the last argument to true when calling formatLength.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -449,12 +449,12 @@ string formatNoteLength(double start, double end) {
 }
 
 string formatLength(double start, double end, TimeFormat timeFormat,
-	FormatTimeCacheRequest cache, bool includeZeros) {
+	FormatTimeCacheRequest cache, bool includeZeros, bool includeProjectStartOffset) {
 	timeFormat = getTimeFormat(timeFormat);
 	if(timeFormat != TF_MEASURE && timeFormat != TF_MEASURETICK) {
 		// In all other cases, position and length reports are the same, so
 		// formatTime can handle them.
-		return formatTime(end - start, timeFormat, cache, includeZeros, false);
+		return formatTime(end - start, timeFormat, cache, includeZeros, includeProjectStartOffset);
 	}
 	int startMeasure, startMeasureLength, timeDenom, endMeasure, endMeasureLength;
 	double startBeat = TimeMap2_timeToBeats(nullptr, start, &startMeasure,
@@ -485,7 +485,7 @@ string formatLength(double start, double end, TimeFormat timeFormat,
 	}
 	return formatTimeMeasure(measures, beats, measureLength, timeDenom,
 		timeFormat == TF_MEASURETICK, true, cache == FT_USE_CACHE, includeZeros,
-		false);
+		includeProjectStartOffset);
 } 
 
 string formatCursorPosition(TimeFormat format, FormatTimeCacheRequest cache) {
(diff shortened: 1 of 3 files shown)


src/reaper_osara.cpp
```diff
@@ -449,12 +449,12 @@ string formatNoteLength(double start, double end) {
 }
 
 string formatLength(double start, double end, TimeFormat timeFormat,
-	FormatTimeCacheRequest cache, bool includeZeros) {
+	FormatTimeCacheRequest cache, bool includeZeros, bool includeProjectStartOffset) {
```
jcsteh, line 452: Do we have a reasonable degree of certainty that Cockos will change the render behaviour? Do they acknowledge that it's a bug?

Generally, it's best to try to restrict a pull request for a feature to the changes that are necessary for that feature right now. That said, I can live with this if we think it's likely to change in the near term.

src/reaper_osara.cpp
```diff
@@ -4528,6 +4528,28 @@ void cmdReportItemLength(Command* command) {
 	outputMessage(formatLength(start, end, TF_RULER, FT_NO_CACHE, false));
 }
 
+void cmdReportProjectLength(Command* command) {
+	TimeFormat tf;
+	if (lastCommandRepeatCount == 0) {
```
jcsteh, line 4533: If I'm not mistaken, this is code duplicated from cmdReportCursorPosition. That's a strong indicator that we should split this into a utility function; e.g.
```
// Get the time format to be used for a command where pressing once uses the
// primary ruler unit and pressing twice uses the secondary ruler unit.
TimeFormat getPrimaryOrSecondaryTimeFormatForCommand() {
	if (lastCommandRepeatCount == 0) {
		// Use primary ruler unit.
		return TF_RULER;
	}
	if (GetToggleCommandState(42361)) {
...
```

jcsteh (changes requested):
Thanks.

src/reaper_osara.cpp
jcsteh: It'd be good to get this question answered:

> Do we have a reasonable degree of certainty that Cockos will change the render behaviour? Do they acknowledge that it's a bug?

Otherwise, I think we should remove the new formatLength argument from this patch.

src/reaper_osara.cpp
```diff
@@ -849,6 +849,29 @@ string gridDivisionToFriendlyName(double division) {
 	return translate("grid  unknown");
 }
 
+TimeFormat getPrimaryOrSecondaryTimeFormatForCommand() {
+	TimeFormat tf;
```
jcsteh, line 853: A variable is redundant here because we can just return from the function. Specifically:

1. Remove this variable.
2. Make all the tf = lines into return lines.
3. Remove the else statements, since else is unnecessary after return.
4. Return TF_RULER as a fallback at the bottom, but we should never hit that.

jcsteh (changes requested):

jcsteh (approved):
I applied the changes myself.

jcsteh:
Damn it. Failed to commit the removal of the project offset argument even though I made the change locally.

jcsteh:
Oh wait, no, I did do it properly. 😕

jcsteh:
I split the includeProjectStartOffset argument to formatLength into a branch called formatLengthProjStart so we still have it if/when Cockos change things.

PR #1167, opened 2024-11-21
Report pitch or name when navigating to chords that only contain one …

ScottChesworth:
…note, bound note reporting to the relevant check box in OSARA Config, fixed missing white space when reporting muted notes

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1108,18 +1108,23 @@ void moveToChord(int direction, bool clearSelection=true, bool select=true) {
 	if (cursorMoved && !select && !isNoteSelected(take, chord.first.getIndex())) {
 		s << translate("unselected") << " ";
 	}
-	if (cursorMoved && settings::reportNotes && settings::reportPositionMIDI) {
+	if (cursorMoved && settings::reportNotes) {
 		int count = chord.second - chord.first;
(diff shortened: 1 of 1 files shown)


jcsteh:
Note that this isn't the full info you get when moving with up and down arrows. That also reports length and velocity. Are you happy with this discrepancy? You can still get this by just pressing up or down arrow of course, but it seems odd to make this more convenient for the user but only go half way.

ScottChesworth:
Yep that's intentional. Before this change we'd hear "1 note" from the screen reader, but hearing the preview already tells me there was a single note, so I figured we could speak the note pitch instead without adding verbeage. It'll also be a good change for JAWS/Braille display users, Snowman's scripts already print OSARA's reporting to displays, so now people can hover their finger over where the note is shown as they navigate if needed.

jcsteh:
Okay. Happy to take this change then, but how would you feel about me submitting a follow-up to add velocity and length? My reasoning is that in many cases, I'm working with a MIDI item that mostly has single notes. Now that I get the note name, I can avoid pressing down arrow... but I still have to press down arrow to get the other info, so the advantage of the new convenience isn't fully realised. This would add additional verbiage, which I realise you were trying to avoid, but it's verbiage at the end, not the start, which means it is easily ignored or silenced.

ScottChesworth:
Sure, I'm fine with reporting all of it if that's useful.

PR #1171, opened 2024-11-28
Key map: bound OSARA: Report project length to Control+Win+J, Command…

ScottChesworth:
…+Control+J on Mac


ScottChesworth:
Any idea why the AppVeyor Windows build fails with scons excit code 2 @jcsteh? Local build works as expected so I don't think I've done anything wonky here. I've already tried re-running AppVeyor just in case it was a glitch.

ScottChesworth:
Seems AppVeyor went squiffy for a wee while, builds are working again and I've changed nothing, on account of not knowing what to change.

PR #1180, opened 2024-12-17
Support interception of commands provided by extensions such as SWS

### LINKED ISSUE #1179: How do I intercept an SWS action and call a cmd function?
I'd like to intercept an SWS action. I've found a bunch of them listed in POST_CUSTOM_COMMANDS, they all call post functions, but I need to call a cmd instead because I want to compare some item properties before and after this action runs, aiming to make a more accurate report.

So I tried adding this line to the COMMANDS lookup table:
`{MAIN_SECTION, {DEFACCEL, nullptr}, "_SWS_DEFAULTFADES", cmdResetItemFadesToDefault},`
It builds, but REAPER crashes instantly with that build installed. DOH!

Using an SWS action ID in COMMANDS seems to be what causes REAPER to crash. I verified that by switching to a stock action instead. REAPER opens fine and my function seems to work as well, but there isn't a stock action that has the exact behaviour of this SWS one.

Any advice on how to wangle it appreciated.

jcsteh:
This is done in the COMMANDS array. For example:

```
	{MAIN_SECTION, {DEFACCEL, nullptr}, "_SWS_DEFAULTFADES", cmdResetItemFadesToDefault},
```

Fixes #1179.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -5594,11 +5594,31 @@ void delayedInit() {
 	plugin_register("csurf_inst", (void*)surface);
 	NF_GetSWSTrackNotes = (decltype(NF_GetSWSTrackNotes))plugin_getapi(
 		"NF_GetSWSTrackNotes");
+
+	for (int i = 0; COMMANDS[i].execute; ++i) {
+		if (COMMANDS[i].id) {
+			// This is our own command.
+			if (COMMANDS[i].section == MAIN_SECTION) {
+				COMMANDS[i].gaccel.accel.cmd = plugin_register("command_id", (void*)COMMANDS[i].id);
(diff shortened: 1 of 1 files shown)


jcsteh:
@ScottChesworth, I've tested that this doesn't crash, but I don't know if it actually works. Can you give it a shot with your change on top? This is in the swsCmd branch. If this works, I'll merge this into master and you can then submit a PR with your change. Thanks.

jcsteh:
Note that there's a very slight chance this will break some kind of command or another, so it's worth testing with various kinds of commands: intercepted main commands, intercepted MIDI editor commands, our own main commands, our own MIDI editor commands, etc.

ScottChesworth:
My change works now, thanks! I checked all the types of intercepted commands suggested above, those seem stable too.

PR #1188, opened 2025-01-11
Add feedback for MIDI event mute/unmute

MatejGolian:
I wanted to add feedback when muting/unmuting MIDI events. The reported message should depend on the selection (notes vs CCs vs both). The basic logic seems to work, but I can't figure out how to count muted CCs, because a method completely analogous to how it's done with notes is clearly not possible - at least that's my impression.
@jcsteh could you take a look and perhaps even fix the problem? I think it would take far less time for you to fix it than explain to me what the problem is, but of course I'm not opposed to fixing it myself - I just don't know how.
Many thanks...

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -2295,16 +2295,16 @@ void postToggleFunctionKeysAsStepInput(int command) {
 void postMidiToggleMute(int command) {
 	HWND editor = MIDIEditor_GetActive();
 	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int eventCount = countSelectedEvents(take);
+	if (eventCount == 0) {
+		return;
+	}
 	// Get selected notes.
 	vector<MidiNote> selectedNotes = getSelectedNotes(take);
 	// Get selected CCs.
 	vector<MidiControlChange> selectedCCs = getSelectedCCs(take);
-	int totalNoteCount = selectedNotes.size();
+	int noteCount = selectedNotes.size();
 	int CCCount = selectedCCs.size();
-	int eventCount = selectedNotes.size() + selectedCCs.size();
-	if (eventCount == 0) {
-		return;
-	}
 	ostringstream s;
 	if (CCCount == 0) { // If only notes are selected
 		auto chord = findChord(take, 0, {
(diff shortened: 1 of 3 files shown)


jcsteh:
Thanks for your work. Just FYI, I'm about to go away on holiday for a couple of weeks, so I likely won't be able to look at this until the end of January at the very earliest.

MatejGolian:
I understand. Have a nice holiday then.

MatejGolian:
BTW I have a vague understanding of why it isn't working, but of course not on a level that would help me come up with a solution.
I have a variable called selectedCCs which is a vector. Now when its size is 1 my goal would be to get the 1st CC and check if its muted. The problem is that if I just use something like
auto cc = selectedCCs[0];
or
auto cc = selectedCCs.begin(); ,
checking the value of cc.muted does not work as I would have expected.
On the other hand
auto cc = findCC(take, 0);
returns an object which works, but doing it this way is probably not optimal. So the thing is that I don't get the difference between the objects returned in each scenario.
Next this is the line responsible for counting the total number of muted CCs, but it uses the begin() and end() methods and probably that's the reason for it not working.
			int mutedCount = count_if(selectedCCs.begin(), selectedCCs.end(), [](auto cc) { return cc.muted; });
So hopefully it's not a big issue, but it's not something I'm able to figure out.

RDMurray:
@MatejGolian  the problem seems to be in `getSelectCCs`. It doesn't get the `muted` field. You could replace lines 1001-1005 to use MidiControlChange::get.

Remember that toggle commands with multiple selection in Reaper can get kind of crazy; you could end up with w notes muted, x notes unmuted, y CCs muted and z ccs unmuted. 

See postToggleItemMute.

MatejGolian:
@RDMurray, thanks.
What would the relevant section than look like? Something like this? I don't know how to declare the 3rd parameter that should be passed to MidiControlChange::get correctly, so the first line below is definitely incorrect.
		ReqParams params;
		double position;
		MidiControlChange::get(take, ccIndex, &params);
		position = MIDI_GetProjTimeFromPPQPos(take, params.position);
		ccs.push_back({params.channel, ccIndex, params.msg1, params.msg2, params.msg3, position});

RDMurray:
The third parameter is MidiControlChange::ReqParams, which is a struct of 7 bools used to indicate which params of the event should be requested from Reaper. In this case we probably want everything, something like below (untested):

```c++
		ccs.push_back(MidiControlChange::get(take, ccIndex, {
			true,  // position
			true,  // message1
			true,  // channel
			true,  // message2
			true,  // message3,
			true,  // selected
			true  // muted
		}));
```

Btw what does it mean for a CC to be muted? I'm surprised it's even a thing.

MatejGolian:
Thanks, I'll have a go at it.
Actually I'm not completely sure either of what it does, but I'd assume that it behaves similar to muting notes where the muted note behaves as if it wasn't there at all = it's not being sent out.

MatejGolian:
I've got it. Turns out that the fix was much simpler than it looked like. But thanks @RDMurray, I couldn't have done it without you.
So I guess my work on this is done until Jamie checks it out.

RDMurray:
This is a toggle. It should report when events are unmuted as well.

MatejGolian:
I've added reporting whenever single events become unmuted. I haven't altered the behavior in case toggling mute for multiple events though. The message being reported in the latter case is analogous to the one Osara reports when navigating to chords containing muted notes which seems reasonable IMO and I think that we don't have to be any more verbose than that.

MatejGolian:
Closed by accident.

PR #1189, opened 2025-01-15
Improve feedback for Item: Copy loop of selected area of audio items,…

RDMurray:
… fixes ##1074

@jcsteh this action sometimes pops up an error dialog so I did an ugly CallLater hack to prevent it speaking in that case. Not sure if there's a cleaner way to do that.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3753,9 +3753,8 @@ void cmdRemoveOrCopyAreaOfItems(Command* command) {
 	// the command might show an error, so we need to avoid speaking if the focus changes
 	HWND oldFocus = GetFocus();
 	bool focusChanged = false;
-	auto guard = make_shared<bool>(true); // guard against the lambda accessing out of scope variables
-	auto later = CallLater([&focusChanged, &oldFocus, guard] {
-		if(*guard && GetFocus() != oldFocus) {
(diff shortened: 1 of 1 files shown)


RDMurray:
I haven't tested on Mac yet, but I do have one so I'll test it.

RDMurray:
It is also working on Mac.

jcsteh:
1. What's the error dialog? is there no other way to detect that error condition before it occurs; e.g. by looking at the selection?
2. Is this the only one of those commands that can show an error dialog? In that case, I wonder whether we should split handling of that one into a separate function. Or would there be a huge amount of duplicate code in that case? (I haven't looked closely.)
3. The callLater is a bit ugly, but I can't really think of a better way to handle it. However, please call `later.cancel()` at the bottom of the function. That also means you can get rid of the guard boolean, since the lambda will never run if the error dialog never appears.

RDMurray:
1. The error specific to "copy loop of selected area of items" occurs if any of the items have a looppoint within the time selection. I think that would be tricky to detect if it's possible at all.
2. at least "copy selected area of items" does if there is no time selection, but that would be easy to detect. I don't know if there are any other cases where a dialog can appear. Splitting them up wouldn't mean much duplication because most of the logic is in the countAffected lambda.

jcsteh (approved):

PR #1192, opened 2025-02-04
Report midi event mute toggle

MatejGolian:
Opening this again as I accidentally close the original one.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -2318,15 +2318,19 @@ void postMidiToggleMute(int command) {
 				}
 				s << getMidiNoteName(take, selectedNotes[0].pitch, selectedNotes[0].channel);
 			} else {
-				// Translators: used when reporting the number of notes.
-				// {} will be replaced by the number of notes. E.g. "3 notes"
-			s << format(translate("{} notes"), noteCount);
 				int mutedCount = count_if(selectedNotes.begin(), selectedNotes.end(), [](auto note) { return note.muted; });
+				int unmutedCount = noteCount - mutedCount;
 				if (mutedCount > 0) {
 					// Translators: used when reporting the number of muted notes.
-					// {} will be replaced by the number of muted notes. E.g. "3 muted"
+					// {} will be replaced by the number of muted notes. E.g. "3 notes muted"
 					s << " " << format(
-					translate_plural("{} muted", "{} muted", mutedCount), mutedCount);
+					translate_plural("{} note muted", "{} notes muted", mutedCount), mutedCount);
+				}
+				if (unmutedCount > 0) {
+					// Translators: used when reporting the number of unmuted notes.
+					// {} will be replaced by the number of notes. E.g. "3 notes unmuted"
+					s << " " << format(
+					translate_plural("{} note unmuted", "{} notes unmuted", unmutedCount), unmutedCount);
 				}
 			}
 		}
@@ -2341,31 +2345,39 @@ void postMidiToggleMute(int command) {
 			}
 			s << describeCC(take, cc);
 		} else {
-			// Translators: used when reporting the number of CCs.
-			// {} will be replaced by the number of CCs. E.g. "3 CCs"
-			s << format(translate("{} CCs"), CCCount);
 			int mutedCount = count_if(selectedCCs.begin(), selectedCCs.end(), [](auto cc) { return cc.muted; });
+			int unmutedCount = CCCount - mutedCount;
 			if (mutedCount > 0) {
 				// Translators: used when reporting the number of muted CCs.
-				// {} will be replaced by the number of muted CCs. E.g. "3 muted"
+				// {} will be replaced by the number of muted CCs. E.g. "3 CCs muted"
 				s << " " << format(
-				translate_plural("{} muted", "{} muted", mutedCount), mutedCount);
+				translate_plural("{} CC muted", "{} CCs muted", mutedCount), mutedCount);
+			}
+			if (unmutedCount > 0) {
+				// Translators: used when reporting the number of unmuted CCs.
+				// {} will be replaced by the number of CCs. E.g. "3 CCs unmuted"
+				s << " " << format(
+				translate_plural("{} CC unmuted", "{} CCs unmuted", unmutedCount), unmutedCount);
 			}
 		}
 	} else { // If both notes and CCs are selected
-		// Translators: used when reporting the number of events.
-		// {} will be replaced by the number of events. E.g. "3 events"
-		s << format(translate("{} events"), eventCount);
 		int mutedNoteCount = count_if(selectedNotes.begin(), selectedNotes.end(), [](auto note) { return note.muted; });
 		int mutedCCCount = count_if(selectedCCs.begin(), selectedCCs.end(), [](auto cc) { return cc.muted; });
 		int mutedCount = mutedNoteCount + mutedCCCount;
+		int unmutedCount = eventCount - mutedCount;
 		if (mutedCount > 0) {
 			// Translators: used when reporting the number of muted events.
-			// {} will be replaced by the number of muted events. E.g. "3 muted"
+			// {} will be replaced by the number of muted events. E.g. "3 events muted"
 			s << " " << format(
-			translate_plural("{} muted", "{} muted", mutedCount), mutedCount);
+			translate_plural("{} event muted", "{} events muted", mutedCount), mutedCount);
 		}
-	}
+		if (unmutedCount > 0) {
+			// Translators: used when reporting the number of unmuted events.
+			// {} will be replaced by the number of events. E.g. "3 events unmuted"
+			s << " " << format(
+			translate_plural("{} event unmuted", "{} events unmuted", unmutedCount), unmutedCount);
+		}
+}
 	if (s.tellp() > 0) {
 		outputMessage(s);
 	}
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -2295,16 +2295,16 @@ void postToggleFunctionKeysAsStepInput(int command) {
 void postMidiToggleMute(int command) {
 	HWND editor = MIDIEditor_GetActive();
 	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int eventCount = countSelectedEvents(take);
+	if (eventCount == 0) {
+		return;
+	}
 	// Get selected notes.
 	vector<MidiNote> selectedNotes = getSelectedNotes(take);
 	// Get selected CCs.
 	vector<MidiControlChange> selectedCCs = getSelectedCCs(take);
-	int totalNoteCount = selectedNotes.size();
+	int noteCount = selectedNotes.size();
 	int CCCount = selectedCCs.size();
-	int eventCount = selectedNotes.size() + selectedCCs.size();
-	if (eventCount == 0) {
-		return;
-	}
 	ostringstream s;
 	if (CCCount == 0) { // If only notes are selected
 		auto chord = findChord(take, 0, {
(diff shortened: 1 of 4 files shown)


src/midiEditorCommands.cpp
```diff
@@ -998,11 +998,12 @@ vector<MidiControlChange> getSelectedCCs(MediaItem_Take* take, int offset=-1) {
 		if (ccIndex == -1) {
 			break;
 		}
+		bool muted;
 		double position;
 		int chan, msg1, msg2, msg3;
-		MIDI_GetCC(take, ccIndex, nullptr, nullptr, &position, &msg1, &chan, &msg2, &msg3);
+		MIDI_GetCC(take, ccIndex, nullptr, &muted, &position, &msg1, &chan, &msg2, &msg3);
```
jcsteh, line 1004: Hmm. I don't understand why we aren't using MidiControlChange::get here instead of calling MIDI_GetCC directly. We do use MidiNote::get in getSelectedNotes. Still, that's something we can clean up separately.

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+		auto chord = findChord(take, 0, {
+			true,  // start
+			true,  // end
+			true,  // channel
+			true,  // pitch
+			true,  // velocity
+			false,  // selected
+			true  // muted
+		});
+		// Select and play the current chord.
```
jcsteh, line 2320: 1. If I understand this correctly, this assumes that the selected notes are the current chord, but this might not be the case. It's possible to select notes all over the place. We probably shouldn't try to preview if any of the selected notes aren't within the current chord.
2. We aren't selecting anything here.

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+		vector<MidiNote> notes(chord.first, chord.second);
+		previewNotes(take, notes);
+		fakeFocus = FOCUS_NOTE;
+		if (settings::reportNotes) {
+			if (noteCount == 1) {
+				if (selectedNotes[0].muted)
+				s << translate("muted") << " ";
+				else
+				s << translate("unmuted") << " ";
```
jcsteh, line 2329: 1. These two `s <<` lines should be indented 1 additional level.
2. I'd suggest bracing the if and else blocks here. Even though there's only 1 line of code each, it can be confusing, particularly in a nested if statement and particularly if the indentation gets messed up.

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+			} else {
+				// Translators: used when reporting the number of notes.
+				// {} will be replaced by the number of notes. E.g. "3 notes"
+			s << format(translate("{} notes"), noteCount);
+				int mutedCount = count_if(selectedNotes.begin(), selectedNotes.end(), [](auto note) { return note.muted; });
+				if (mutedCount > 0) {
+					// Translators: used when reporting the number of muted notes.
+					// {} will be replaced by the number of muted notes. E.g. "3 muted"
+					s << " " << format(
+					translate_plural("{} muted", "{} muted", mutedCount), mutedCount);
```
jcsteh, line 2340: I realise this is consistent with the messaging for muted notes when we report chords, but I worry this is confusing here. If I understand correctly, this might result in something like "7 notes 3 muted", which really means "3 notes muted 4 notes unmuted". It's unclear in this context what "7 notes" means. In the case of a chord, you're navigating to it, so "7 notes" means the total number of notes in the chord. But when you're toggling mute, "7 notes" is unqualified.

jcsteh (changes requested):
Thanks. Sorry for the delay in reviewing this.

In addition to my other comments here, please also add this action to the appropriate section of the readme.

MatejGolian:
@jcsteh, what kind of feedback would you advise then? Something like "X Notes/CCs/Events muted Y Notes/CCs/Events unmuted"? And what should get reported in case one of the counts was 0?
Apart from that and the MidiControlChange::get thing, everything else should be fixed now.

jcsteh:
Yes, I'd suggest, for example, "x notes muted, y notes unmuted". If only muting or unmuting (not both), it should just be "x notes muted". Note that you can just subtract muted from the selection count to determine unmuted, since REAPER will always mute or unmute everything in the selection.

MatejGolian:
@jcsteh, it may not be exactly what you had in mind, but perhaps it could work?

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,96 @@ void postToggleFunctionKeysAsStepInput(int command) {
 	}
 }
 
+void postMidiToggleMute(int command) {
+	HWND editor = MIDIEditor_GetActive();
+	MediaItem_Take* take = MIDIEditor_GetTake(editor);
+	int eventCount = countSelectedEvents(take);
```
jcsteh, line 2299: Sorry I didn't spot this in the first pass, but I just realised countSelectedEvents is an OSARA function which has to walk all the selected events itself. We then go and walk them again when we get selected notes and CCs. That seems like redundant walk and could be slower than it needs to be if there are a lot of selected events. I suggest you just get the selected notes and CCs (since you're going to do that anyway) and add the counts together. Unless there's another specific reason you did it this way?

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,96 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+				}
+				s << getMidiNoteName(take, selectedNotes[0].pitch, selectedNotes[0].channel);
+			} else {
+				int mutedCount = count_if(selectedNotes.begin(), selectedNotes.end(), [](auto note) { return note.muted; });
+				int unmutedCount = noteCount - mutedCount;
+				if (mutedCount > 0) {
+					// Translators: used when reporting the number of muted notes.
+					// {} will be replaced by the number of muted notes. E.g. "3 notes muted"
+					s << " " << format(
+					translate_plural("{} note muted", "{} notes muted", mutedCount), mutedCount);
```
jcsteh, line 2327: When wrapping code onto a new line like this, please indent an additional level so it's clear that it's a continuation of the previous line.

jcsteh:
> it may not be exactly what you had in mind, but perhaps it could work?

Pretty much what I had in mind. Thanks.

Just a few little bits of polish and then I think we're done here.

MatejGolian:
Didn't know that that is an Osara function. Regarding that second thing, great catch. Removing that check/condition definitely makes sense irrespective of whether REAPER provides its own preview or not IMO. After all one may start of without having any kind of sound generator active and having absolutely no feedback in that case would be a bummer.

jcsteh:
> After all one may start of without having any kind of sound generator active and having absolutely no feedback in that case would be a bummer.

That's already true for a bunch of commands that provide note preview, and I think that's actually okay. If you turn off note reporting, you accept that you're relying on note preview. But I suspect toggling mute doesn't preview, so this case is an exception I think.

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+			}
+			s << getMidiNoteName(take, selectedNotes[0].pitch, selectedNotes[0].channel);
+		} else {
+			int mutedCount = count_if(selectedNotes.begin(), selectedNotes.end(), [](auto note) { return note.muted; });
+			int unmutedCount = noteCount - mutedCount;
+			if (mutedCount > 0) {
+				// Translators: used when reporting the number of muted notes.
+				// {} will be replaced by the number of muted notes. E.g. "3 notes muted"
+				s << " " << format(
+				translate_plural("{} note muted", "{} notes muted", mutedCount), mutedCount);
```
jcsteh, line 2326:
```suggestion
					translate_plural("{} note muted", "{} notes muted", mutedCount), mutedCount);
```

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+			}
+			if (unmutedCount > 0) {
+				// Translators: used when reporting the number of unmuted notes.
+				// {} will be replaced by the number of notes. E.g. "3 notes unmuted"
+				s << " " << format(
+				translate_plural("{} note unmuted", "{} notes unmuted", unmutedCount), unmutedCount);
```
jcsteh, line 2332:
```suggestion
					translate_plural("{} note unmuted", "{} notes unmuted", unmutedCount), unmutedCount);
```

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+			}
+			s << describeCC(take, cc);
+		} else {
+			int mutedCount = count_if(selectedCCs.begin(), selectedCCs.end(), [](auto cc) { return cc.muted; });
+			int unmutedCount = CCCount - mutedCount;
+			if (mutedCount > 0) {
+				// Translators: used when reporting the number of muted CCs.
+				// {} will be replaced by the number of muted CCs. E.g. "3 CCs muted"
+				s << " " << format(
+				translate_plural("{} CC muted", "{} CCs muted", mutedCount), mutedCount);
```
jcsteh, line 2352:
```suggestion
					translate_plural("{} CC muted", "{} CCs muted", mutedCount), mutedCount);
```

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+			}
+			if (unmutedCount > 0) {
+				// Translators: used when reporting the number of unmuted CCs.
+				// {} will be replaced by the number of CCs. E.g. "3 CCs unmuted"
+				s << " " << format(
+				translate_plural("{} CC unmuted", "{} CCs unmuted", unmutedCount), unmutedCount);
```
jcsteh, line 2358:
```suggestion
					translate_plural("{} CC unmuted", "{} CCs unmuted", unmutedCount), unmutedCount);
```

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+	} else { // If both notes and CCs are selected
+		int mutedNoteCount = count_if(selectedNotes.begin(), selectedNotes.end(), [](auto note) { return note.muted; });
+		int mutedCCCount = count_if(selectedCCs.begin(), selectedCCs.end(), [](auto cc) { return cc.muted; });
+		int mutedCount = mutedNoteCount + mutedCCCount;
+		int unmutedCount = eventCount - mutedCount;
+		if (mutedCount > 0) {
+			// Translators: used when reporting the number of muted events.
+			// {} will be replaced by the number of muted events. E.g. "3 events muted"
+			s << " " << format(
+			translate_plural("{} event muted", "{} events muted", mutedCount), mutedCount);
```
jcsteh, line 2370:
```suggestion
				translate_plural("{} event muted", "{} events muted", mutedCount), mutedCount);
```

src/midiEditorCommands.cpp
```diff
@@ -2292,6 +2293,94 @@ void postToggleFunctionKeysAsStepInput(int command) {
...
+		}
+		if (unmutedCount > 0) {
+			// Translators: used when reporting the number of unmuted events.
+			// {} will be replaced by the number of events. E.g. "3 events unmuted"
+			s << " " << format(
+			translate_plural("{} event unmuted", "{} events unmuted", unmutedCount), unmutedCount);
```
jcsteh, line 2376:
```suggestion
				translate_plural("{} event unmuted", "{} events unmuted", unmutedCount), unmutedCount);
```

jcsteh (commented):
Fix indentation.

jcsteh (approved):
Thanks!

PR #1193, opened 2025-02-07
Added new action OSARA; Select items under edit cursor on selected tracks

ScottChesworth:
This PR adds a new OSARA action to replace an SWS one, partly because code from the original contributor is no longer maintained, also because it's been bothering me that something fundamental like splitting items has an external dependency. It follows the SWS implementation closely, but is easier to read and is future-proofed by using newer APIs. I intend to follow up soon with improved feedback when items couldn't be split, figured I should get this new action checked over and merged before I go any deeper down the rabbit hole though.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -783,7 +783,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: go to first track: Control+Alt+Home
 - OSARA: go to last track: Control+Alt+End
 - OSARA: go to master track: Control+Alt+Shift+Home
-- OSARA: Select items under edit cursor on selected tracks
+- OSARA: Select items under edit cursor on selected tracks: Shift+A
 - OSARA: Move to next item (leaving other items selected): Control+Shift+RightArrow
 - OSARA: Move to previous item (leaving other items selected): Control+Shift+LeftArrow
 - OSARA: View properties for current media item/take/automation item (depending on focus): Shift+F2
--- a/readme.md
+++ b/readme.md
@@ -155,7 +155,6 @@ Most of these are actions built into REAPER, but a few are useful actions from t
 - Item grouping: Remove items from group: Control+Shift+G
 - Item grouping: Select all items in groups: Shift+G
 - Item: Select all items in current time selection: Control+Shift+A
-- (SWS extension) Xenakios/SWS: Select items under edit cursor on selected tracks: Shift+A
 - Item properties: Toggle mute: Control+F5
 - Item properties: Toggle solo exclusive: Control+F6
 - Item properties: Toggle lock
@@ -784,6 +783,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: go to first track: Control+Alt+Home
 - OSARA: go to last track: Control+Alt+End
 - OSARA: go to master track: Control+Alt+Shift+Home
+- OSARA: Select items under edit cursor on selected tracks
 - OSARA: Move to next item (leaving other items selected): Control+Shift+RightArrow
 - OSARA: Move to previous item (leaving other items selected): Control+Shift+LeftArrow
 - OSARA: View properties for current media item/take/automation item (depending on focus): Shift+F2
@@ -865,6 +865,7 @@ This list is worth referencing when making your own key map additions, assigning
 - Global automation override: No override (set automation modes per track)
 - Grid: Set to 1/64
 - Grid: Set to 1/128
+- (SWS extension) Xenakios/SWS: Select items under edit cursor on selected tracks
 - Item: Invert selection
 - Xenakios/SWS: Invert item selection
 - Edit: Cut items
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -3980,27 +3980,24 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
 
 void cmdSelectItemsUnderEditCursorOnSelectedTracks(Command* command) {
 	vector<MediaItem*> items;
-	for (int i = 0; i < CountTracks(nullptr); ++i) {
-		MediaTrack* track = GetTrack(nullptr, i);
-		for (int j = 0; j < CountTrackMediaItems(track); ++j) {
-			MediaItem* item = GetTrackMediaItem(track, j);
+	for (int t = 0; t < CountSelectedTracks(nullptr); ++t) {
+		MediaTrack* track = GetSelectedTrack2(nullptr, t, false);
+		for (int i = 0; i < CountTrackMediaItems(track); ++i) {
+			MediaItem* item = GetTrackMediaItem(track, i);
 			items.push_back(item);
 		}
 	}
-	PreventUIRefresh(1); // SWS use this
+	PreventUIRefresh(1); // Bringing this line in from SWS...
+	// API isn't clear waht this does, best guess is it's more efficient to redraw UI once for the entire operation
+	// instead of redrawing for every item we select.
 	Main_OnCommand(40289,0); // unselect all items
-	int i;
 	double cursorPosition=GetCursorPosition();
-	for (i=0;i<(int)items.size();++i) {
-		MediaTrack* track = (MediaTrack*)GetSetMediaItemInfo(items[i],"P_TRACK", 0);
-		if (isTrackSelected(track)) {
-			double itemStart = *(double*)GetSetMediaItemInfo(items[i], "D_POSITION", 0);
-			double itemLength = *(double*)GetSetMediaItemInfo(items[i], "D_LENGTH", 0);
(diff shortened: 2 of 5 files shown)


tbdalgaard:
thanks for doing this PR. I thought of creating an issue about this mainly because I saw some interesting results when I ran the SWS action in projects where two items were placed directly next to each other without any empty space between them. Here both items would be selected in a few cases. I was not able to figure the situation out, but you gave me the advice to uncheck the following checkbox under preferences, Editing behaviour: If no items are selected, some split/trim/delete actions affect all items at the edit cursor
I have not seen the issue since I changed this. But I hope the updated version of this action might fix this as well as a side effect.

ScottChesworth:
This version of the action won't behave any differently in practical terms. Should we consider unchecking the option you mentioned in Editing Behaviour as part of OSARA's optimal config?

tbdalgaard:
That would be great, however I wonder if this could get complicated as we allow more and more changes when OSARA is installed.Sendt fra min iPhone via Braille Sense U2/MBrailleDen 8. feb. 2025 kl. 11.09 skrev ScottChesworth ***@***.***>:﻿
This version of the action won't behave any differently in practical terms. Should we consider unchecking the option you mentioned in Editing Behaviour as part of OSARA's optimal config?

readme.md
```diff
@@ -784,6 +783,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: go to first track: Control+Alt+Home
 - OSARA: go to last track: Control+Alt+End
 - OSARA: go to master track: Control+Alt+Shift+Home
+- OSARA: Select items under edit cursor on selected tracks
```
jcsteh, line 786: Has shift+a been remapped to this? Should the keyboard shortcut be listed here?

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
 	Main_OnCommand(40256, 0); // Tempo envelope: Insert tempo/time signature change marker at edit cursor...
 }
 
+void cmdSelectItemsUnderEditCursorOnSelectedTracks(Command* command) {
+	vector<MediaItem*> items;
+	for (int i = 0; i < CountTracks(nullptr); ++i) {
+		MediaTrack* track = GetTrack(nullptr, i);
+		for (int j = 0; j < CountTrackMediaItems(track); ++j) {
```
jcsteh, line 3963: Short variable names are fine for small scopes, but it's easier to read if they have some relationship to what you're doing. For example, I would rename `i` to `t` (for track), and rename `j` to `i` (for item).

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
...
+			MediaItem* item = GetTrackMediaItem(track, j);
+			items.push_back(item);
+		}
+	}
+	PreventUIRefresh(1); // SWS use this
```
jcsteh, line 3968: I'd prefer to document the real reason we're doing this, rather than "because SWS does it", unless we really don't know. My guess is that we do this so that we redraw the UI once for the entire operation, rather than once for every item we select.

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
...
+	Main_OnCommand(40289,0); // unselect all items
+	int i;
```
jcsteh, line 3970: Why declare this here?

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
...
+	double cursorPosition=GetCursorPosition();
+	for (i=0;i<(int)items.size();++i) {
```
jcsteh, line 3972: Because you have a vector of MediaItems here and you don't need the index, you can just do:
`for (MediaItem* item: items) {`

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
...
+		MediaTrack* track = (MediaTrack*)GetSetMediaItemInfo(items[i],"P_TRACK", 0);
```
jcsteh, line 3973: Please use nullptr, not 0, since this parameter is a pointer.

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
 	Main_OnCommand(40256, 0); // Tempo envelope: Insert tempo/time signature change marker at edit cursor...
 }
 
+void cmdSelectItemsUnderEditCursorOnSelectedTracks(Command* command) {
+	vector<MediaItem*> items;
+	for (int i = 0; i < CountTracks(nullptr); ++i) {
+		MediaTrack* track = GetTrack(nullptr, i);
```
jcsteh, line 3962: Should we be skipping tracks that aren't selected? That way, we don't fetch a bunch of items we don't need. So you'd move the isTrackSelected check here and remove it later.

src/reaper_osara.cpp
```diff
@@ -3956,6 +3956,37 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
...
+	double cursorPosition=GetCursorPosition();
+	for (i=0;i<(int)items.size();++i) {
+		MediaTrack* track = (MediaTrack*)GetSetMediaItemInfo(items[i],"P_TRACK", 0);
+		if (isTrackSelected(track)) {
+			double itemStart = *(double*)GetSetMediaItemInfo(items[i], "D_POSITION", 0);
+			double itemLength = *(double*)GetSetMediaItemInfo(items[i], "D_LENGTH", 0);
+			double itemEnd = itemStart + itemLength;
+			if (cursorPosition >= itemStart && cursorPosition <= itemEnd) {
+				bool makeSelected = true;
+				GetSetMediaItemInfo(items[i], "B_UISEL", &makeSelected);
```
jcsteh, line 3980: You can use `&bTrue` here rather than defining your own bool. (This is defined in osara.h for this purpose.)

src/reaper_osara.cpp
```diff
@@ -5363,7 +5394,7 @@ Command COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Configure REAPER for optimal screen reader accessibility")}, "OSARA_CONFIGREAPEROPTIMAL", cmdConfigReaperOptimal},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Check for update")}, "OSARA_UPDATE", cmdCheckForUpdate},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Open online documentation")}, "OSARA_OPENDOC", cmdOpenDoc},
-	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report tempo and time signature at play cursor; press twice to add/edit tempo markers")}, "OSARA_MANAGETEMPOTIMESIGMARKERS", cmdManageTempoTimeSigMarkers},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Select items under edit cursor on selected tracks")}, "OSARA_SELITEMSEDITCURSSELTRACKS", cmdSelectItemsUnderEditCursorOnSelectedTracks},	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report tempo and time signature at play cursor; press twice to add/edit tempo markers")}, "OSARA_MANAGETEMPOTIMESIGMARKERS", cmdManageTempoTimeSigMarkers},
```
jcsteh, line 5397: Two lines have accidentally been merged into one here.

jcsteh (changes requested):
Thanks.

src/reaper_osara.cpp
```diff
@@ -3978,6 +3978,34 @@ void cmdManageTempoTimeSigMarkers(Command* command) {
...
+	for (int t = 0; t < CountSelectedTracks(nullptr); ++t) {
+		MediaTrack* track = GetSelectedTrack2(nullptr, t, false);
+		for (int i = 0; i < CountTrackMediaItems(track); ++i) {
+			MediaItem* item = GetTrackMediaItem(track, i);
+			items.push_back(item);
+		}
+	}
+	PreventUIRefresh(1); // Bringing this line in from SWS...
+	// API isn't clear waht this does, best guess is it's more efficient to redraw UI once for the entire operation
+	// instead of redrawing for every item we select.
```
jcsteh, line 3992:
```suggestion
	// We might select multiple items. To improve performance, only refresh the UI
	// after the entire operation is complete.
	PreventUIRefresh(1);
```

jcsteh (approved):

PR #1199, opened 2025-02-16
Fix reporting of the slider in the Parameters dialog with Windows Narrator.

### LINKED ISSUE #758: Parameters dialog: Narrator always reads value slider as 0%

jcsteh:
Narrator uses UIA, not MSAA. Unfortunately, we can't annotate the UIA Value pattern's Value property. Instead, we use a custom UIA implementation.

Fixes #758.

### DIFF
--- a/src/archBuild_sconscript
+++ b/src/archBuild_sconscript
@@ -79,6 +79,7 @@ if env["PLATFORM"] == "win32":
 		"Advapi32",
 		"UIAutomationCore",
 		"ws2_32",
+		"Comctl32",
 	]
 
 else: # Mac
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -17,6 +17,9 @@
 #include "paramsUi.h"
 #include "midiEditorCommands.h"
 #include "translation.h"
+#ifdef _WIN32
+# include "uia.h"
+#endif
 
 using namespace std;
 
--- a/src/osara.h
+++ b/src/osara.h
@@ -353,14 +353,6 @@ std::string narrow(const std::wstring& text);
 
 extern IAccPropServices* accPropServices;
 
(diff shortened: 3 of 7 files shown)


jcsteh:
This works with Narrator and NVDA. It should hopefully work with JAWS too, but I'll need someone else to test this. It had better work or I shall be cross. This is nearly 300 lines and several hours of work.

ScottChesworth:
Wooo, it works! I checked with JAWS 2025 and Snowman scripts version 82, crashed around in the params dialog with approx a bazillion params for a worst case scenario. I could move through the combo box with Snowman's scripts appending values as expected, also tried Control+tabbing through sliders, adjusting them with arrows, page keys, home and end. To me, this feels identical with JAWS.
Is there anything specific I should try to make it angry?

ScottChesworth:
Working great with Narrator too but I guess you already knew that. I don't have a touch screen here, am getting someone to take that for a spin now just in case.

jcsteh:
That slider isn't going to work with touch, I'm afraid, in the same way that VO+up and VO+down won't increment or decrement it on the Mac. It'll read just fine, but altering the value has to be done with the keyboard. I may be able to fix that one day on Windows, but not today.

ScottChesworth:
Gotcha. Well so far as I can tell this is good to go. Thanks! Shout if you want anything else checked before merging.

PR #1203, opened 2025-02-21
Don't provide feedback for SWS actions if SWS is not installed.

### LINKED ISSUE #1202: Can OSARA check whether SWS is running?
Just found out that some feedback I added a while ago can lead to an inaccurate report if SWS isn't running. Is there a way to check? If not, can a way be added?

jcsteh:
If an extension action is bound in the key map, it seems REAPER registers a numeric id for it even if the extension isn't running. Previously, we intercepted that id such that OSARA's code for that action ran, even though the action didn't exist. Now, we check if there is a name registered for the action. If there isn't, we know the extension isn't installed and we thus ignore the action.
Fixes #1202.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -5668,14 +5668,25 @@ void delayedInit() {
 		} else if (COMMANDS[i].id) {
 			// This command is provided by an extension.
 			COMMANDS[i].gaccel.accel.cmd = NamedCommandLookup(COMMANDS[i].id);
+			KbdSectionInfo* section = SectionFromUniqueID(COMMANDS[i].section);
+			if (!kbd_getTextFromCmd(COMMANDS[i].gaccel.accel.cmd, section)[0]) {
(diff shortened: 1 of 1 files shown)


ScottChesworth:
It's working great here. I've Removed and readded SWS, checked all was expected using bindings in main section, Alt recording, alt layers and other sections. Is there anything deeper I should check?

jcsteh:
Ah. It doesn't crash because kbd_getTextFromCmd returns an empty string (never null) even if you pass it 0.

jcsteh:
Oh. If/when someone adds a cmd function for an SWS action (e.g. `_SWS_DEFAULTFADES` as discussed in #1179), this will need to be tested too to make sure that it doesn't break when SWS isn't installed, the action isn't in the key map, etc. I don't anticipate any problems, but strictly speaking, it's a different code path which hasn't been tested yet because we don't intercept any extension actions like that yet.

PR #1212, opened 2025-03-16
Add context menu options to restrict the minimum and maximum value of FX parameters in OSARA's FX Parameters dialog.

### LINKED ISSUE #1195: Osara parameter plugin dialog - ability to lock parameter range
Hey there,
I’d like to ask if it would be possible to add an option to the OSARA Plugin Dialog that allows setting a parameter slider's range.
Here’s an example:
When cleaning a specific area using an EQ, I might need to find the most prominent resonance in the lower mids, say between 200 Hz and 800 Hz. It would be great to be able to lock the lower and upper boundaries of the slider within this range, and once I find the exact frequency, I could unlock the slider again to work across the full range.
Thanks a lot!
Best regards,
Henry

jcsteh:
Fixes #1195.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -685,7 +685,7 @@ class ParamsDialog {
 		} else if (after == Param::AfterOption::invalidateParams) {
 			this->source->rebuildParams();
 			this->updateParamList();
-		} else {
+		} else if (after == Param::AfterOption::dismiss) {
 			SendMessage(this->dialog, WM_CLOSE, 0, 0);
 		}
 	}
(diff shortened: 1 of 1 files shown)


jcsteh:
This pull request courtesy of a 35 minute Uber ride in which I had nothing to do. :)

@HenriNH6, @Drizzell, can you try the build linked in https://github.com/jcsteh/osara/pull/1212#issuecomment-2727267026 and let me know if this works for you in practice?

HenriNH6:
Hi @jcsteh,
  Sorry it took me a while to test this. I've talked about it with @Drizzell too, and we're really happy with how the feature works - it's awesome!
  I just had one small suggestion: Would it be possible to keep the lower and upper limits of a parameter active, even after closing the plugin window (like pressing Escape) or switching to another parameter in the same plugin? Right now, the limits reset when I do either of those things. It would be great if the limits stayed until I manually clear them from the context menu.
  Even so, we are both very grateful for this feature, and we're really happy that OSARA has it.
  Thanks a lot for all your work on this feature - it's super useful!

ScottChesworth:
If these settings persist, we would need a way of notifying when a parameter has been range restricted.

jcsteh:
Remembering the restrictions across invocations is tricky in terms of both user experience (the concern Scott flagged) and implementation (how to reliably store the limits associated to a particular parameter, etc.). It's not something I'm able to dig into right now, even assuming it's possible to make it reliable.

PR #1219, opened 2025-03-24
Fix broken menu items in the Extensions -> OSARA menu on Mac.

### LINKED ISSUE #1218: Mac: Menu items does not trigger the expected action
This was discovered on RWP back in December 2024 by another member of the group, John.

I could not get a clean discussion link, so I have tried to make a short version from the thread here Note that the OSARA build versions are quite out of date, but the findings are still relevant in the latest development snapshots:

When going to the extensions and the OSARA submenu the follow menu items does not trigger the action they are bound to:

* Online documentation
* Check for update
* Configure REAPER for optimal screen reader accessibility

Expected: When activating any of the menu items above the action should be triggered. Like online documentation, will show up in a browser window etc.
Observed: Menu closes and focus is back to the Reaper window.

This was first located around December 6th 2024.
I ran an older version of OSARA in a new, portable Reaper.
Version: 2024.10.18.1617,2b17b87b
Here all the menu items work as expected.
I then updated that version of OSARA to:
Version: 2024.12.18.1644,aae6bab6

After doing so the menu items stopped working again.

The actions work though, if I run them via the action list, so the code itself seems to work correctly.

jcsteh:
In #1180, OSARA started registering all commands shortly after startup instead of immediately on startup. This was done to enable interception of SWS commands. Unfortunately, this broke items in the OSARA menu on Mac. Presumably, the menu hook gets called slightly earlier on Mac, before our commands get registered, and so our menu items can't be associated with the correct command. To fix this, we now register OSARA's commands immediately on startup, but we continue to register other commands that we want to intercept slightly later. Developers should note that OSARA's own commands are now listed in the OSARA_COMMANDS array, not the COMMANDS array.
Fixes #1218.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -5295,8 +5295,8 @@ void cmdJumpToTime(Command* command) {
 
 #define DEFACCEL {0, 0, 0}
 
+// REAPER or extension commands that we want to intercept.
 Command COMMANDS[] = {
-	// Commands we want to intercept.
 	{MAIN_SECTION, {{0, 0, 40285}, nullptr}, nullptr, cmdGoToNextTrack}, // Track: Go to next track
 	{MAIN_SECTION, {{0, 0, 40286}, nullptr}, nullptr, cmdGoToPrevTrack}, // Track: Go to previous track
 	{MAIN_SECTION, {{0, 0, 40287}, nullptr}, nullptr, cmdGoToNextTrackKeepSel}, // Track: Go to next track (leaving other tracks selected)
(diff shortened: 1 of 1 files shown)


jcsteh:
I haven't actually tested this yet on Mac. If someone is able to do that with the build that should arrive here in a few minutes, I'd be grateful.

tbdalgaard:
Thanks for at build so fast. I'll check it out tonight. Sendt fra min iPhone via Braille Sense U2/MBrailleDen 24. mar. 2025 kl. 06.42 skrev James Teh ***@***.***>:﻿
  I haven't actually tested this yet on Mac. If someone is able to do that with the build that should arrive here in a few minutes, I'd be grateful.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>



jcsteh left a comment (jcsteh/osara#1219)
I haven't actually tested this yet on Mac. If someone is able to do that with the build that should arrive here in a few minutes, I'd be grateful.

jennykbrennan:
Nice. All Osara menu items trigger just as they are supposed to here. Cheers!

tbdalgaard:
Same results as @jennykbrennan  says. NICE work!

Issue #593, opened 2021-11-10
Info sought: could we increase the range of values available in OSARA Params window beyond 10000 steps?


ScottChesworth:
In a recent thread on RWP, a user has run into problems controlling the midi ctrl src 1 and 2 parameters of Synth1 (a free, accessible and highly automatable VSTI). This seems to stem from the instrument supporting 65536 values per parameter, whereas I'm lead to believe that right now, the slider control OSARA uses supports 10000 steps.
The upshot of this is that with some parameters, it's not currently possible to increment or decrement in single steps using OSARA's Parameters dialog. Synth1 is being used as an example here, but I've also seen this with other VSTI's. IN particular, it seems to be tricksy wherever a parameter is relaying CC messages.

The discussion lead me to a couple of questions whenever someone with knowledge has time to jot down their thoughts:
1. Would there be much work involved to increase the amount of values our slider can support?
2. Would that be cross platform if implemented?
3. If there's any mileage in this, is there anything I can do to be useful? :)

TIA

jcsteh:
> I'm lead to believe that right now, the slider control OSARA uses supports 10000 steps.

Unless I'm forgetting something, I think it's 1000, not 10000.

>     1. Would there be much work involved to increase the amount of values our slider can support?

No. To be more precise, it's not about the number of values, but rather the minimum increment/decrement of the raw value that OSARA will try. Right now, that's the range of the value divided by 1000 (normally = 0.001). Look for this line in paramsUi.cpp:
`this->step = (this->max - this->min) / 1000;`
If you do change that, you should probably change the multiplier below it for largeStep as well. That's the value that gets used for page up/page down.

> 2. Would that be cross platform if implemented?

Yes.

>     3. If there's any mileage in this, is there anything I can do to be useful? :)

You can try changing that value and see what happens across various plugins. :)

This change is not without risk:

1. The smaller you make that step, the larger the number of values OSARA has to try. For a parameter which has a lot of granularity, that's a good thing. For a parameter which does not (in the worst case, an on/off parameter), that's a bad thing. Instead of trying 1000 possible values to get the next useful value, we might have to try 100000 values. That could make things noticeably slower. Or not, which is why it needs testing.
2. The smaller that value, the greater the possibility of floating point accuracy errors. That basically means you might see weird results sometimes (weirder than we already see in some cases).

[...]

PR #1227, opened 2025-04-18
Param sliders for kontakt and kk

ScottChesworth:
After discussion in #593, I've been investigating why values aren't being reported consistently when adjusting sliders in OSARA's FX Parameters dialog. Test plug-ins here have been Komplete Kontrol 3, Kontakt 6, 7 and 8.
STR:
1. Running a recent OSARA snapshot, load an instrument in any of the plug-ins specified above.
2. Go into the OSARA FX Parameters dialog, find a volume parameter.
3. Move the slider with arrows, page keys, home or end. All will seem ok until you change the direction you're moving in. Compare the consistency of formatted values being reported to the values displayed in OSARA's accompanying edit field. The former jumps all over the place when switching direction, the latter stays consistent and is the true value (can be verified by listening to the effect your adjustments are having on an instrument).
4. Now switch to the latest snapshot linked in this PR and repeat the test. This time you should find the formatted values are reported consistently when changing direction and jumping to ends of slider.

A 5ms sleep in paramsUI.cpp before we update value seems to stabilize reporting. As yet I don't know why that helps and I'm not proposing this as an actual solution, just a temporary hack to see whether behaviour also improves on a wider pool of machines. I'm especially interested in hearing about whether it helps on older/slower Windows and Macs. Not sure how long we need to wait for yet, or what a nicer implementation would be.

Tagging @timtam and @jennykbrennan, they've been helping chase it.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -340,12 +340,14 @@ class ParamsDialog {
 	}
 
 	void updateValue() {
-		std::this_thread::sleep_for(std::chrono::milliseconds(5));	
-		this->valText = this->param->getValueText(this->val);
-		this->updateValueText();
-		if (this->param->isEditable) {
-			SetWindowText(this->valueEdit, this->param->getValueForEditing().c_str());
-		}
+		// Testing callLater to delay operations.
+		CallLater([this]() {
+			this->valText = this->param->getValueText(this->val);
+			this->updateValueText();
+			if (this->param->isEditable) {
+				SetWindowText(this->valueEdit, this->param->getValueForEditing().c_str());
+			}
+		}, 5); // 5 milliseconds delay
 	}
 
 	void onParamChange() {
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -347,7 +347,7 @@ class ParamsDialog {
 			if (this->param->isEditable) {
 				SetWindowText(this->valueEdit, this->param->getValueForEditing().c_str());
 			}
-		}, 5); // 5 milliseconds delay
+		}, 50); // 50 milliseconds delay
 	}
 
 	void onParamChange() {
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -340,14 +340,11 @@ class ParamsDialog {
 	}
 
 	void updateValue() {
-		// Testing callLater to delay operations.
-		CallLater([this]() {
-			this->valText = this->param->getValueText(this->val);
-			this->updateValueText();
-			if (this->param->isEditable) {
-				SetWindowText(this->valueEdit, this->param->getValueForEditing().c_str());
-			}
-		}, 50); // 50 milliseconds delay
+		this->valText = this->param->getValueText(this->val);
+		this->updateValueText();
+		if (this->param->isEditable) {
+			SetWindowText(this->valueEdit, this->param->getValueForEditing().c_str());
+		}
 	}
 
 	void onParamChange() {
@@ -360,6 +357,8 @@ class ParamsDialog {
 	}
 
 	void onSliderChange(double newVal) {
+		static CallLater later;
+		later.cancel();
 		if (newVal == this->val
 				|| newVal < this->param->min || newVal > this->param->max) {
 			return;
@@ -389,7 +388,9 @@ class ParamsDialog {
 			}
 		}
 		this->param->setValue(this->val);
-		this->updateValue();
+		later = CallLater([this]() {
+			this->updateValue();
+		}, 50);
 	}
 
 	void onValueEdited() {
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -861,8 +861,8 @@ class FxParam: public Param {
 				}
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


jcsteh:
I'm away for a few days, so I can't look at where exactly you've put this sleep. If I had to guess, I'd say this relates to the fact that many plugins can't report formatted values for anything other than the current value. That means OSARA can't figure out where the next discrete value is. That would be even worse if the plugin also refused to report the formatted value for a newlyset value for a short period of time. OSARA would have no way of knowing this. I'm not ruling out some weird timing bug in OSARA though.

ScottChesworth:
Should I wait for you to get back before testing on a wider spread of machines?

jcsteh:
Not necessarily - data is always useful - but a 5 ms sleep is definitely not an acceptable solution to land, as it will slaughter performance when moving fast.

Timtam:
> Not necessarily - data is always useful - but a 5 ms sleep is definitely not an acceptable solution to land, as it will slaughter performance when moving fast.

I obviously considered that, but didn't find any other solution. We tried to go even lower and noticed that setting e.g. 1ms will not fix the issue. Unless we can actually get hold of the issue and fix it at its core 5 ms feels reasonably good ATM really, we tried with various constellations already and in none of them the delay is actually noticeable, as it also just occurs when arrowing or paging up or down within the parameter dialog. Obviously fixing the actual issue is to be preferred for sure.

jcsteh:
My concern is that holding down the arrow keys for a long time will get very laggy. Is that somewhing you've tested for? If it is a problem, it could be fixed with CallLater instead of sleep. This way, you can cancel it if the next key press arrives before the previous one is processed. It also avoids blocking the UI.

ScottChesworth:
Holding down keys was still ok on a good machine. I'll update the PR to use callLater before checking on slower ones, thanks for the tip.

LeonarddeR:
Is this only reproducible when a sound is loaded? I tested several channel volume parameters with KK but didn't have a patch loaded at the time, and I can't reproduce the issue.

jcsteh:
Yeah, I was playing with KK parameters yesterday and didn't experience this either. It wasn't snapping to values when I pressed arrow keys - i.e. I had to keep pressing with no reporting until it got to a new formatted value - but that's expected with plugins that can't format values other than the current set value.

jcsteh:
1. ParamsDialog::onSliderChange sets the value based on this->val.
2. ParamsDialog::updateValue then calculates the value text from this->val; i.e. the value that was just set. This is the value used for the slider.
3. However, updateValue sets the raw value in the text box by calling this->param->getValueForEditing().
4. If the plugin doesn't update the value immediately or it internally snaps the value to some other value (or both), (2) and (3) mean that we'll end up with inconsistent representations of the value between the slider and the text box.

We could get rid of the cached value in this->val and always get the latest value from the plugin. We would then pass it by argument to updateValue and updateValueText. The downside is that if the plugin snaps the value back to the original value (e.g. if we didn't step far enough to make it snap to the next value), we'll just be stuck on the same value forever. I think I've seen plugins do this. For example, Surge stepped parameters take a while to update and they also snap to the nearest step.

It also sounds like that's not the issue you were seeing with Kontakt and KK. updateValue always uses the cached value for the slider, so I guess TrackFX_FormatParamValue is providing a different string for the same raw value depending on when you call it.

Another possibility could be to wait for REAPER to tell us that the parameter value has changed via the surface API before we report anything. That will be a bit tricky to wire up, since they're in separate class instances, but there'll be some way. The plugin hopefully won't tell REAPER about a change until the plugin has done whatever it needs to do internally. But again, we have the snapping problem: if the plugin just snaps us back to our original value, we'll just be stuck there forever. Unless we just use the value change notification and keep using the value we're caching, but that seems kinda weird.

The snapping problem might not be such an issue for plugins which support formatting text values for all raw values, since OSARA will snap to the correct value in the first place. But for plugins which don't, it will be a problem. The question is whether there are any plugins that snap and don't support formatting values for any given raw value.

ScottChesworth:
> The question is whether there are any plugins that snap and don't support formatting values for any given raw value.

How could we find out such a thing in situ? I gots a lots of plug-ins, don't know what I'm looking for though.

jcsteh:
Both of the following would be true in this problematic case:

1. The plugin would only report percentages, or would report a friendly value but then report nothing when you repeatedly press the arrow keys until you eventually hit a different friendly value.
2. If you change the value in the text box, move away from the parameter and back again, the text box value is different to what you set it to.

jcsteh:
> Another possibility could be to wait for REAPER to tell us that the parameter value has changed via the surface API before we report anything.

This was a wonderfully intriguing idea, but actually, it's impossible because the surface API only notifies us about FX parameter changes for tracks, not takes.

jcsteh:
But maybe we can:

1. Store the value before the change.
2. Set the new value.
3. Retrieve the current value.
4. Check the stored original value (1) against the current value (3).
5. If they're different, great. Proceed to use either the new value we set (2) or the new current value (3). To be determined; each option has pros and cons.
6. If not, CallLater and repeat from step 3, probably with some limit on the number of attempts.

jcsteh:
Superseded by #1418.

PR #1235, opened 2025-04-28
Improve Actions build times

### LINKED ISSUE #1232: Major differences in build times with GitHub Actions between Mac and Windows
The build step on Mac takes around 20 seconds, where the Windows step takes more than 3 minutes. I wonder where this discrepancy is coming from. Of course the Windows version has to build the sources twice and builds a setup, but that doesn't justify these longer build times IMO.

LeonarddeR:
Fixes #1232 

The major problem here is that running vcvarsall.bat is extremely slow on GithubActions. We can fix this by caching the output of vcvarsall, as SCons offers a mechanism for this. This at least gives us a build that is around one minute faster.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ defaults:
 env:
   publisher: James Teh
   VSCMD_SKIP_SENDTELEMETRY: 1
+  SCONS_CACHE_MSVC_CONFIG: 1
 concurrency:
   # There's no point continuing to run a build if it's already outdated by
   # another commit.
@@ -37,6 +38,7 @@ jobs:
           fi
           if [ ${{ matrix.os }} == windows-latest ]; then
             echo os=windows >> "$GITHUB_ENV"
+            echo "CACHE_KEY=${RUNNER_OS}-${ImageVersion}" >> $GITHUB_ENV
           else
             echo os=mac >> "$GITHUB_ENV"
           fi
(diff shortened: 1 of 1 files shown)


LeonarddeR:
This looks like a broader SCons problem. Closing for now.

LeonarddeR:
Still some tricks to try

.github/workflows/build.yml
```diff
@@ -36,6 +38,7 @@ jobs:
           fi
           if [ ${{ matrix.os }} == windows-latest ]; then
             echo os=windows >> "$GITHUB_ENV"
+            echo "CACHE_KEY=${RUNNER_OS}-${ImageVersion}" >> $GITHUB_ENV
```
jcsteh, line 41: Out of interest, is there any harm (or performance hit) in setting CACHE_KEY on Mac even if we don't use it? My thinking is that we might one day use it and its value isn't specific to Windows, so we may as well just set it across the board. That way, we could do it in the env: block instead of doing it conditionally here.

.github/workflows/build.yml
```diff
@@ -48,6 +51,14 @@ jobs:
         run: python tools/sortKeymap.py -t config/${{ env.os }}/reaper-kb.ini
       - name: setup
         run: pip install scons
+      - id: get-scons-msvc-cache
+        name: Get SCons MSVC Cache
+        if: ${{ env.CACHE_KEY != '' }}
```
jcsteh, line 56: I think this should check the os rather than the presence of CACHE_KEY.We might want one day want CACHE_KEY for Mac, so this seems like a potential easy-to-miss footgun.

jcsteh (approved):
Thanks for working on this. Approving with comments answered or addressed. Feel free to merge once done.

.github/workflows/build.yml
LeonarddeR: I don't think we can do it in the env block. The runner os can be accessed with runner.os, but I'm afraid the image version is not available in the env block. Will test, though.

.github/workflows/build.yml
LeonarddeR: Addressed that.

PR #1236, opened 2025-05-02
Report moving or inserting project markers with actions on Control+NumRow keys

ScottChesworth:


ScottChesworth:
@jcsteh  I've requested a review because it's more lines than I typically add. Have tried to be tidy and I'm happy that the reporting works reliably, so it's probably not a biggy if you don't have time.

src/reaper_osara.cpp
```diff
@@ -4025,6 +4025,61 @@ void cmdMoveItemsOrEnvPoint(Command* command) {
...
+		ostringstream s;
+		if (name[0]) {
+			// Translators: used when reporting a named marker has been moved. {} will be
+			// replaced with the name of the marker; e.g. "v2 marker moved"
+			s << format(translate("{} marker moved"), name);
+		} else{
+			if (!markerExistsBefore) {
+				// Translators: Reports an unnamed marker has been inserted. {} is replaced with the marker number.
+				s << format(translate("marker {} inserted"), num);
+			} else if (beforePos == -1) {
```
jcsteh, line 4071: 1. Can beforePos be -1 if markerExistsBefore is true? In what situation?
2. If not, can we get rid of markerExistsBefore and just use beforePos -1 to detect this case?
3. If beforePos can be -1 in some other situation, can we combine these two checks into one to avoid code duplication? That is, `if (!markerExistsBefore || beforePos == -1)`

src/reaper_osara.cpp
```diff
@@ -4025,6 +4025,61 @@ void cmdMoveItemsOrEnvPoint(Command* command) {
...
+	int afterCount = CountProjectMarkers(0, nullptr, nullptr);
+	for (int i = 0; i < afterCount; ++i) {
+		bool reg;
+		int num;
+		const char* name;
+		EnumProjectMarkers(i, &reg, nullptr, nullptr, &name, &num);
+		if (num != wantNum || reg)
+			continue;
+		fakeFocus =FOCUS_MARKER;
+		ostringstream s;
```
jcsteh, line 4062: The `ostringstream` isn't really useful here. We're not using it to join strings together with separators, so we can just `outputMessage(translate(format(...))`.

src/reaper_osara.cpp
```diff
@@ -4025,6 +4025,61 @@ void cmdMoveItemsOrEnvPoint(Command* command) {
 	}
 }
 
+void cmdInsertOrMoveSpecificMarker(Command* command) {
+	int cmd = command->gaccel.accel.cmd;
+	int beforeCount = CountProjectMarkers(0, nullptr, nullptr);
+	int wantNum;
+	// Work out the desired marker based on the command ID.
+	if (cmd == 40656)
+		wantNum = 10;
```
jcsteh, line 4034: I'd really prefer blocks to be braced, even if they're only a single statement, but it's not a big deal.

jcsteh (changes requested):
Thanks. This looks pretty good; just minor code polish.

PR #1252, opened 2025-05-17
Optimal REAPER config: Disable snap to visible grid in the MIDI Editor so movement by grid is not dependent on the horizontal zoom setting.

### LINKED ISSUE #1250: Optimal config: Disable MIDI Editor: Options: snap to visible grid (AKA Grid Cursor Actions Broken After Grid Change in MIDI Editor)
When the following actions are triggered via Misc: Pass through key to main window in the MIDI editor, they completely stop working as soon as the grid division is changed:
But note, this bug only occurs when an item containing MIDI notes is opened in the MIDI editor.
• View: Move cursor left to grid division
• View: Move cursor right to grid division
Issue encountered:
After changing the grid division in the MIDI editor, the actions to move the cursor to the previous or next grid division no longer work at all. The cursor does not respond.
Steps to reproduce:
1. Open a MIDI item in the Reaper MIDI editor.
2. Make sure the shortcuts for the actions above are assigned using "Misc: Pass through key to main window".
3. Change the grid division (for example, from 1/16 to 1/8).
4. Try moving the cursor using the corresponding actions.
Expected result:
The cursor should continue to move according to the newly selected grid division.
Actual result:
The actions have no effect after the grid division is changed.
Additional note:
This issue only occurs when these actions are triggered via the pass-through in the MIDI editor. Outside the MIDI editor, they work as expected.
Is this a bug in OSARA or in Reaper?
Thanks.

jcsteh:
Fixes #1250.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -219,7 +219,7 @@ struct ReaperSetting {
 };
 // If any settings are added, changed or removed below, this number should be
 // increased.
-constexpr int REAPER_OPTIMAL_CONFIG_VERSION = 3;
+constexpr int REAPER_OPTIMAL_CONFIG_VERSION = 4;
 const char KEY_REAPER_OPTIMAL_CONFIG_VERSION[] = "reaperOptimalConfigVersion";
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


jcsteh:
@ScottChesworth, feel like taking this for a spin? I'm pretty sure I've got it right and I have tested it myself, but messing with people's configs always makes me a bit nervous, so I think it's worthy of an extra check.

ScottChesworth:
Worked with a fresh ini and with mine. I reckon it's good to go. Thanks!

PR #1253, opened 2025-05-19
Provide feedback when inserting notes using f1-f12 step input.

### LINKED ISSUE #1247: Advice needed: trying to implement playing note previews when using FKeys for step input
This was requested on RWP and would be useful. I've added actions "Step input: Insert note at current note" and all the others for inserting notes up to +12 and -12 semitones to Command COMMANDS lookup table, they're all calling cmdMidiInsertNote, just like the insertion actions on I and Shift+I.
So far no dice, I don't get any note previews or speech feedback when the edit cursor advances. Figured maybe REAPER just doesn't support note previews for step input, although there is a likely-sounding option for previews on keyboard action which I've made sure is enabled, but even if we did need something from Cockos for that to work, cursor position not being reported is especially confusing to me.
What am I missing? Any advice appreciated.

jcsteh:
Fixes #1247.
I've tested this, but additional testing and scrutiny would be ideal:

1. Normal insertion (i and shift+i), since I heavily refactored that code to make it reusable for f1-f12 step input.
2. F1-f12 in the MIDI editor without step input enabled (to make sure I didn't somehow break these).
3. Step input with f1-f12 (advances cursor).
4. Step input with shift+f1-f12 (does not advance cursor).

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -1223,18 +1223,15 @@ void postMidiMovePitchCursor(int command) {
 	}
 }
 
-void cmdMidiInsertNote(Command* command) {
+void cmdhInsertNote(int oldCount, int relativeNote, bool reportNewPos) {
 	HWND editor = MIDIEditor_GetActive();
 	MediaItem_Take* take = MIDIEditor_GetTake(editor);
-	int oldCount;
-	MIDI_CountEvts(take, &oldCount, nullptr, nullptr);
-	MIDIEditor_OnCommand(editor, command->gaccel.accel.cmd);
 	int newCount;
 	MIDI_CountEvts(take, &newCount, nullptr, nullptr);
 	if (newCount <= oldCount) {
 		return; // Not inserted.
 	}
-	int pitch = MIDIEditor_GetSetting_int(editor, "active_note_row");
+	int pitch = MIDIEditor_GetSetting_int(editor, "active_note_row") + relativeNote;
 	// Get selected notes.
 	vector<MidiNote> selectedNotes = getSelectedNotes(take);
 	// Find the just inserted note based on its pitch, as that makes it unique.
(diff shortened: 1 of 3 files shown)


ScottChesworth:
Done the suggested testing and this is all working very goodly. I only have two questions (only he says, as if one of them isn't going to be the sort of question that'll make you sigh like an old bike tyre).
1. Unmodified FKeys advancing cursor and Shift+FKeys not advancing is the opposite of how I and Shift+I work. It doesn't make sense to have people holding Shift for what will obviously be the more useful scenario, so I thought  we'll swap I and Shift+I. Then I thought about how much learning material that's difficult/impossible to update it'll invalidate. Then I thought... heeeeelp!
2. Is Shift+FKeys a REAPER function or something you coded? If the latter, would you be open to Shift inserting octave down from the pitch cursor instead, still advancing cursor? Looks like the actions are there for it.

jcsteh:
>     1. Unmodified FKeys advancing cursor and Shift+FKeys not advancing is the opposite of how I and Shift+I work.

is it?

- Edit: Insert note at edit cursor: I or NumPad5
- Edit: Insert note at edit cursor (no advance edit cursor): Shift+I or Shift+NumPad5

>     2. Is Shift+FKeys a REAPER function or something you coded?

REAPER. It's not clear that this will happen though; i discovered it by mistake.

jcsteh:
> Shift inserting octave down from the pitch cursor instead, still advancing cursor? Looks like the actions are there for it.

I don't use the actions here; I let REAPER handle the key. Recall that those actions never advance the cursor.

ScottChesworth:
Fiddlesticks! The ReadMe is wrong about I and Shift+I. Those actions got switched around years ago, and in slightly more recent times both keystrokes got bound to custom actions that preserve rectangle note shapes, because users kept getting caught out by differing behaviour of the others. Sorry, no idea how that's slipped through for so long.
Lemme quiz some users who do note insertion a lot, see how mad they'd be if I make these two things consistent.

jcsteh:
Why did they get switched around anyway, if even you're saying that advancing is the more common scenario?

ScottChesworth:
I don't remember the rationale that won out. Was back in the key map consortium group days, half a bazillion voice notes per decision so not easy to search. I'm checking which training materials it would impact now, feels like a lot, but I do tend to react quite strongly to fragmentation, it might not be so bad.

jcsteh:
> Done the suggested testing and this is all working very goodly.

Merging on that basis, since none of the other stuff here is specifically related to this change.

PR #1262, opened 2025-06-12
Make space key play/stop in Parameters dialogs, letting REAPER handle…

ScottChesworth:
… the key when on relevant controls

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -304,8 +304,6 @@ class ParamsDialog {
 	HWND valueLabel;
 	HWND moreButton;
 	string filter;
-	HWND unnamedParamsCheckbox;
-	HWND closeButton;
 	vector<int> visibleParams;
 	int paramNum;
 	unique_ptr<Param> param;
@@ -551,13 +549,10 @@ class ParamsDialog {
 			}
 			return 1; // Eat the keystroke.
 		}
-		if (msg->wParam == VK_SPACE && (
-			msg->hwnd == dialog->moreButton ||
-			msg->hwnd == dialog->unnamedParamsCheckbox ||
-			msg->hwnd == dialog->closeButton
-		)) {
-			// Let REAPER handle the space key when the More button, the Include unnamed parameters check box and the Close button have focus, so users can audition results on the fly.
-			return 0; // Not interested.
+		if (msg->wParam == VK_SPACE && isClassName(GetFocus(), "Button")) {
+			// Pass the space key to our window for buttons and check boxes so they can
+			// be activated as expected.
+			return -1; // Pass to our window.
 		}
 		const bool alt = GetAsyncKeyState(VK_MENU) & 0x8000;
 		if (msg->hwnd == dialog->paramCombo ||
@@ -569,8 +564,8 @@ class ParamsDialog {
 				(VK_F1 <= msg->wParam && msg->wParam <= VK_F12) ||
 				// Anything with both alt and shift.
 				(alt && shift) ||
-				// Or Space.
-				(msg->wParam == VK_SPACE)
+				// Modified space.
+				(msg->wParam == VK_SPACE && (alt || control || shift))
 			) {
 				return -666; // Force to main window.
 			}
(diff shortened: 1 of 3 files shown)


src/paramsUi.cpp
```diff
@@ -549,8 +551,12 @@ class ParamsDialog {
 			}
 			return 1; // Eat the keystroke.
 		}
-		if (msg->wParam == VK_SPACE) {
-			// Let REAPER handle the space key so control+space works.
+		if (msg->wParam == VK_SPACE && (
+			msg->hwnd == dialog->moreButton ||
```
jcsteh, line 555: I don't understand why we we need to restrict this when we didn't before. Previously, we always passed the space key through to REAPER. Now, we're passing it through in less situations than we did before, not more. Can you explain?

src/paramsUi.cpp
```diff
@@ -562,7 +568,9 @@ class ParamsDialog {
 				// A function key.
 				(VK_F1 <= msg->wParam && msg->wParam <= VK_F12) ||
 				// Anything with both alt and shift.
-				(alt && shift)
+				(alt && shift) ||
+				// Or Space.
+				(msg->wParam == VK_SPACE)
```
jcsteh, line 573: If I read this correctly, this will cause space to play/stop in edit fields. That will prevent space from being typed into an edit field, which is bad for the filter in particular.

src/paramsUi.cpp
ScottChesworth: > Previously, we always passed the space key through to REAPER.
> 
That relied on what Cockos described as an "ancient hardcoded ctrl+space playback binding".
> Now, we're passing it through in less situations than we did before, not more. Can you explain?
> 
The hardcoded binding was removed in REAPER 7.40, so OSARA needs to handle transport differently in our params dialogs now.

src/paramsUi.cpp
ScottChesworth: > If I read this correctly, this will cause space to play/stop in edit fields. That will prevent space from being typed into an edit field, which is bad for the filter in particular.
> 
Space is working fine in edit fields, have already verified that on Windows and Mac. I don't really know why either, some other internal REAPER handling I guess, but it does work, and being able to hit the usual keystroke to play is a nice little intuition upgrade.

src/paramsUi.cpp
jcsteh: That doesn't really answer my question, though. We're actually more tightly restricting this key with this change, not less tightly restricting it. As I understand it, this change shouldn't be necessary at all; space should always have worked. I'm not arguing that this code doesn't work - I'm sure you've tested it - but I also don't think it's correct, and that makes me nervous about weird regressions down the line.

src/paramsUi.cpp
jcsteh: So... what happens if you remove this code? This bit of code is only related to edit fields.

ScottChesworth: > what happens if you remove this code?
> 
I'm terrible at reading snippets on here, sorry. Removing what I thought you wanted me to try stops it from building lol.
Do you want it tested without the if statement handling F12/Alt and Shift/Space, or something else?

ScottChesworth:
I'm trying to get Space playing/stopping so that users can hear results in context while adjusting parameters, keeping standard functionality in tact when focus is in edit boxes, on check boxes and buttons. All of that works reliably as things stand. Control+Space plays/pauses now, which wasn't intentional really, but is nice to have working in case users want to move their in point when they're auditioning changes.
Control+Space still plays/pauses when focus is in edit fields. I didn't anticipate that but don't think it matters, as there's no expected functionality for Control+Space in an edit box anyway. As things stand Control+Space will toggle check boxes and click buttons when they have keyboard focus. both feel ok to me.

jcsteh:
Oh I get it. It's the comment that confused me. The first comment says that we pass space to REAPER when on the buttons and check boxes to allow auditioning. On the contrary, we allow space to pass unhindered there because we want the controls to activate. aside from the comment, I still think it might be possible to make this cleaner, so I'll take a look soonish.

ScottChesworth:
Thanks for getting to it. Space doesn't work with focus on the Parameters combo box anymore, but that's not a biggy, most important thing is that people can hear results while they're focused on sliders again. I'm merging.

PR #1265, opened 2025-06-14
Let unmodified space play/stop when focused on Parameters combo box

ScottChesworth:
The behaviour of unmodified Space was changed in code review. I'm trying another PR with it factored into the cleaned up approach as I received several requests via REAPER Access and PMs.
Use case is that sometimes you want to have a quick listen to the project before choosing which parameter you're going to adjust next. When exploring available parameters, without unmodified Space working the way it does in this PR, users must move to a random slider before they can play/stop, they're reporting that doesn't feel intuitive.
I've checked that REAPER still passes unmodified Space through to edit fields.
If there's something I've overlooked that makes this complex or risky, feel free to close.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -582,7 +582,7 @@ Clearing the text in the Filter field shows the entire list.
 
 As an alternative to using the Parameter combo box, you can press control+tab or control+shift+tab anywhere in the dialog to move to the next or previous parameter, respectively.
 
-Space can be used to play/stop the project when focused on sliders.
+Space can be used to play/stop the project when focused on sliders or the Parameters combo box.
 Control+Space can be used to pause, thus moving the edit cursor if you want to cue up a different part of the project.
 
 Some effects expose a lot of unnamed parameters, which can make finding the useful ones challenging.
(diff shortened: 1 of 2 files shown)


jcsteh:
It's redundant to have a check for both space and modified space. The second check will never be reached; the first check will always be true before the second is.

The reason I don't love this code (and the reason I changed it in the first PR) is that strictly speaking, it passes space through to the main section even in edit fields. By a strict interpretation of "pass through to main section", that should mean that space plays/stops even in edit fields, but it doesn't seem to. I don't understand that and it worries me. My concern is that REAPER could "correct" that oversight at some point in the future and then we end up with a nasty regression here.

ScottChesworth:
> By a strict interpretation of "pass through to main section", that should mean that space plays/stops even in edit fields, but it doesn't seem to. I don't understand that and it worries me. My concern is that REAPER could "correct" that oversight at some point in the future and then we end up with a nasty regression here.
> 
Was thinking on that earlier, I'm not so sure it's an oversight. There's two global scopes, one called "Global" and t'other called "Global + text fields". I reckon it's reasonable to assume the behaviour we have hear is intended/stable, otherwise why would the separate "Global + text fields" scope exist? That one is the behaviour you're worried about, and that one has to be assigned intentionally.

I can revert until that's clarified with Justin if you'd prefer that approach?

PR #1267, opened 2025-06-16
dev: Refactor the code used to pass tab, arrow keys, etc. through to the Transient Detection Settings dialog so it can be used for other dialogs.

### LINKED ISSUE #1266: Question: passing tab and arrows to more dialogs
In reaper_osara.cpp at line 5103, we have code that makes the Transient Detection Settings dialog easier to navigate. It registers an accelerator hook passing tab, arrows, home/end etc to the dialog, so those keys serve our navigation purposes instead of going to the main window.

There are other places where this behaviour would be helpful, EG the new Phase Alignment dialog, some SWS dialogs.

OSARA collects/stores a relevant window handle dynamically and it unregisters the accelerator on dialog close, so is it ok if I refactor this, making it easier to read as a generic function we can call from more places?

When I roughly reused existing code just now to get the phase align dialog easier to work with, it felt like a lot of duplication.

Thanks in advance for any advice.

jcsteh:
All you have to do to use this for other toggle actions which open such a dialog is bind them to postDialogOverridingDialogKeys.
Fixes #1266.
@ScottChesworth, can you try adding post commands for other dialogs like this and see if this refactor works for your purposes? If it does, I can merge this and you can then add support for those dialogs in a separate PR.
Also, if you can come up with a better name than dialogOverridingDialogKeys, please, please let me know. I couldn't. :)

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -2590,6 +2590,55 @@ void postVirtualMidiKeyboard(int command) {
 	}
 }
 
+// Certain dialogs (e.g. Transient Detection Settings) deliberately pass most
+// keys to the main section. This makes it impossible for keyboard users to
+// navigate. To work around this, when these dialogs are open, we take note of
+// the window and use our global accelerator hook to pass tab, arrow keys, etc.
+// to the dialog.
+// The HWND of a dialog which overrides normal dialog keys.
+HWND dialogOverridingDialogKeys = nullptr;
+// Called from translateAccel.
+int dialogOverridingDialogKeysTranslateAccel(MSG* msg,
+	accelerator_register_t* accelReg
+) {
+	if (!dialogOverridingDialogKeys) {
+		return 0; // Normal handling.
+	}
+	if (!IsWindow(dialogOverridingDialogKeys)) {
+		// Dialog was closed.
+		dialogOverridingDialogKeys = nullptr;
+		return 0; // Normal handling.
+	}
+	if (msg->message != WM_KEYDOWN ||
+			GetParent(msg->hwnd) != dialogOverridingDialogKeys) {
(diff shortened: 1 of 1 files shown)


ScottChesworth:
> Also, if you can come up with a better name than dialogOverridingDialogKeys...
> 

interceptDialogKeyPresses? routeKeyEventsToDialog?

ScottChesworth:
This is working to add other REAPER native dialogs. Thanks! How do I bind dialogs from SWS?

jcsteh:
forceNormalDialogKeysHwnd and forceNormalDialogKeysTranslateAccel? Is that better?

ScottChesworth:
Yep that's good.

jcsteh:
But does postForceNormalDialogKeys make sense? That makes it seem like this is after normal dialog keys have been forced. What we're really saying is that this is after a dialog opens for which we want to force normal dialog keys.

horribleDiaogHwnd, horribleDialogTranslateAccel, postHorribleDialog...

jcsteh:
I'm not doing it. Hahaha.

ScottChesworth:
> At this rate, I might merge this some time in 2073.
> 
McLaren will arrive to +1 something long before then...

jcsteh:
Okay. I'm going to bed. Feel free to merge this as is if it works with the SWS actions, then submit another PR to deal with the new REAPER + SWS actions.

jcsteh:
Someone please put me out of my misery.

ScottChesworth:
REAPER and SWS dialogs are easy to pass to this now so I'm merging. Thanks! Phase Align tool and perhaps one other dialog will be in another PR shortly, couple of others I want to pass to this will come later, with more testing its turned out they'll need some TLC from SWS first.

PR #1279, opened 2025-06-29
Hide midi utility params

ScottChesworth:
This is a test to see whether navigation can be made more responsive on Mac in the FX parameters dialog.

1. Load a VST3 or VST3I that exposes thousands of unused params.
2. Navigate the list of parameters to see how VO responds. Horribly, probably
3. Now uncheck "Include unnamed, utility and internal reserved parameters  check box  checked", move through the list again and tell me whether it feels any better.

I've added a bunch more commonplace utility params to what we consider worth hiding. Note that I've done it in a stupid way that isn't merge-worthy, will do it nicer if this proves useful in practice.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [windows-2022, macos-latest]
+        os: [windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - name: set env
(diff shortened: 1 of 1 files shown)


jcsteh:
Any idea where these MIDI parameters even come from? They aren't exposed in REAPER's parameter lists in the FX dialog menus, nor can you create envelopes for them. That makes me think they must be special REAPER parameters only relevant to VST3i plugins. If so, is there some way we can skip them other than their names?

ScottChesworth:
Ooh good catch! I assumed those were exposed by default with VST3 since I was seeing them so often, hadn't occured they might be an internal REAPER thing. Makes me wonder whether Cockos could tighten stuff up. I can show them plenty of examples where RPL libraries would be way smaller and preset switching faster if these bloaty params weren't included. Gonna ask Justin about it now.

ScottChesworth:
> Could that be due to the spaces before/after the parentheses in your code?
> 
Kinda wish I could say no, it was way more complicated.

> Before we merge this, I'd want to merge all of this into the regular expression,
> 
Gotcha. Didn't do that because I don't know how to read or write those yet. Will try to learn about it.

jcsteh:
I'm happy to take a run at the regexp once you have confirmation that these strings are all working as expected.

PR #1280, opened 2025-06-30
Make Control+Alt+L work in FX Parameters dialog

ScottChesworth:


jcsteh:
I haven't tested yet, but I think there is a much simpler, more general fix here. I think the problem is this code:

```
		if (alt && 'A' <= msg->wParam && msg->wParam <= 'Z') {
			// Alt+letter could be an access key in our dialog; e.g. alt+p to focus
			// the Parameter combo box.
			return -1; // Pass to our window.
		}
```

That catches alt+letter, but it also catches control+alt+letter because I'm apparently a rookie and wasn't explicit enough. That should probably be something like:

`if (alt && !shift && !control && 'A' <= msg->wParam && msg->wParam <= 'Z') {`

ScottChesworth:
Oooh now I see it. Cool. You can fix your own stuff up right, rookie? :)

PR #1289, opened 2025-07-11
POC for ARM64EC osara

LeonarddeR:
This pr is a proof of concept for ARM64ec osara.
The current problem is that this is not yet arm64ec, but native arm64. arm64ec doesn't yet build, but that's probably because clang 19 doesn't fully support arm64ec yet. I think we'd have to wait until clang 21 is there, and if that still doesn't support arm64ec, we probably have to switch back to native msvc.

### DIFF
--- a/src/archBuild_sconscript
+++ b/src/archBuild_sconscript
@@ -63,6 +63,10 @@ if env["PLATFORM"] == "win32":
 	env.Append(PDB="${TARGET}.pdb")
 	# having symbols usually turns this off, but we have no need for unused symbols.
 	env.Append(LINKFLAGS='/OPT:REF')
+	match env["TARGET_ARCH"]:
+		case "arm64":
+			env.Append(CCFLAGS="--target=arm64-pc-windows-msvc")
+			env.Append(LINKFLAGS="/machine:arm64")
 	sources.extend((
 		"uia.cpp",
 		env.Object("win32_utf8.obj", "#include/WDL/WDL/win32_utf8.c"),
@@ -136,12 +140,12 @@ addExternalSources("#include/fmt/src", ("format.cc", "os.cc"))
 addExternalSources(
 	"#include/WDL/WDL/jnetlib",
 	("httpGet.cpp", "connection.cpp", "asyncdns.cpp", "util.cpp"),
-	CCFLAGS="-Wno-deprecated-declarations"
+	CCFLAGS="$CCFLAGS -Wno-deprecated-declarations"
 )
 addExternalSources(
 	"#include/simpleson",
 	("json.cpp",),
-	CCFLAGS="-Wno-deprecated-declarations"
+	CCFLAGS="$CCFLAGS -Wno-deprecated-declarations"
 )
 
 env.SharedLibrary(
--- a/sconstruct
+++ b/sconstruct
@@ -20,7 +20,11 @@ env.SetOption('num_jobs', multiprocessing.cpu_count())
 print("Building using {} jobs".format(env.GetOption('num_jobs')))
 
 if env["PLATFORM"] == "win32":
-	for arch, suffix in (("x86", "32"), ("x86_64", "64")):
+	for arch, suffix in (
+		("x86", "32"),
+		("x86_64", "64"),
+		("arm64", "_arm64ec"),
+	):
 		archEnv = Environment(tools=["msvc", "mslink", "textfile"],
 			TARGET_ARCH=arch, HOST_ARCH=arch, libSuffix=suffix,
 			version=env["version"], copyright=env["copyright"],


src/archBuild_sconscript
```diff
@@ -63,6 +63,10 @@ if env["PLATFORM"] == "win32":
 	env.Append(PDB="${TARGET}.pdb")
 	# having symbols usually turns this off, but we have no need for unused symbols.
 	env.Append(LINKFLAGS='/OPT:REF')
+	match env["TARGET_ARCH"]:
+		case "arm64":
+			env.Append(CCFLAGS="--target=arm64-pc-windows-msvc")
+			env.Append(LINKFLAGS="/machine:arm64")
```
LeonarddeR, line 69: Should become
```suggestion
			env.Append(CCFLAGS="--target=arm64ec-pc-windows-msvc")
			env.Append(LINKFLAGS="/machine:arm64ec")
```

src/archBuild_sconscript
```diff
@@ -38,31 +38,52 @@ sources = [
...
+	else:
+		env.Append(CPPDEFINES=[
+			"_CRT_SECURE_NO_DEPRECATE",
+			"_CRT_NONSTDC_NO_DEPRECATE",
+		])
+		env.Append(CXXFLAGS=[
+			# #479: fmt/format.h includes UTF-8 characters in comments. However, cl
+			# treats the file as ANSI unless it has a BOM. This causes a warning (and
+			# thus an error) on systems with certain locales. Just disable this warning.
+			"/wd4819",
```
jcsteh, line 78: I'm trying to move towards disabling warnings for includes using compiler directives at the location of the include where reasonable. Where a file is included more than once, this doesn't make a lot of sense, but I think we only include fmt/format.h from translation.h. That said, there are plenty of other places where we just do it in sconscript, so this is totally fine if you don't want to/can't deal with that here.

src/exports.cpp
```diff
@@ -25,7 +25,7 @@ bool osara_isShortcutHelpEnabled() {
 	return isShortcutHelpEnabled;
 }
 
-void osara_getVersion(char* versionOut, int versionOut_sz) {
+void osara_getVersion(char* versionOut, uintptr_t versionOut_sz) {
```
jcsteh, line 28: We can't change this to uintptr_t, as this would change how this has to be called from API consumers on 64 bit systems, since uintptr_t is 64 bit there. I guess we'll need to change `_vararg_osara_getVersion(` instead to do `(int)(uintptr_t)` or something like that. I can't quite remember the syntax to tell the compiler "yes, we know we're truncating, we're okay with it".

src/midiEditorCommands.cpp
```diff
@@ -740,7 +740,7 @@ void cmdMidiMoveCursor(Command* command) {
 		fakeFocus = FOCUS_NOTE;
 		s << " " << format(
 			translate_plural("{} note", "{} notes", count), count);
-		int mutedCount = count_if(notes.begin(), notes.end(), [](auto note) { return note.muted; });
+		int mutedCount = (int)count_if(notes.begin(), notes.end(), [](auto note) { return note.muted; });
```
jcsteh, line 743: The correct fix for a lot of these is probably to change the type to size_t (or just auto) instead, unless we end up passing them to REAPER directly or indirectly, in which case casting to int is easier/more correct.

jcsteh (changes requested):
Thanks for working on this.

One major open question here is still how we prevent REAPER from loading both the ARM64EC dll and the x64 dll on an ARM64 system. I think we need x64 ifdeffed code in the entry point that refuses to load the dll on an ARM64 system. Unless you have reason to think this isn't necessary?

src/archBuild_sconscript
LeonarddeR: It is a bit difficult to do with this for both MSVC and Clang. May be we should just wait for clang 21 to be release with Visual Studio. I think that will be able to create ARM64EC.

src/exports.cpp
LeonarddeR: I think I fixed it now.

LeonarddeR:
> One major open question here is still how we prevent REAPER from loading both the ARM64EC dll and the x64 dll on an ARM64 system. I think we need x64 ifdeffed code in the entry point that refuses to load the dll on an ARM64 system. Unless you have reason to think this isn't necessary?

We can do such a thing, but I'm not sure whether it's necessary. Won't REAPER automatically refuse loading another instance of OSARA when there's already one loaded? May be we can bring this up with Cockos?

jcsteh:
> Won't REAPER automatically refuse loading another instance of OSARA when there's already one loaded?

Why would it? From REAPER's perspective, how is it supposed to know that they're both OSARA? They look similar to us, but there's no identifying information for REAPER beyond the file name, unless I'm missing something?

src/archBuild_sconscript
jcsteh: We can live with this for now. If clang 21 does bring the needed goodness, we can drop msvc then.

That said, we need to find someone with an ARM64 system to test this, so we may be blocked on that for a bit.

jcsteh (changes requested):
Marking changes requested until we sort out this ARM64EC/x64 loading issue.

jcsteh:
For anyone who is lazy like me and can't be bothered updating to VS 2026 on their machine yet, you can just run `scons build/x86_64` and copy the dll out of `build/x86_64/reaper_osara64.dll`. :)

PR #1290, opened 2025-07-14
Support new scrub style and actions

ScottChesworth:


ScottChesworth:
REAPER 7.42 is released now with the new one-shot scrub style and actions I've added feedback for in this PR. It's quite a few translation strings but based on usage throughout the testing cycle, I reckon this is going to be a huge hit with newcomers, so I want to get the reporting friendly. Merge right away if you're happy with it, it's working as intended here.

jcsteh:
Even after reading the action names and the diff, I have no idea what this does or how it's supposed to work. I guess I'm going to have to play with it to understand it, heh.

src/reaper_osara.cpp
```diff
@@ -1261,6 +1261,11 @@ void postCursorMovementScrub(int command) {
 		fakeFocus = FOCUS_RULER; // Set this even if we aren't reporting.
 }
 
+void postInvertSegmentScrubRange(int command) {
+	// Translators: Reported when users invert the looped or one shot segment offsets.
+	outputMessage(translate("inverted segment offsets"));
```
jcsteh, line 1266: Is there a reason this isn't defined in `POST_COMMAND_MESSAGES` instead?

jcsteh (approved):
Thanks.

1. Despite not knowing wtf this actually is, I'm going to trust you know what you're doing with UX :), so I'm just doing a review purely on code here.
2. I pushed a commit with some code formatting nits.
3. See my inline comment regarding the post command. I'm approving this conditional on that being fixed unless there's a reason I've missed.

src/reaper_osara.cpp
ScottChesworth: > Is there a reason this isn't defined in POST_COMMAND_MESSAGES instead?
> 
Nope, I'm just stupidier at that time of night/morning. Moving that and merging now.

PR #1292, opened 2025-07-20
Report frozen tracks

ScottChesworth:
Marking this for review because honestly, I don't entirely understand all of the code in the new isTrackFrozen helper function. I've wanted this for ages, but couldn't find how to get the info in the API docs, so I harvested it from an SWS action and had help from GPT to find alternatives to SWS dependencies and make the styling more OSARA-y. I've left comments in for parts that I haven't seen OSARA using often.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -851,6 +851,7 @@ OSARA also includes some other miscellaneous actions.
 - OSARA: Report global / Track Automation Mode: Control+Shift+L
 - OSARA: Toggle global automation override between latch preview and off: Control+Alt+Shift+L
 - OSARA: Cycle through midi recording modes of selected tracks: Alt+Shift+\
+- OSARA: Report frozen tracks
 - OSARA: Report groups for current track
 - OSARA: Report length of last touched item
 - OSARA: Report number of takes in last touched item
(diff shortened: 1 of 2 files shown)


jcsteh:
I think this is a no go, sorry. In #383 regarding solo defeat, we discovered that retrieving the track state chunk can be quite slow on tracks containing a lot of data. That's fine if you're running an action on demand which queries a single track. It is not okay if you're doing this while navigating tracks (which needs to be very snappy) or you're running this across all tracks (where the performance problem multiplies). If we want to do this, we need to ask Cockos to give us a way to get this using GetSetMediaTrackInfo.

ScottChesworth:
Doh, forgot about that. Leave the PR open for now, I'll ask Cockos and either refactor when something shows up or close if they say no.

ScottChesworth:
Justin has added I_FREEZECOUNT to GetSetMediaTrackInfo, should be in 7.43. In the meantime I want to make sure I've grasped how this stuff fits together. If I add a helper function and start calling that in relevant places, am I right in thinking users on REAPER 7.43 or later will get the new reports, reporting can't happen in earlier versions but it shouldn't make older REAPER less stable for users that haven't upgraded yet?

PR #1296, opened 2025-07-28
Play MIDI note previews when tabbing to transients if a MIDI take is selected and Preferences -> Editing Behavior -> Tab through MIDI notes is enabled.

### LINKED ISSUE #1294: When Prefs/Editing Behavior/Tab through MIDI notes is checked and a take containing MIDI is selected, can OSARA play note previews?
A neat REAPER preference got mentioned on RWP recently. When it's checked, tab navigates through transients with a take containing audio selected as per usual, but when a take containing MIDI is selected, REAPER tabs through notes instead, even while the MIDI editor is closed. We already have speech feedback for the cursor position changing, I'd like to add note previews as well. Gonna need some advice.

1. Is there a way to learn whether a take contains MIDI without parsing its state chunk? If not, I can ask Cockos for another API addition.
2. How would we find out whether that preference is checked? My simple mind says ideally we need a toggle action for it.
3. Assuming I can get to a point where previews play, do you think this navigation should respect "Report position when navigating events in MIDI editor" and "Report MIDI notes in MIDI editor"? I feel like it should, but then, those configs specifically say "in MIDI Editor". This navigation will be happening with the editor closed.

Any thoughts/advice welcome. Commits that get this done faster than I'll manage to even more so.

jcsteh:
Fixes #1294.

Note that this uses the last focused item. That means this won't be very useful in some cases if you have multiple items selected. It's unclear to me whether that is a desired use case or not, and if it is, how it should be handled.

### DIFF
--- a/src/midiEditorCommands.cpp
+++ b/src/midiEditorCommands.cpp
@@ -2444,3 +2444,27 @@ int midiStepTranslateAccel(MSG* msg, accelerator_register_t* accelReg) {
 	}, 0);
 	return 0;
 }
+
+void previewChordForMoveToTransient(MediaItem_Take* take) {
+	// We specify direction as -1 because REAPER usually positions the cursor
+	// just slightly after the note, so we need the chord just before the cursor.
+	auto chord = findChord(take, -1, {
+		true,  // start
+		true,  // end
+		true,  // channel
+		true,  // pitch
+		true,  // velocity
+		false,  // selected
+		true  // muted
+	});
+	if (chord.first == chord.second) {
+		return;
+	}
+	if (chord.first->start < GetCursorPosition() - 0.0002) {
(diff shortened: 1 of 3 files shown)


jcsteh:
> 1. Sometimes previews don't play.

Hmm. I was seeing that too, but I thought I fixed it. 😕

> 2. REAPER tabs through notes in selected items, not only last focused. Is that hard to implement?

Yeah, it is pretty hard, unfortunately. This requires that we play notes in multiple takes, which means potentially multiple tracks. OSARA's MIDI preview code can't handle that. It could possibly be extended to do so, but that's not going to be easy.

I guess we should draft this one then until I/someone else has time to poke at it?

ScottChesworth:
Oki doke, I've drafted it for now.

PR #1297, opened 2025-08-14
Clarify reporting when REAPER automatically places start of time sele…

ScottChesworth:
…ction at the beginning of the timeline


ScottChesworth:
Users found the feedback I added a while back confusing. I've clarified per consensus on RWP and added missing translator comments.

PR #1300, opened 2025-08-24
Sign and Notarize the Current Mac installer

mulcahy1000:
Package the Mac installer in a .zip file instead of a .dmg. The .zip file includes a Mac app bundle (OSARAInstaller.app). This app is signed and notarized, and contains what used to be the "Install OSARA.command" script, the plugin shared library, the keymap, and the localization resources.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,6 +81,168 @@ jobs:
           certificate-profile-name: ${{ vars.AZURE_SIGNING_CERT_NAME }}
           files-folder: ${{ github.workspace }}\installer
           files-folder-filter: exe
+      - name: sign Mac
+        if: ${{ github.event_name == 'push' && matrix.os == 'macos-latest' }}
+        env:
+          APPLE_ID: ${{ secrets.APPLE_ID }}
+          APPLE_ID_PASSWORD: ${{ secrets.APPLE_ID_PASSWORD }}
+          TEAM_ID: ${{ secrets.TEAM_ID }}
+          APPLE_CERTIFICATE_P12: ${{ secrets.APPLE_CERTIFICATE_P12 }}
+          APPLE_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
+        run: |
+          set -euxo pipefail
+
+          # Create and configure keychain
+          security create-keychain -p temp_password build.keychain
+          security default-keychain -s build.keychain
+          security unlock-keychain -p temp_password build.keychain
+          security list-keychains -d user -s build.keychain
+
+          # Decode and save the certificate from the environment variable
+          echo "$APPLE_CERTIFICATE_P12" | base64 --decode > certificate.p12
+          
+          # Import certificate 
+          security import certificate.p12 -k build.keychain -P "$APPLE_CERTIFICATE_PASSWORD" -T /usr/bin/codesign -T /usr/bin/security
+
+          # Set key partition list for codesign access with specific certificate targeting
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k temp_password -D "Developer ID Application" -t private build.keychain
+
+          # Extract identity from build.keychain only
+          IDENTITY=$(security find-identity -v -p codesigning build.keychain | grep "Developer ID Application" | head -1 | sed 's/.*) \([^"]*\).*/\1/' | xargs)
+          echo "Signing with identity: $IDENTITY"
+          if [[ -z "$IDENTITY" ]]; then
+            echo "No codesigning identity found in build.keychain!"
+            exit 1
+          fi
+
+          cd installer/mac
+
+          # Set up app bundle and entitlements
+          APP_BUNDLE="OSARAInstaller.app"
+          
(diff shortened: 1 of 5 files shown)


ScottChesworth:
Nice! Feels like this approach could go out sooner. How hard would it be to:
1. Add a confirmation screen with a Close button?
2. Include the license in the app bundle?

mulcahy1000:
My perspective on this is to invest as little as possible into this approach. My goal here was to get something simpler to review that we could merge quickly as a quality of life improvement while we review the native installer (which I still think is the better long term approach).

Adding the license to the app bundle won't be very helpful, as it's hard for users to find it, and there's no dialog to display it as part of the installation process (one of several things sorted in the new installer) :)

I've added it to the .zip file. So the process now is:

- Download the.zip file
- Unzip it (creates a directory osara_version)
- Go into osara_version folder
- Run OSARAInstaller.app

We could of course put the license in the app bundle (it might already be there in fact), but users will never read it. :)

There isn't very much error handling in the current installer. I simply added a "Installation Complete" dialog with an "OK" button. We need better error handling and reporting in general, and the new installer gives us a better framework for that.

ScottChesworth:
Totally understand perspective and goal. The confirmation screen will help to decrease support requests when this rolls out, big thanks for adding that.

.github/workflows/build.yml
```diff
@@ -127,6 +290,8 @@ jobs:
     if: ${{ github.event_name == 'push' }}
     needs: build
     runs-on: ubuntu-latest
+    environment:
+      name: github-pages
```
jcsteh, line 294: Why do we need this?

jcsteh (approved):
Thanks for working on this interim approach.

> Unzip it (creates a directory osara_version)

To clarify, is this Finder behaviour; i.e. it creates a directory for the contents of the zip when you open the zip? So users could be expected to understand that this is what will happen? The only reason I ask is that on Windows, having to extract a .zip, going into the directory, etc. is considered more advanced user behaviour, hence self-extracting installers. But if this is just how things are done on Mac (except in the case of a pkg), that's fine.

.github/workflows/build.yml
mulcahy1000: Sorry, I was working on Workflows in my own repo. Removed. Thanks for catching this!

mulcahy1000:
As we discussed, this latest .zip file includes only the app bundle. So Finder behavior should be to put the app in in the same folder where the .zip file was unzipped. This behavior should remain consistent when we eventually upgrade to the new native installer.

.github/workflows/build.yml
```diff
@@ -81,6 +81,168 @@ jobs:
           certificate-profile-name: ${{ vars.AZURE_SIGNING_CERT_NAME }}
           files-folder: ${{ github.workspace }}\installer
           files-folder-filter: exe
+      - name: sign Mac
+        if: ${{ github.event_name == 'push' && matrix.os == 'macos-latest' }}
+        env:
+          APPLE_ID: ${{ secrets.APPLE_ID }}
+          APPLE_ID_PASSWORD: ${{ secrets.APPLE_ID_PASSWORD }}
+          TEAM_ID: ${{ secrets.TEAM_ID }}
```
jcsteh, line 89:
```suggestion
          APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
```

.github/workflows/build.yml
```diff
@@ -81,6 +81,168 @@ jobs:
...
+          zip -r "osara_temp.zip" $APP_BUNDLE
+
+          # Submit for notarization
+          echo "=== Submit for notarization ==="
+          
+          # Submit without verbose to get clean JSON output
+          if xcrun notarytool submit osara_temp.zip \
+            --apple-id "$APPLE_ID" \
+            --password "$APPLE_ID_PASSWORD" \
+            --team-id "$TEAM_ID" \
```
jcsteh, line 168:
```suggestion
            --team-id "$APPLE_TEAM_ID" \
```

.github/workflows/build.yml
```diff
@@ -81,6 +81,168 @@ jobs:
...
+                echo "❌ Notarization failed with status: $STATUS"
+                
+                # Try to get detailed log
+                SUBMISSION_ID=$(jq -r '.id // empty' notarization_output.json)
+                if [[ -n "$SUBMISSION_ID" ]]; then
+                  echo "=== Fetching detailed notarization log for submission $SUBMISSION_ID ==="
+                  if xcrun notarytool log "$SUBMISSION_ID" \
+                    --apple-id "$APPLE_ID" \
+                    --password "$APPLE_ID_PASSWORD" \
+                    --team-id "$TEAM_ID" > notarization_log.txt 2>&1; then
```
jcsteh, line 192:
```suggestion
                    --team-id "$APPLE_TEAM_ID" > notarization_log.txt 2>&1; then
```

jcsteh (approved):
TEAM_ID -> APPLE_TEAM_ID to make this more explicit in the list of secrets.

jcsteh:
I realised this needs an update to the readme around line 46 and below. I went ahead and merged it anyway, but if someone could update the readme with correct instructions, that'd be great. I don't feel comfortable doing that without being able to test it precisely myself to run through the steps.

PR #1301, opened 2025-08-25
When moving to items that have more than one take, only report number…

ScottChesworth:
… of non-empty takes for clarity

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -885,6 +885,19 @@ TimeFormat getPrimaryOrSecondaryTimeFormatForCommand() {
 	return TF_RULER;
 }
 
+int countNonEmptyTakes(MediaItem* item) {
+	if (!item)
+		return 0;
+	int totalTakes = CountTakes(item);
+	int nonEmptyTakes = 0;
+	for (int t = 0; t < totalTakes; ++t) {
+		if (GetTake(item, t)) {
+			nonEmptyTakes++;
+		}
+	}
+    	return nonEmptyTakes;
+}
+
 // End of utility/helper functions
 
 // Functions exported from SWS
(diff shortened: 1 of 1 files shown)


jcsteh:
Hmm. Should we include empty takes in the count when empty takes are switchable but not include them otherwise? Also, in the presence of empty takes, doesn't this sometimes mean that users will get take numbers that are larger than the count we reported?

ScottChesworth:
> Should we include empty takes in the count when empty takes are switchable but not include them otherwise?
> 
I didn't do that because nowadays REAPER seems to be doing a weird thing where it pads all takes you've passed over during a recording with empties, even if nothing was recorded into some of them. Seems like it's trying to keep the number of takes uniform. Probably useful when you can tell what's populated at a glance, but I found it super annoying, couldn't find a way to turn it off and didn't find it easy to track down when it changed.

> in the presence of empty takes, doesn't this sometimes mean that users will get take numbers that are larger than the count we reported?
>
Nope. Empties are there when you're cycling through with T or Shift+T, but only the populated takes get reported with numbers.

src/reaper_osara.cpp
```diff
@@ -3713,11 +3713,18 @@ void moveToItem(int direction, bool clearSelection=true, bool select=true) {
 		if (take) {
 			s << " " << GetTakeName(take);
 		}
-		int takeCount = CountTakes(item);
-		if (takeCount > 1) {
+		int totalTakes = CountTakes(item);
```
jcsteh, line 3716: This is used twice, so let's split this into a utility function in the utility area at the top of this file: countNonEmptyTakes or similar.

src/reaper_osara.cpp
```diff
@@ -3713,11 +3713,18 @@ void moveToItem(int direction, bool clearSelection=true, bool select=true) {
...
+		int nonEmptyTakes = 0;
+		for (int i = 0; i < totalTakes; ++i) {
```
jcsteh, line 3718: nit: Let's use `t` instead of `i` here to be a bit more descriptive: we're dealing with takes.

src/reaper_osara.cpp
```diff
@@ -3713,11 +3713,18 @@ void moveToItem(int direction, bool clearSelection=true, bool select=true) {
...
+			MediaItem_Take* take = GetTake(item, i);
+			if (take) {
```
jcsteh, line 3720: nit: We aren't using this variable except to check its value and the code to get the value is short, so we can simplify to: `if (GetTake(item, i))`

jcsteh (approved):
Fair enough. Just a bit of polish here.

PR #1305, opened 2025-09-05
Nudge and zoom step

ScottChesworth:
This provides convenient nudge steps which are linked to horizontal zoom. New OSARA actions cycle through nudge steps, reporting them in ms and seconds, adjusting horizontal zoom under the hood. This will replace stock horizontal zoom adjustment actions on Dash and Equals, although those adjustments can stay on their NumPad assignments just in case anyone wants to work with zoom the prior (less intuitive) way.
The concepts were thought up by @GarthHum and @mulcahy1000, so if you don't like the idea, talk to them.
The code was written by me and @Timtam, so if you don't like the code, talk to Toni. :)


src/reaper_osara.cpp
```diff
@@ -397,7 +409,12 @@ string formatTime(double time, TimeFormat timeFormat,
 			break;
 		}
 		case TF_FRAME: {
-			// Frames
+		case TF_ROUNDSEC: {
```
jcsteh, line 412: I don't think this is meant to be immediately below case TF_FRAMES. That will break TF_FRAMES. :)

src/reaper_osara.cpp
```diff
@@ -3756,6 +3778,53 @@ void cmdMoveToPrevItem(Command* command) {
 	}
 }
 
+void reportNudgeAndZoomStep(double nudgeTime) {
+	TimeFormat tf = TF_MILSEC;
+	if (nudgeTime >= 1.0) tf = TF_ROUNDSEC;
+	outputMessage(formatTime(nudgeTime, tf));
+}
+
+void nudgeAndZoomStep(int direction) {
```
jcsteh, line 3787: I don't think there's ever a chance that we will want to pass direction=0 here. If that's correct, I would change this to bool previous instead of int direction. That way, there's never any confusion that 0 is a valid value for direction.

src/reaper_osara.cpp
```diff
@@ -3756,6 +3778,53 @@ void cmdMoveToPrevItem(Command* command) {
...
+	// direction = +1 → zoom out (increase nudge time), -1 → zoom in (decrease nudge time)
+	double currentPPS = GetHZoomLevel();
+	if (currentPPS <= 0.0) {
+		outputMessage(translate("invalid zoom level"));
```
jcsteh, line 3791: Should this ever happen? If not, we could just assert here, or if you want to be extra paranoid, return without reporting anything. Otherwise, translators have to translate something that will never be used.

src/reaper_osara.cpp
```diff
@@ -3756,6 +3778,53 @@ void cmdMoveToPrevItem(Command* command) {
...
+		return;
+	}
+	double currentNudgeTime = 1.0 / currentPPS;
+	double nextNudgeTime = currentNudgeTime;
+	if (direction > 0) {
+		// Find next larger nudge
+		for (int i = 0; i < NUDGE_COUNT; ++i) {
```
jcsteh, line 3798: This can be simplified using a range-based for loop:
`for (double value : NUDGE_VALUES) {`
This also means we can get rid of NUDGE_COUNT.

src/reaper_osara.cpp
```diff
@@ -3756,6 +3778,53 @@ void cmdMoveToPrevItem(Command* command) {
...
+			if (currentNudgeTime < NUDGE_VALUES[i]) {
+				nextNudgeTime = NUDGE_VALUES[i];
+				break;
+			}
+		}
+	} else {
+		// Find next smaller nudge
+		for (int i = NUDGE_COUNT - 1; i >= 0; --i) {
```
jcsteh, line 3806: `for (double value : std::views::reverse(NUDGE_VALUES)) {`
You'll need to `#include <ranges>` near the top of the file.

src/reaper_osara.cpp
```diff
@@ -5732,6 +5801,8 @@ Command OSARA_COMMANDS[] = {
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Select items under edit cursor on selected tracks")}, "OSARA_SELITEMSEDITCURSSELTRACKS", cmdSelectItemsUnderEditCursorOnSelectedTracks},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report tempo and time signature at play cursor; press twice to add/edit tempo markers")}, "OSARA_MANAGETEMPOTIMESIGMARKERS", cmdManageTempoTimeSigMarkers},
 	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Report scrub segment offsets; press twice to edit")}, "OSARA_REPORTANDEDITSCRUBSEGMENT", cmdReportAndEditScrubSegmentOffsets},
+	{MAIN_SECTION, {DEFACCEL, _t("OSARA: Increase nudge and scrub faster (zoom out)")}, "OSARA_INCREASENUDGEANDZOOMSTEP", cmdIncreaseNudgeAndZoomStep},
```
jcsteh, line 5804: See my earlier comment regarding whether we want to put scrub first in this name.

jcsteh (changes requested):

src/osara.h
```diff
@@ -232,6 +233,12 @@ const int MAIN_SECTION = 0;
 const int MIDI_EVENT_LIST_SECTION = 32061;
 const int MEDIA_EXPLORER_SECTION = 32063;
 
+// Provide stepped choices for horizontal zoom, report equivalent nudge values in time instead of pixels for easier understanding
+static const double NUDGE_VALUES[] = {
```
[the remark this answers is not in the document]
jcsteh: You could potentially define it as a static const array inside the function, since it's only used in one function.

src/osara.h
```diff
@@ -317,10 +318,12 @@ typedef enum {
 	TF_MEASURE,
 	TF_MINSEC,
 	TF_SEC,
+	TF_ROUNDSEC,
 	TF_FRAME,
 	TF_HMSF,
 	TF_SAMPLE,
-	TF_MEASURETICK
+	TF_MEASURETICK,
+	TF_MILSEC
```
jcsteh, line 326: Rename TF_MILSEC to TF_MS.
```suggestion
	TF_MS
```

src/reaper_osara.cpp
```diff
@@ -411,6 +430,11 @@ string formatTime(double time, TimeFormat timeFormat,
 			s = formatTimeSample(time);
 			break;
 		}
+		case TF_MILSEC: {
```
jcsteh, line 433:
```suggestion
		case TF_MS: {
```

src/reaper_osara.cpp
```diff
@@ -3756,6 +3780,58 @@ void cmdMoveToPrevItem(Command* command) {
 	}
 }
 
+// REAPER's stock actions adjust horizontal zoom in increments of 20%, returning pixels per second.
+// We provide stepped zoom settings and report expected behaviour in time per keypress instead. It's easier to understand non-visually.
+void reportZoomStepsAsTime(double time) {
+	TimeFormat tf = TF_MILSEC;
```
jcsteh, line 3786:
```suggestion
	TimeFormat tf = TF_MS;
```

src/reaper_osara.cpp
```diff
@@ -397,10 +411,15 @@ string formatTime(double time, TimeFormat timeFormat,
 			break;
 		}
 		case TF_FRAME: {
-			// Frames
+		// Frames
```
jcsteh, line 414: Fix indentation.
```suggestion
			// Frames
```

src/reaper_osara.cpp
```diff
@@ -3756,6 +3780,58 @@ void cmdMoveToPrevItem(Command* command) {
...
+// We provide stepped zoom settings and report expected behaviour in time per keypress instead. It's easier to understand non-visually.
+void reportZoomStepsAsTime(double time) {
+	TimeFormat tf = TF_MILSEC;
+	if (time >= 1.0) tf = TF_ROUNDSEC;
+	// TRANSLATORS: Reported when users change horizontal zoom in steps.
+	// {} will be replaced with a formatted time increment (e.g. "1 ms", "2 sec").
+	outputMessage(format(translate("{} per keypress"), formatTime(time, tf)));
+}
+
+void zoomSteps(bool zoomOut) {
```
jcsteh, line 3793: Rename zoomSteps to adjustZoomByStep.
```suggestion
void adjustZoomByStep(bool zoomOut) {
```

src/reaper_osara.cpp
```diff
@@ -3756,6 +3780,58 @@ void cmdMoveToPrevItem(Command* command) {
...
+	}
+	reportZoomStepsAsTime(nextZoomStep);
+}
+
+void cmdZoomOutStepped(Command* command) {
+	zoomSteps(true);
+}
+
+void cmdZoomInStepped(Command* command) {
+	zoomSteps(false);
```
jcsteh, line 3832:
```suggestion
	adjustZoomByStep(false);
```

src/reaper_osara.cpp
```diff
@@ -3756,6 +3780,58 @@ void cmdMoveToPrevItem(Command* command) {
...
+	}
+	if (nextZoomStep != currentZoomStep) {
+		double newPPS = 1.0 / nextZoomStep;
+	adjustZoom(newPPS, 1, true, -1);
+	}
+	reportZoomStepsAsTime(nextZoomStep);
+}
+
+void cmdZoomOutStepped(Command* command) {
+	zoomSteps(true);
```
jcsteh, line 3828:
```suggestion
	adjustZoomByStep(true);
```

src/reaper_osara.cpp
```diff
@@ -3756,6 +3780,58 @@ void cmdMoveToPrevItem(Command* command) {
...
+void reportZoomStepsAsTime(double time) {
+	TimeFormat tf = TF_MILSEC;
+	if (time >= 1.0) tf = TF_ROUNDSEC;
+	// TRANSLATORS: Reported when users change horizontal zoom in steps.
+	// {} will be replaced with a formatted time increment (e.g. "1 ms", "2 sec").
+	outputMessage(format(translate("{} per keypress"), formatTime(time, tf)));
+}
+
+void zoomSteps(bool zoomOut) {
+	// direction = +1 to zoom out, -1 to zoom in
```
jcsteh, line 3794: Remove defunct comment.
```suggestion
```

jcsteh:
This is good to merge once the key map thingies are all thingied.

ScottChesworth:
Did key map thingies, gonna do one more round of testing in situ before I mergermajig though. All feedback welcome on extra steps and tweaked action names. To me, these steps feel like a good balance between flexible tape-style scrub speed and nudge values that are typically provided.

ScottChesworth:
I reckon the double zippage might be a factor because there's a difference in how versions of Mac OS handle that. I've seen downloads from PR threads work fine on a Mac running Monterey where the OS doesn't do anything fancy, but they're damaged on anything newer, where the OS extracts zipped apps automatically. I don't know how to stop it from doing that to see if it's the auto-extraction thing making a mess of a doubled zip.

jcsteh:
Err, GitHub seems to have pooped its pants and isn't detecting that I pushed a merge here. 😕

mulcahy1000:
I don't know of a way to make the message go away or change it. It's the standard Gatekeeper notification for unsigned apps or disk images. It's always been this way for pull request builds.

ScottChesworth:
I'd have known what to do about a gatekeeper prompt if it was a scary warning about security, but this message says the app is damaged and specifically says you should move it to the bin. Nothing about it even being possible to allow it anywhere. Is that normal language for Apple nowadays? Yikes.

ScottChesworth:
Cheers @pitermach, Sentinel makes it easy. I'll add a note to the ReadMe somewhere.

PR #1317, opened 2025-10-07
Improve feedback and map arm selected tracks

ScottChesworth:


jcsteh:
Yeah, seems reasonable to me. These behave quite differently, so it's reasonable to have quite different code for them.

I'm less thrilled about the duplicate boilerplate we have for looping through selected tracks for mute, solo, phase and now arm, but that problem predates your work here and can be addressed separately, if we can ever be bothered.

PR #1327, opened 2025-10-17
Bind "Transport: Record" action to Control+Enter and Control+NumPadEnter for alternative ways to record while focus is in VKB, FX window or other dialogs

ScottChesworth:
A while ago we added NumPadEnter as an alternative way to start recording, but there are situations where that keystroke didn't pass through to the main window as intended. Adding Control modifier and making the two alternatives Global scope addresses all the situations where users might want to start recording/punch-out. We've kept the Control modifier consistent on Mac in line with other transport keys that typically stay the same, easing transition for folks who use OSARA on both platforms.


ScottChesworth:
Thanks, I hadn't heard about that being added. Any other suggestions for a secondary record binding?

PR #1332, opened 2025-10-29
Add feedback for more actions moving item edges, refactored cmdMoveIt…

ScottChesworth:
…emEdgeOrSource function to match codestyle


ScottChesworth:
I've added "Item edit: Move position of item to edit cursor" to the key map on Control+Alt+Shift+Comma Windows, Command+Option+Shift+Comma on Mac. It's replacing "Item edit: Move left edge of item to edit cursor" because this is a more intuitive behaviour paired up with the move right edge action. This way, both actions preserve item length. Cockos plan to clarify the name of the action we had mapped before, but they don't plan to change its behaviour, hence this key map adjustment.

tbdalgaard:
This might break a few workflows like what @GarthHum does for editing. I have used this myself, and found it quite effective, because I can easily shrink an item to the cursor. Extremely useful for small edits, and ripple is not an issue in that regard. I can however see why you may want this change, here I would have used the nudge/set dialog instead, but if that helps more users then I might just use the custom actions we have for now, and remember the state of ripple. Just wanted to mention this.

PR #1337, opened 2025-11-24
Fix build on clang 20

LeonarddeR:
Build with Visual Studio 2026 (Clang 20) fails. This should fix that by ignoring another deprecation.
@jcsteh is there a particular reason why we're at such an old version of the fmt submodule?


LeonarddeR:
I think this is ready to roll so we can continue with #1289.

PR #1353, opened 2026-02-03
strip "&" from translations on macOS, fixes #1352

### LINKED ISSUE #1352: OSARA settings menu translation
A user reported to me that the OSARA settings menu with Ctrl+F12 is not translated on Mac. Is this normal? I don't have a Mac to test.
Otherwise everything else seems translated.

RDMurray:
Swell doesn't support nemonics for controls indicated by '&'. This adds entrys to the translation dictionary with the '&' chars stripped.

also add "Deutsch" to the langpack aliases because that is the name of the German translation linked from https://www.reaper.fm/langpack/

### DIFF
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -173,6 +173,5 @@ void translateDialog(HWND dialog) {
 	// Translate dialog title.
 	translateWindow(dialog, lParam);
 	// Translate controls.
-	HWND enumRoot = dialog;
-	EnumChildWindows(enumRoot, translateWindow, lParam);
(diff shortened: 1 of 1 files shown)


jcsteh:
Yeah, I had that thought. It's unclear where we'd do that stripping though. I guess we could strip it from the .po files as we put them into the Mac installer? That feels a bit weird, but I also don't know where else it could go. Unless we somehow wrap ifstream with our own istream implementation that can strip '&' characters? Whether that works depends on what ifstream features POParser requires. Looking briefly, it uses std::getline.

jcsteh:
Superseded by #1354.

PR #1365, opened 2026-02-25
Update fmt to 12.1.0 and use header only mode

LeonarddeR:
This updates the fmt module to version 12.1.0. It updates the `format` function in `translate.h` to use runtime formatting. For format calls that use string literals, i changed to fmt::format directly because that enables compile time checking of the format strings.
I also switched to header only mode, since it is sufficient to use the library this way and it fixes a showstopper for #1289.

### DIFF
--- a/src/archBuild_sconscript
+++ b/src/archBuild_sconscript
@@ -133,11 +133,6 @@ addExternalSources("#include/tinygettext/src", (
 	"tinygettext.cpp",
 ))
 addExternalSources(
-	"#include/fmt/src",
-	("format.cc", "os.cc"),
-	CCFLAGS="-Wunused-result"
-)
-addExternalSources(
 	"#include/WDL/WDL/jnetlib",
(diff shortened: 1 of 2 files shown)


src/reaper_osara.cpp
```diff
@@ -2572,7 +2572,7 @@ void postChangeVerticalZoom(int command) {
 	}
 	// Translators: Used when reporting the vertical zoom level as a number.
 	// {} will be replaced with the number; e.g. "35 vertical zoom".
-	outputMessage(translate(format("{} vertical zoom", zoom)));
+	outputMessage(format(translate("{} vertical zoom"), zoom));
```
jcsteh: Oops. Nice catch. I'm surprised no translation user ever noticed this was broken. :)

jcsteh (approved):
Great, thanks. Consider the format() compile error I mentioned, but if it's not feasible or too hard, you're good to squash merge. Please prefix the commit with "dev:" to make it clear this is about development stuff. (I forgot to do that with my merge of your previous PR, oops.)

PR #1368, opened 2026-03-02
Media explorer preview feedback

ScottChesworth:


jcsteh (changes requested):
Thanks.

I do feel like we should really bump our minimum API version at some point - we never did drop support for REAPER <= 6 - but pragmatically, I agree it's probably a bit intense to do that for this one case.

PR #1374, opened 2026-03-07
Alternative approach to hiding unusable parameters using string matching.

jcsteh:
This is based on @ScottChesworth's work in #1279, but it extends the regular expression rather than doing separate string searches. This does make the code less readable, but it means we handle all the string matching once. That said, in C++, this probably shouldn't make any difference from a performance perspective. The original code did match both at the start and the end of the string though, whereas this only matches at the start. I think if we go this path, we need to be a lot clearer about exactly what forms these strings come in. Single examples of each different type of string would be helpful.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -849,18 +849,7 @@ class FxParams: public ParamSource {
 			// Named config params aren't FX params; keep them visible.
 			return true;
 		}
-		const int fxParam = param - namedCount;
-		char buf[16] = {};
-		const bool success = this->_GetNamedConfigParm(
-			this->obj,
-			this->fx,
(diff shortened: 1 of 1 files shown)


jcsteh:
Btw, I finally learned where these irritating parameters are coming from. Apparently, VST3 doesn't support MIDI CC/etc. events directly 🤯, so VST3 plug-ins create a bunch of dummy parameters to which DAWs can map MIDI CCs/etc. In Juce, this is done automatically using the JUCE_VST3_EMULATE_MIDI_CC_WITH_PARAMETERS macro. See [this post](https://forum.juce.com/t/what-is-the-intention-of-macro-juce-vst3-emulate-midi-cc-with-parameters/17858) for details.

PR #1383, opened 2026-04-27
Fix localisation for the Analyse Loudness menu.

### LINKED ISSUE #1382: Missing localization for several items in the Analyze Loudness menu
Pressing Ctrl+Shift+U in REAPER brings up the "Statistics for peak" menu (Analyze Loudness), but several items in it are missing localization strings, meaning they can't be translated.
Specifically:
In the top-level menu, aside from "Master", entries like "tracks", "items", and "Dry run project using the most recent render settings" are not localized.
Inside the "Master" submenu, "Master mix" and "Time selection" are also missing localization.
Additionally, none of the items in the other submenus are localizable.
It would be great if this could be fixed in a future update. Thanks!

jcsteh:
Previously, the addMenuItem and addSubMenu helper functions called translate(). However, that meant that makePot couldn't identify the strings that needed to be translated. To fix this, remove the translate() calls from those helpers and instead call translate() around each string.
Fixes #1382.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -5705,7 +5705,7 @@ static void addMenuItem(HMENU menu, int position, const char* label, UINT id, bo
 	info.cbSize = sizeof(MENUITEMINFO);
 	info.fMask = MIIM_TYPE | MIIM_ID | MIIM_STATE;
 	info.fType = MFT_STRING;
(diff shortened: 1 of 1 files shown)


jcsteh:
No idea how I missed this in code review. 😳

PR #1385, opened 2026-05-02
Report recordseparate config

ScottChesworth:
We've had a handful of requests over the years for report recording to be a dedicated setting. It's like a speaky red light, a safety check that you might want to keep on even though you're confident enough with keystrokes that you don't want play states talking all day long.

### DIFF
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -111,7 +111,7 @@ class Surface: public IReaperControlSurface {
 		}
 		// Calculate integer based transport state
 		int TransportState = (int)play | ((int)pause << 1) | ((int)rec << 2);
-		reportTransportState(TransportState);
+		reportTransportState(0, TransportState);
 	}
 
 	void SetRepeatState(bool repeat) final {
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -106,7 +106,7 @@ class Surface: public IReaperControlSurface {
 		if (play) {
 			cancelPendingMidiPreviewNotesOff();
 		}
-		int transportState = (int)play | ((int)pause << 1) | ((int)rec << 2);
+		const int transportState = (int)play | ((int)pause << 1) | ((int)rec << 2);
 		int previousTransportState = this->lastTransportState;
 		this->lastTransportState = transportState;
 		if (previousTransportState == transportState) {
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -106,12 +106,16 @@ class Surface: public IReaperControlSurface {
 		if (play) {
 			cancelPendingMidiPreviewNotesOff();
 		}
+		int transportState = (int)play | ((int)pause << 1) | ((int)rec << 2);
+		int previousTransportState = this->lastTransportState;
+		this->lastTransportState = transportState;
+		if (previousTransportState == transportState) {
+			return;
+		}
 		if (this->wasCausedByCommand()) {
 			return;
 		}
-		// Calculate integer based transport state
-		int TransportState = (int)play | ((int)pause << 1) | ((int)rec << 2);
-		reportTransportState(0, TransportState);
+		reportTransportState(previousTransportState, transportState);
 	}
 
 	void SetRepeatState(bool repeat) final {
@@ -447,6 +451,7 @@ class Surface: public IReaperControlSurface {
 	const int PARAM_PAN = -3;
 	int lastParam = PARAM_NONE;
 	map<MediaTrack*, uint8_t> trackCache;
+	int lastTransportState = GetPlayState() & (1 | 2 | 4);
 	double lastPlayPos = 0;
 	int lastMarker = -1;
 	int lastRegion = -1;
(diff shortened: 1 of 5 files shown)


jcsteh:
I'm a bit confused about what this is trying to do. On one hand, we want to separate reporting of the record state, which I understand. On the other hand, this also reports stop and pause if those things happen while recording. But isn't the main purpose of this to ensure that the user knows they are recording, since that is writing to the project and it might not be so easy to distinguish from play? Hearing stop and pause seems redundant if the user has made an explicit decision not to hear most transport messages. Presumably, they're confident enough in this situation to trust pause and stop based on their action and the audio response.

Also, this is going to be inconsistent with transport changes from surfaces, etc., since those don't handle before/after states.

ScottChesworth:
The intent is to hear everything around recording where the stakes are higher. Its useful because REAPER has some behaviours once recording that aren't obvious. EG hitting R toggles between record and play, hitting Control+Space from recording pauses, but then Space resumes recording instead of play. Wah? IMO useful to hear if you don't record often.

Hmm I'm not very familiar with surface code yet. I'll take a run at making it behave the same once we've agreed what behaviour we want.

jcsteh:
Fair enough. I'll take your word on this one; you're in a better position to know the kinds of things that commonly trip up users. I just wanted to make sure there was clear intent behind this one, as it does add a fair bit of complexity to hitherto simple code.

I'll take a closer look at this later. Right now, I'm really struggling to follow this code, but I can't work out whether it's because my brain is having a bad day (very likely) or whether it's just inherent, unavoidable complexity in the change being made.

ScottChesworth:
Point taken about adding complexity, happy to rewrite it if you can see a simpler approach.

jcsteh:
What I'm thinking of is an early return when we know we're not going to report something; i.e. separate the question of whether we should report this from the message we're reporting. Is it correct to say that we only want to report record related things if we were recording either before or after the command? I'm not sure that's correct, but if it is, we could do something like this:

```
bool shouldReport = settings::reportTransport;
if (before & 4 || after & 4) {
	// We are recording now, or we were recording and we just stopped. That means this change is related to recording.
	shouldReport = settings::reportRecord;
}
if (!shouldReport) {
	return;
}
```

The problem is that my brain isn't currently capable of following that logic the entire way through, so I don't know if this would actually work.

jcsteh:
Or you could write it like this if it seems clearer:

```
auto shouldReport = [&] () {
	if (before & 4 || after & 4) {
		// We are recording now, or we were recording and we just stopped. That means this change is related to recording.
		return settings::reportRecord;
	}
	return settings::reportTransport;
};
if (!shouldReport()) {
	return;
}
```

ScottChesworth:
> Is it correct to say that we only want to report record related things if we were recording either before or after the command?
> 
Yep, now I've tried it in situ, that's definitely better than what I had. I've refactored following your first suggestion (was easier to read here). Gonna try to get surfaces behaving the same as well.

ScottChesworth:
Think this is good to go if you're happy with it

src/reaper_osara.cpp
```diff
@@ -1940,31 +1940,38 @@ void postToggleMasterTrackVisible(int command) {
...
-	} else if (state & 4 && repeat == true) {
+	// Recording also sets the playing bit, so handle record before play.
+	} else if (after & 4 && repeat) {
 		s << translate("record") << ", " << translate("repeat on");
-	} else if (state & 4) {
+	} else if (after & 4) {
 		s << translate("record");
-	} else if (state & 1 && repeat == true) {
+	// Only report play here if recording is not active.
+	} else if (after & 1 && (after & 4) == 0 && repeat) {
```
jcsteh: I don't think these `after & 4` bits are necessary now because we already handled `after & 4` above and we returned early for any recording case that we aren't reporting.

src/controlSurface.cpp
```diff
@@ -106,12 +106,16 @@ class Surface: public IReaperControlSurface {
 		if (play) {
 			cancelPendingMidiPreviewNotesOff();
 		}
+		int transportState = (int)play | ((int)pause << 1) | ((int)rec << 2);
```
jcsteh: nit: `const int` while you're here.

jcsteh (changes requested):

src/reaper_osara.cpp
```diff
@@ -1940,31 +1940,38 @@ void postToggleMasterTrackVisible(int command) {
...
-	} else if (state & 1 && repeat == true) {
+	// Only report play here if recording is not active.
+	} else if (after & 1 && repeat) {
 		s << translate("play") << ", " << translate("repeat on");
-	} else if (state & 1) {
+	} else if (after & 1) {
 		s << translate("play");
-	} else {
+	// If none of the transport bits are set, REAPER is stopped.
+	} else if ((after & (1 | 2 | 4)) == 0) {
```
jcsteh, line 1969: I know you already merged this, but is there any reason this isn't just `} else {` like it was before, rather than `else if`? By this point, we should have returned already or handled any other case, leaving stop as the only remaining case.

ScottChesworth:
> is there any reason this isn't just } else { like it was before,?
> 
Broadly it's because I'm a silly billy. Sorry about that. Merging simplified branch now.

jcsteh:
It could be worse. You could be just learning about something fundamental like, I dunno, reverb pre-delay, after decades of messing about with audio. 😂

ScottChesworth:
I'm worried. We're gonna need someone in charge of the CI when they make reverb certification compulsory.

PR #1329, opened 2025-10-20
Report move track actions

ScottChesworth:

PR #1390, opened 2026-05-06
Report move track actions v2

ScottChesworth:
Prior PR #1329 got away from me, too complex for me to follow. I've started over with implementation that seems more readable here and simpler feedback. This time around I'm intentionally overriding some undesirable REAPER behaviour, preventing tracks from disappearing inside closed folders which would prevent users from examining context.

### DIFF
--- a/config/mac/reaper-kb.ini
+++ b/config/mac/reaper-kb.ini
@@ -53,8 +53,8 @@ KEY 5 66 16 0		 # Main : Shift+B : Track: Toggle FX bypass for master track
 KEY 1 112 40007 0		 # Main : F1 : Help: About REAPER
 KEY 41 76 40014 0		 # Main : Cmd+Control+L : Item: Copy loop of selected area of audio items
 KEY 9 80 40016 0		 # Main : Cmd+P : Options: Preferences...
-KEY 1 32813 40018 0		 # Main : Insert : OVERRIDE DEFAULT : Insert media files...
-KEY 9 73 40018 0		 # Main : Cmd+I : Insert media files...
+KEY 1 32813 40018 0		 # Main : Insert : OVERRIDE DEFAULT : Insert/import media files...
+KEY 9 73 40018 0		 # Main : Cmd+I : Insert/import media files...
 KEY 9 71 40032 0		 # Main : Cmd+G : OVERRIDE DEFAULT : Item grouping: Group items
 KEY 13 71 40033 0		 # Main : Cmd+Shift+G : OVERRIDE DEFAULT : Item grouping: Remove items from group
 KEY 5 71 40034 0		 # Main : Shift+G : OVERRIDE DEFAULT : Item grouping: Select all items in groups
@@ -252,7 +252,7 @@ KEY 32 41 41212 0		 # Main : Control+) : Grid: Set to 1/48 (1/32 triplet)
 KEY 32 38 41213 0		 # Main : Control+& : Grid: Set to 1/24 (1/16 triplet)
 KEY 32 37 41214 0		 # Main : Control+% : Grid: Set to 1/6 (1/4 triplet)
 KEY 32 60 41305 0		 # Main : Control+< : Item edit: Trim left edge of item to edit cursor
-KEY 24 62 41307 0		 # Main : Cmd+Opt+> : OVERRIDE DEFAULT : Item edit: Move right edge of item to edit cursor
+KEY 24 62 41307 0		 # Main : Cmd+Opt+> : OVERRIDE DEFAULT : Item edit: Move right edge of item to edit cursor (preserving item length)
 KEY 32 62 41311 0		 # Main : Control+> : Item edit: Trim right edge of item to edit cursor
 KEY 5 67 41383 0		 # Main : Shift+C : OVERRIDE DEFAULT : Edit: Copy items/tracks/envelope points (depending on focus) within time selection, if any (smart copy)
 KEY 5 88 41384 0		 # Main : Shift+X : OVERRIDE DEFAULT : Edit: Cut items/tracks/envelope points (depending on focus) within time selection, if any (smart cut)
@@ -283,8 +283,11 @@ KEY 9 114 42348 0		 # Main : Cmd+F3 : Reset all MIDI control surface devices
 KEY 5 117 42455 0		 # Main : Shift+F6 : OVERRIDE DEFAULT : FX: Toggle delta solo for last focused FX
 KEY 13 78 42460 0		 # Main : Cmd+Shift+N : OVERRIDE DEFAULT : Item properties: Normalize items (peak/RMS/LUFS)...
 KEY 21 117 42467 0		 # Main : Opt+Shift+F6 : FX: Clear delta solo for all project FX
+KEY 57 32 43355 0		 # Main : Cmd+Opt+Control+Space : Transport: Play (stop at end of selected items, ignore loop selection)
 KEY 32 47 43593 0		 # Main : Control+/ : Scrub: Toggle preference for one-shot segment scrub when moving edit cursor
 KEY 0 63 43617 0		 # Main : ? : OVERRIDE DEFAULT : Scrub: Invert looped-segment scrub range
+KEY 5 32801 43647 0		 # Main : Shift+Page Up : Track: Move tracks up
+KEY 5 32802 43648 0		 # Main : Shift+Page Down : Track: Move tracks down
 KEY 1 116 6 0		 # Main : F5 : OVERRIDE DEFAULT : Track: Toggle mute for selected tracks
 KEY 255 3048 65535 0		 # Main :  : No-op (no action)
 KEY 255 3304 65535 0		 # Main :  : No-op (no action)
@@ -725,5 +728,4 @@ KEY 21 32806 42177 32063		 # Media Explorer : Opt+Shift+Up : Preview: increase v
 KEY 21 32808 42178 32063		 # Media Explorer : Opt+Shift+Down : Preview: decrease volume by 1 dB
 KEY 13 82 42184 32063		 # Media Explorer : Cmd+Shift+R : Preview random file in current folder/database
 KEY 5 32805 42185 32063		 # Media Explorer : Shift+Left : OVERRIDE DEFAULT : Preview: Play reversed
-KEY 9 8 42239 32063		 # Media Explorer : Cmd+Backspace : Preview: reset pitch
-KEY 5 32814 42317 32063		 # Media Explorer : Shift+Delete : File: Delete (no trash bin)
+KEY 9 8 42239 32063		 # Media Explorer : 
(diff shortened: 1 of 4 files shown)


src/reaper_osara.cpp
```diff
@@ -962,6 +962,114 @@ const char* (*NF_GetSWSTrackNotes)(MediaTrack* track) = nullptr;
 
 bool shouldMoveToAutoItem = false;
 
+string getTrackNameOrNumber(MediaTrack* track) {
```
jcsteh: I feel like we do this in a few places. Can we reuse this anywhere else? Fine if we can't, just wondering whether you considered it.

src/reaper_osara.cpp
```diff
@@ -962,6 +962,114 @@ const char* (*NF_GetSWSTrackNotes)(MediaTrack* track) = nullptr;
...
+		return format(translate("{} {}"), getTrackNameOrNumber(track), folderType);
+	}
+	return getTrackNameOrNumber(track);
+}
+
+string formatInsideFolder(MediaTrack* track) {
+	// Translators: Used to report the folder containing a track after it has moved.
+	// The first {} will be replaced with the name or number of the folder.
+	// The second {} will be replaced with "folder" or "nested folder".
+	return format(translate("inside {} {}"),
```
jcsteh: Please use named translation arguments when there are multiple translatable bits, rather than multiple unnamed arguments. This makes it much easier for translators to follow, as well as making it possible to reorder things if needed in a particular language. Example:

```
		// Translators: Reported when moving to a time signature change. {num} will
		// be replaced with the time signature numerator. {denom} will be replaced
		// with the time signature denominator. For example: "time sig 6/8".
		s << " " << format(translate("time sig {num}/{denom}"),
			"num"_a=sigNum, "denom"_a=sigDenom);
```

jcsteh (changes requested):
Thanks. I haven't actually reviewed this properly yet, but here's an initial thing that definitely needs to be adjusted in a few places. Feel free to leave this until I do a full review if you'd rather catch it all in one pass though.

jcsteh:
Also, this needs a readme update.

src/reaper_osara.cpp
ScottChesworth: It became possible to reuse in some places *after* your question made me realise I'd been testing without track numbers being reported lol. Refactored to obey that now.

src/reaper_osara.cpp
ScottChesworth: Done that

src/reaper_osara.cpp
```diff
@@ -962,6 +972,104 @@ const char* (*NF_GetSWSTrackNotes)(MediaTrack* track) = nullptr;
 
 bool shouldMoveToAutoItem = false;
 
+const char* getTrackFolderType(MediaTrack* track) {
```
jcsteh, line 975: I keep wondering whether this could be unified with formatFolderState, but I guess you don't want it reporting some of the details that function reports and trying to make some details optional might just make the code less readable rather than more readable. Just dropping this comment in case it gives you any good ideas, but if not, just leave it as is.

src/reaper_osara.cpp
```diff
@@ -962,6 +972,104 @@ const char* (*NF_GetSWSTrackNotes)(MediaTrack* track) = nullptr;
...
+	}
+}
+
+void openClosedFolderAndParents(MediaTrack* folder) {
+	for (MediaTrack* track = folder; track; track = GetParentTrack(track)) {
+		if (GetMediaTrackInfo_Value(track, "I_FOLDERDEPTH") != 1) {
+			continue;
+		}
+		int* const compacting = (int*)GetSetMediaTrackInfo(track, "I_FOLDERCOMPACT", nullptr);
+		if (!compacting || *compacting != 2) {
```
jcsteh: nit: Please simplify to:

`if (GetMediaTrackInfo_Value(track, "I_FOLDERDEPTH") != 2) {`

src/reaper_osara.cpp
```diff
@@ -962,6 +972,104 @@ const char* (*NF_GetSWSTrackNotes)(MediaTrack* track) = nullptr;
...
+	if (!prevTrack) {
+		return;
+	}
+	openClosedFolderAndParents(GetParentTrack(prevTrack));
+}
+
+void maybeOpenClosedFolderBeforeTrackMoveDown(MediaTrack* track) {
+	const int trackIndex = (int)GetMediaTrackInfo_Value(track, "IP_TRACKNUMBER") - 1;
+	MediaTrack* const nextTrack = GetTrack(nullptr, trackIndex + 1);
+	if (!nextTrack || GetMediaTrackInfo_Value(nextTrack, "I_FOLDERDEPTH") != 1) {
```
jcsteh, line 1066: Why do we need to check FOLDERDEPTH here but not in maybeOpenClosedFolderBeforeTrackMoveDown? Very likely I'm missing something - there's some subtlety here - but the asymmetry stuck out for me, so I think it at least needs a comment if it's legit.

src/reaper_osara.cpp
```diff
@@ -5661,6 +5757,76 @@ void cmdVideoWindowVisibility(Command* command) {
...
+		maybeOpenClosedFolderBeforeTrackMoveUp(track);
+	} else {
+		maybeOpenClosedFolderBeforeTrackMoveDown(track);
+	}
+	Main_OnCommand(command->gaccel.accel.cmd, 0);
+	MediaTrack* const newParent = GetParentTrack(track);
+	const int trackIndex = (int)GetMediaTrackInfo_Value(track, "IP_TRACKNUMBER") - 1;
+	const bool atTopOfTrackList = trackIndex == 0;
+	const bool atBottomOfTrackList = trackIndex == CountTracks(nullptr) - 1;
+	// Adjacent track is where we're coming from.
```
jcsteh: I find this comment slightly confusing. Do you mean that this is the track that now occupies the position that our origin track had before the move?

Aside from clarifying the comment, I think an example might not go astray here for thick brains. Something like:

"For example, say we have 3 tracks: a, b and c. If we move b up, the order is now b, a, c. adjacentTrack is a."

That's assuming my understanding is correct. This hurts my head for some reason...

src/reaper_osara.cpp
```diff
@@ -5661,6 +5757,76 @@ void cmdVideoWindowVisibility(Command* command) {
...
+			s << translate("top of track list");
+		} else {
+			// Translators: Reported when trying to move a track down when it is
+			// already at the bottom of the track list.
+			s << translate("bottom of track list");
+		}
+	} else if (atTopOfTrackList) {
+		// Translators: Reported when moving a track to the top of the track list.
+		// {track} will be replaced with a track reference; e.g. "vocal",
+		// "3", "3 vocal", "drums folder" or "3 drums nested folder".
```
jcsteh, line 5802: I normally try to provide an example of how the string would appear when fully expanded. For example: "top of track list, above 3 drums nested folder"

src/reaper_osara.cpp
```diff
@@ -5661,6 +5757,76 @@ void cmdVideoWindowVisibility(Command* command) {
...
+	if (!adjacentTrack) {
+		if (up) {
+			// Translators: Reported when trying to move a track up when it is
+			// already at the top of the track list.
+			s << translate("top of track list");
+		} else {
+			// Translators: Reported when trying to move a track down when it is
+			// already at the bottom of the track list.
+			s << translate("bottom of track list");
+		}
```
jcsteh, line 5795: I think some early returns would really help with readability here.

It looks like this `if (!adjacent)` block doesn't use any of the code below except for the outputMessage. So just use outputMessage(translate( directly and return early. That also means you can get rid of the `else` for this block and unindent everything below it. It also means you can define the ostringstream below.

src/reaper_osara.cpp
```diff
@@ -5661,6 +5757,76 @@ void cmdVideoWindowVisibility(Command* command) {
...
+		s << format(translate("top of track list, above {track}"),
+			"track"_a=formatTrackReference(adjacentTrack));
+	} else if (atBottomOfTrackList) {
+		// Translators: Reported when moving a track to the bottom of the track list.
+		// {track} will be replaced with a track reference; e.g. "vocal",
+		// "3", "3 vocal", "drums folder" or "3 drums nested folder".
+		s << format(translate("bottom of track list, below {track}"),
+			"track"_a=formatTrackReference(adjacentTrack));
+	} else if (adjacentTrack == newParent) {
+		primaryMessageIncludesParent = true;
```
jcsteh: Similar early return here. primaryMessageIncludesParent  prevents everything below except outputMessage. So remove the primaryMessageIncludesParent  variable, outputMessage directly and early return. This way, you don't have to wonder about whether anything else can be appended to the string for this code path; it's very clear that we're done.

src/reaper_osara.cpp
```diff
@@ -5661,6 +5757,76 @@ void cmdVideoWindowVisibility(Command* command) {
...
+		s << formatTrackMoveInsideFolder(up, newParent, contextTrack);
+	} else {
+		s << formatTrackMoveRelative(up, adjacentTrack);
+	}
+	if (newParent && !primaryMessageIncludesParent && newParent != adjacentTrack) {
+		// Translators: Appended after the new position when moving a track inside a folder.
+		// {folder} will be replaced with a folder reference; e.g. "guitars folder"
+		// or "3 guitars nested folder".
+		s << format(translate(", {folder}"),
```
jcsteh: For comma separation like this, I normally don't use a translation string because it's just a separator and because I worry translators will skip the comma. So:

`s << ", " << formatInsideFolder(newParent);`

It isn't ideal, but I don't think the translation string helps here either.

jcsteh (commented):
Sorry, this is probably going to take a few rounds of review. It kinda hurts my head, but some of that is just the complexity of what we're trying to do. Hopefully, these comments will make it easier for my brain to parse and hopefully other people's as well.

config/mac/reaper-kb.ini
```diff
@@ -283,8 +283,11 @@ KEY 9 114 42348 0		 # Main : Cmd+F3 : Reset all MIDI control surface devices
 KEY 5 117 42455 0		 # Main : Shift+F6 : OVERRIDE DEFAULT : FX: Toggle delta solo for last focused FX
 KEY 13 78 42460 0		 # Main : Cmd+Shift+N : OVERRIDE DEFAULT : Item properties: Normalize items (peak/RMS/LUFS)...
 KEY 21 117 42467 0		 # Main : Opt+Shift+F6 : FX: Clear delta solo for all project FX
+KEY 57 32 43355 0		 # Main : Cmd+Opt+Control+Space : Transport: Play (stop at end of selected items, ignore loop selection)
```
MarcoZehe: This will conflict with a VoiceOver command to open the menu of custom actions available on the item under the VoiceOver cursor. If there is no action defined, it will simply issue an error bonk, but still capture that key command. You could instead use a combination of cmd, option, control and shift, but without control and option being in the combination together.

ScottChesworth: Oops, that's an incorrect binding. I've got a new person helping me out with the Mac key map tasks, will get it fixed to be Command+Control+Space before this goes into a snapshot. Thanks for flagging it.

src/reaper_osara.cpp
ScottChesworth: Made a start, didn't like how it was effecting readability so I've left things how they are.

src/reaper_osara.cpp
ScottChesworth: I think the anchor was stail by the time I got to this, couldn't figure out what you meant. Shout if it's still a thing.

ScottChesworth:
Not sure how to re-request review so @jcsteh, have a gander whenever you get a sec. Think it's in better shape now.

jcsteh (approved):
This is much easier to follow, thank you. I've fixed some tiny nits on top of that, but otherwise, this is good to merge.

jcsteh:
Zarquan damn it, forgot to clear the description field when merging.

jcsteh:
Fixed and force pushed as e526083966ef4f23c297b2ddd46467f6467d3775.

PR #1391, opened 2026-05-12
Fix track movement feedback when the track can't move further up or down.

jcsteh:
For example, if you had two tracks, you were positioned on track 1 and you pressed shift+pageUp, OSARA would report "top of track list, above 2". However, based on this translators comment:

```
			// Translators: Reported when trying to move a track up when it is
			// already at the top of the track list.
```

I think it's supposed to just report "top of track list".

@ScottChesworth, I'm tagging you for review because I want to make sure this was your intent.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -5774,16 +5774,13 @@ void cmdMoveTracks(Command* command) {
 	} else {
 		maybeOpenClosedFolderBeforeTrackMoveDown(track);
 	}
+	const int origTrackIndex = (int)GetMediaTrackInfo_Value(track, "IP_TRACKNUMBER") - 1;
 	Main_OnCommand(command->gaccel.accel.cmd, 0);
 	MediaTrack* const newParent = GetParentTrack(track);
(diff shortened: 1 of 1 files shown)


ScottChesworth:
The comment was inadequate, sorry. Code was doing as intended. Figured it could be helpful to hear what you're above when you're doing a lot of rearranging.
Assuming you're ok with what I intended, I guess the easiest fix is to improve the comment?

jcsteh:
Maybe? I'm still kinda confused. As it stands in snapshots, the only time this particular code path would occur is if there is only 1 track in the project. So why differentiate between top and bottom at all in that case?

> Figured it could be helpful to hear what you're above when you're doing a lot of rearranging.

Even if the track isn't actually moving anywhere because it's already at the top?

ScottChesworth:
Duh, you're right. I didn't read close enough before replying. Thanks for fixing it.

jcsteh:
Also, when I merged #1390, I forgot to mention the key map update. Sorry for any support churn that has caused/will cause.

ScottChesworth:
That does it, time to put Claude in charge.
really though, nobody has shouted down the phone at me yet. Give it time...

PR #1394, opened 2026-05-28
Report when momentary key map override ends

ScottChesworth:


src/reaper_osara.cpp
```diff
@@ -2707,17 +2707,77 @@ string getShortenedAltSectionName(int sectionId) {
...
 void postToggleOverrideToAltN(int command) {
+	stopTrackingMomentaryOverride();
 	int sectionId = command - CMD_TOGGLE_OVERRIDE_TO_ALT1
 		+ MAIN_ALT1_SECTION;
 	outputMessage(getShortenedAltSectionName(sectionId));
 }
 
+void reportCurrentKeyMapOverride() {
+	KbdSectionInfo* section = SectionFromUniqueID(MAIN_SECTION);
+	if (GetToggleCommandState2(section, CMD_SET_OVERRIDE_TO_DEFAULT) > 0) {
```
jcsteh: I'm a bit conflicted about this. On one hand, I understand what you're trying to do here: once the override is cancelled, we want to tell the user what key map they've reverted to. On the other hand, the key point here is that the momentary override is cancelled; the user should already have known what key map they were in when they triggered the momentary override. Saying something like "momentary cancelled" would simplify things a lot here. That said, if you feel strongly about doing it this way, I can live with it; it's not that much code.

src/reaper_osara.cpp
```diff
@@ -2707,17 +2707,77 @@ string getShortenedAltSectionName(int sectionId) {
...
+		outputMessage(translate("default key map"));
```
jcsteh: What's the difference between "default key map" and "main key map" from the user's perspective?

src/reaper_osara.cpp
```diff
@@ -2707,17 +2707,77 @@ string getShortenedAltSectionName(int sectionId) {
...
+	outputMessage(translate("main key map"));
+}
+
+void pollMomentaryOverride() {
+	if (!activeMomentaryOverrideCommand) {
+		return;
+	}
+	// REAPER clears momentary key map overrides internally, without firing a
+	// command hook, so watch the toggle state until it turns off.
+	if (GetToggleCommandState2(SectionFromUniqueID(MAIN_SECTION),
```
jcsteh: I wonder if we could clean this up a little by using the override interval from the configuration; i.e. reaper.ini -> [REAPER] -> kbd_override_len. I'm not sure whether you can fetch this using get_config_var or whether you have to use GetPrivateProfileString. Note that if the key is missing, it defaults to 1000.

Anyway, if we do that, we shouldn't need to poll. We also don't need to track what command initiated the override. We can just use a single CallLater and assume that the override has now been cancelled. If something else cancels it before the timeout elapses, we can cancel the CallLater elsewhere as you already do.

src/reaper_osara.cpp
```diff
@@ -6376,6 +6437,14 @@ bool handleToggleCommand(KbdSectionInfo* section, int command, int val, int valH
 		if (message) {
 			outputMessage(translate(message));
 		}
+		if (newState &&
+				CMD_MOMENTARILY_SET_OVERRIDE_TO_DEFAULT <= command &&
+				command <= CMD_MOMENTARILY_SET_OVERRIDE_TO_ALT16) {
+			trackMomentaryOverride(command);
```
jcsteh: Why do we need this? Shouldn't this be handled by postMomentarilySetOverrideToAltN?

If we don't need this, combined with the elimination of polling above, we can move trackMomentaryOverride and the CallLater function into postMomentarilySetOverrideToAltN, thus keeping most of this in one place.

jcsteh (changes requested):
Thanks.

src/reaper_osara.cpp
ScottChesworth: No idea, figured I'd add feedback as I noticed REAPER exposes both actions separately: "Main action section: Clear any override" and "Main action section: Set override to default". However it seems like default is also the main map and I don't see a way to change that, so I guess it doesn't need to stay.

src/reaper_osara.cpp
ScottChesworth: Changed to GetPrivateProfileInt, no more polling and one callLater. Thanks, feels way cleaner.

src/reaper_osara.cpp
ScottChesworth: Yep, got rid of the duplication.

src/reaper_osara.cpp
ScottChesworth: I figured we may as well provide more information seeing as it was easy to do.

src/reaper_osara.cpp
```diff
@@ -3332,7 +3368,7 @@ map<pair<int, int>, ToggleCommandMessage> TOGGLE_COMMAND_MESSAGES = {
 	{{MAIN_SECTION, 1157}, {_t("enabled snap"), _t("disabled snap")}}, // Options: Toggle snapping
 	// Reducing verbeage when toggling and momentarily switching to alt keymap layers (Reaper 7)
 	{{MAIN_SECTION, 24800}, {_t("main key map"), nullptr}}, // Main action section: Clear any override
-	{{MAIN_SECTION, 24801}, {_t("default key map"), nullptr}}, // Main action section: Set override to default
+	{{MAIN_SECTION, 24801}, {_t("main key map"), nullptr}}, // Main action section: Set override to default
 	{{MAIN_SECTION, 24802}, {_t("recording key map"), nullptr}}, // Main action section: Toggle override to recording
 	{{MAIN_SECTION, 24851}, {_t("default momentary"), nullptr}}, // Main action section: Momentarily set override to default
```
jcsteh: Ah damn it, sorry, now I think I understand why you originally had the duplication in handleToggleCommand to start tracking. I think as it stands now, these two commands - 24851 and 24852 - won't track momentary switching, because they don't use postMomentarilySetOverrideToAltN. I don't love the idea of handling this in handleCommand, but it might be the cleanest way. Alternatively, we could make postMomentarilySetOverrideToAltN into postMomentarilySetOverride and have it handle 24851 and 24852 as well, rather than putting those messages here. That would also mean we're handling all momentary overrides in one place.

src/reaper_osara.cpp
```diff
@@ -6376,6 +6412,10 @@ bool handleToggleCommand(KbdSectionInfo* section, int command, int val, int valH
 		if (message) {
 			outputMessage(translate(message));
 		}
+		if (CMD_CLEAR_OVERRIDE <= command &&
```
jcsteh: We have a similar pattern here where we handle most of these commands in postToggleOverrideToAltN, but 24800 through 24802 are exceptions, and so we have to handle those specially here. Again, I can live with this, but I'd be curious to get your thoughts as to whether it feels cleaner to instead refactor these exceptions into postToggleOverrideToAltN, renaming that to postToggleOverride or similar.

My thinking is that handleCommand, handleToggleCommand and friends really should be as generic as possible, leaving special cases to specific command hooks. Then again, these commands kinda are special, so... 🤷‍♂️

jcsteh (approved):
Thanks. I can live with this as is, but I had a realisation that may or may not be cleaner. If you disagree, feel free to merge as is.

jcsteh:
Oh, and regardless of which approach, I think 24851 and 24852 need fixing before merge.

src/reaper_osara.cpp
ScottChesworth: All in one place as suggested.

src/reaper_osara.cpp
```diff
@@ -2707,25 +2707,92 @@ string getShortenedAltSectionName(int sectionId) {
...
+			reportAltOverride(command);
+			return;
+		}
+	}
+	outputMessage(translate("main key map"));
+}
+
+void postToggleOverride(int command) {
+	stopTrackingMomentaryOverride();
+	if (command == CMD_CLEAR_OVERRIDE ||
```
jcsteh: Can we replace all of the code below with a simple call to reportCurrentKeyMapOverride? Or am I missing something?

src/reaper_osara.cpp
```diff
@@ -2707,25 +2707,92 @@ string getShortenedAltSectionName(int sectionId) {
...
+}
+
+int getMomentaryOverrideTimeout() {
+	int timeout = GetPrivateProfileInt("REAPER", "kbd_override_len", 1000,
+		get_ini_file());
+	return timeout >= 0 ? timeout : 1000;
+}
+
+void postMomentarilySetOverride(int command) {
 	// Translators: Reported when using "Main action section: Momentarily set
```
jcsteh: This translators comment needs to be moved down to the string it is associated with.

jcsteh (changes requested):

src/reaper_osara.cpp
ScottChesworth: Ya know, it's lucky one of us is good at this lol.

src/reaper_osara.cpp
ScottChesworth: Oops, fixed.

jcsteh (approved):
Thanks. I made a few more tweaks myself.

jcsteh:
Haha, thanks. Sorry, got distracted by a kid in the middle of merging.

Issue #977, opened 2023-12-19
Restructure the code a bit to make it easier for newer contributors


jcsteh:
I've been thinking a bit lately about how we might restructure the code base a little to make it easier for newer contributors to find things and add new things, especially for common tasks like intercepting a new REAPER command, adding a new OSARA command, etc. The code is reasonably clean already for the most part. However, reaper_osara.cpp in particular is getting very large and you do have to hunt around a bit to find the various tables, the best place to put new functions, etc.

Here are things I think are probably no-brainers and fairly straightforward:

1. Move all the commands for the main section out of reaper_osara.cpp into mainCommands.cpp. envelopeCommands.cpp would remain separate.
2. Rename reaper_osara.cpp to core.cpp or similar. This would make it clear that this is the core engine of the plugin and not individual commands, etc.

The command maps should be split out as well. One super controversial idea I've been mulling over is whether to auto generate the commands map from decorator macros near the commands. For example:

```
defineOsaraCommand(MAIN_SECTION, _t("OSARA: About"), "OSARA_ABOUT")
void cmdAbout(Command* command) {
...
}

interceptReaperCommand(MAIN_SECTION, 40285) // Track: Go to next track
void cmdGoToNextTrack(Command* command) {
...
}
```

The nice thing about this is that it keeps command registration near the command function. You don't have to look for a separate command map. The disadvantage is that it's magic: it depends on the build system auto-generating code.

@RDMurray and @LeonarddeR, I'd particularly love your thoughts about the command map magic, but also about anything else here or any other things you think would make life easier for yourselves or newer contributors. This is very much a brainstorming exercise right now; I don't think I'll realistically start working on any of this immediately and I think we should have a pretty solid plan before making any serious changes.

@ScottChesworth and @jennykbrennan, please feel free to flag anything here that would really help you (or would have really helped you) as you're learning the code.

LeonarddeR:
I actually like all the ideas you propose here. I'd probably start with the macro magic, then after that's done, it is probably a bit easier to restructure.

jcsteh:
Intriguing. I thought you'd hate the macro magic idea for some reason. Hahaha.

LeonarddeR:
I have nothing against macros when they make things easier and there's no feasible alternative. That seems to be the case here.
Do you have any thought about how the macro would behave? I assume we can change the several arrays into a constexpr vector, then let the macro push to that?

jcsteh:
[...]

PR #1395, opened 2026-05-29
dev: Make command hooks (cmd functions) take an int instead of a Command struct pointer.

jcsteh:
This simplifies things like this:
`Main_OnCommand(command->gaccel.accel.cmd, 0);`
to this:
`Main_OnCommand(command, 0);`
It also makes these more consistent with post command hooks (post functions). Re #977.


jcsteh:
@ScottChesworth, can you think of any reason not to do this? This is part of my thinking on simplifying some of the code, albeit a very early part. I figure it makes sense to tackle simplification in pieces rather than trying to do it all at once.

ScottChesworth:
Looks good from here. That said, I'm not sure I'm experienced enough to know if it was a bad idea.

PR #1396, opened 2026-05-31
install:add translatable NSIS strings

dpy013:
Move OSARA-specific installer text out of hard-coded English and into NSIS LangString entries so the custom installer UI can participate in localization while standard MUI button text continues to come from NSIS language files.

Add installer string extraction to the POT generation pipeline, generate build/installerStrings.nsh from locale/*.po during installer builds, and wire the generated file into the NSIS build so translated installer strings can be emitted automatically when translations are available.

Also normalize escaped quotes when extracting installer strings so POT entries and generated installer lookups stay in sync for strings containing embedded quotes.


jcsteh:
Thanks for working on this.

I have some high level concerns before we get into the specifics of the code:

1. This adds strings to the main .po file. However, the main .po file will be loaded into memory for OSARA itself, even though OSARA itself doesn't need the installer strings. In fact, the installer strings don't need to be in any .po file at all that is distributed with OSARA, since they are already part of the installer. So, I think these should be split into separate files; e.g. locale/winInstaller/osara_*.pot, locale/winInstaller/*.po. That will require some change to makePot and crowdinSync, though.
2. This only covers the Windows installer. While we don't have to do both at once, if localising the installer is something users want, we need to think about how we're going to handle this for the Mac installer eventually. The simplest path there is probably to move away from NSIS and instead use @mulcahy1000's unified C++ installer app in #1295. There are a few obstacles there though:
    - We need to figure out how to make it self-extracting and self-running on Windows (no intermediate zip archive).
    - We need to figure out how to manage uninstallation on Windows.

dpy013:
Thanks, that makes sense. For this PR, I'd prefer to keep the scope focused on Windows first, since the current work
  is specifically about localizing the NSIS installer. I agree that we should also think about the Mac side, but I don't
   currently have a Mac test environment, so I'm not in a good position to design or validate that part yet.
   My thinking is to land the Windows-side changes first, then revisit Mac installer localization in a follow-up once I
  have a practical way to test it. If we eventually move toward the unified C++ installer, that may also give us a
  better path for cross-platform support.

jcsteh:
That seems reasonable. I think the concern about the .po file is still relevant here though, especially if this is going to be Windows only; it's definitely wasteful to carry those strings around in memory on mac when they will never be used there, not even in the installer.

PR #1398, opened 2026-06-01
add Linux x86-64 support, fixes #176

### LINKED ISSUE #176: Creating an OSARA build for Linux?
Hello guys,
I'm curious if it would be possible to begin a project for OSARA to be used on Linux. As you guys no doubt know, Reaper has released experimental builds for Linux. The problem is that you don't get (so far as I know) OSARA and SWS&SM on there, so for the visually impaired, using Reaper in general would be a challenge. I'd be happy to contribute and get the ball rolling, as it were, but I haven't done very much Linux programming, so don't know the equivalent API calls. If someone could either get the project going or give me resources and help me out with the project, that would be great. While we won't have SWS&SM on Linux if this works, we'll at least have OSARA, which would be a huge step towards music production and audio editing with Reaper on Linux.

RDMurray:
This is basic support for Linux on x86-64. It requires the libSwell.so included with Reaper to be replaced with the accessible branch from https://github.com/RDMurray/WDL

fixes #176 

I haven't implemented an installer yet.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [windows-2025-vs2026, macos-latest]
+        os: [windows-2025-vs2026, macos-latest, ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - name: set env
@@ -39,8 +39,13 @@ jobs:
           fi
           if [ "$RUNNER_OS" = "Windows" ]; then
             echo os=windows >> "$GITHUB_ENV"
-          else
+            echo keymapOs=windows >> "$GITHUB_ENV"
+          elif [ "$RUNNER_OS" = "macOS" ]; then
             echo os=mac >> "$GITHUB_ENV"
+            echo keymapOs=mac >> "$GITHUB_ENV"
+          else
+            echo os=linux >> "$GITHUB_ENV"
+            echo keymapOs=windows >> "$GITHUB_ENV"
           fi
       - uses: actions/checkout@v4
         with:
@@ -48,9 +53,13 @@ jobs:
       # There's no point in building if we're going to fail the build because of
       # an unsorted key map, so we do this as early as we can.
       - name: check key map
-        run: python tools/sortKeymap.py -t config/${{ env.os }}/reaper-kb.ini
+        run: python tools/sortKeymap.py -t config/${{ env.keymapOs }}/reaper-kb.ini
       - name: setup
         run: pip install scons
+      - name: Linux setup
+        if: ${{ runner.os == 'Linux' }}
+        # We need php for swell_resgen and xz-utils for the installer tarball.
+        run: sudo apt-get update && sudo apt-get install -y clang php-cli xz-utils
       - name: SCons MSVC Cache
         if: ${{ runner.os == 'Windows' }}
         uses: actions/cache@v4
@@ -300,6 +309,14 @@ jobs:
           # after the first wildcard pattern"
           # See also https://github.com/actions/upload-artifact/issues/248
           path: installer/mac/*.app
+      - name: upload PR build Linux
+        if: ${{ github.event_name == 'pull_request' && runner.os == 'Linux' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: osara_linux_${{ env.version }}
+          path: installer/osara_*.tar.xz
+          # The tarball is already compressed. Don't try to compress it again.
+          compression-level: 0
       - name: upload debug symbols Win
         if: ${{ runner.os == 'Windows' }}
         uses: actions/upload-artifact@v4
@@ -313,6 +330,7 @@ jobs:
       # outputs won't be set if the value is empty.
       winInstallerUrl: ${{ steps.getBuildUrl.outputs.windowsUrl }}
       macInstallerUrl: ${{ steps.getBuildUrl.outputs.macUrl }}
+      linuxInstallerUrl: ${{ steps.getBuildUrl.outputs.linuxUrl }}
   publish:
     # This job updates the website with the new readme and snapshots.
     if: ${{ github.event_name == 'push' }}
@@ -330,6 +348,7 @@ jobs:
           version: ${{ needs.build.outputs.version }}
           winUrl: ${{ needs.build.outputs.winInstallerUrl }}
           macUrl: ${{ needs.build.outputs.macInstallerUrl }}
+          linuxUrl: ${{ needs.build.outputs.linuxInstallerUrl }}
         run: python ci/buildSite
       - name: upload
         uses: actions/upload-pages-artifact@v3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,13 +39,10 @@ jobs:
           fi
           if [ "$RUNNER_OS" = "Windows" ]; then
             echo os=windows >> "$GITHUB_ENV"
-            echo keymapOs=windows >> "$GITHUB_ENV"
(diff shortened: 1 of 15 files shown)


RDMurray:
The build can be downloaded at https://github.com/jcsteh/osara/actions/runs/26789225657/artifacts/7345740584

src/archBuild_sconscript
```diff
@@ -121,6 +121,30 @@ else: # Mac
...
+	env.Append(CXXFLAGS=("-std=c++20", "-Werror"))
+	env.Append(CPPDEFINES="SWELL_PROVIDED_BY_APP")
+	sources.append(env.SharedObject("swell-modstub-generic",
+		swellDir.File("swell-modstub-generic.cpp")))
+	sources.append("linuxa11y.cpp")
+	libs = ["dl"]
+	# We don't want the "lib" prefix for our shared library.
+	env["SHLIBPREFIX"] = ""
+
+	# SWELL uses the same generated resource file names on generic platforms.
```
jcsteh: There seems to be a lot of common stuff for SWELL here on Mac and Linux. I wonder if we can unify it? I might be missing subtle differences though. Even if we can do it, I can live with it as is for now; we could do it in a follow-up.

src/linuxa11y_wrapper.h
```diff
@@ -0,0 +1,12 @@
+#ifndef linuxa11y_wrapper_h
+#define linuxa11y_wrapper_h
+
+#include <string>
+
+namespace LinuxA11y {
```
jcsteh: nit: Please use linuxA11y.h, linuxA11y.cpp, namespace linuxA11y. I realise this is inconsistent with the Mac stuff, but that's an exception to the rule everywhere else which I didn't catch at the time.

src/linuxA11y.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+	bool init() {
+		announceFn = reinterpret_cast<AnnounceFn>(
+			SWELL_ExtendedAPI("ACCESSIBILITY_ANNOUNCER", nullptr));
+		return announceFn != nullptr;
+	}
+
+	void announce(const std::string& message, bool interrupt) {
+		if (announceFn) {
+			announceFn(message.c_str(), interrupt ? 1 : 0);
```
jcsteh, line 27: Out of curiosity, does this extension use an ATK notification signal? I expected we might do this bit within OSARA, even though SWELL has to handle the UI, but I guess I can understand why you'd put this in SWELL given that you have to do the rest of the ATK stuff there.

jcsteh (changes requested):
Thanks for working on this. This looks really clean and solid. I realise this is a draft, but I thought I'd do a high level review anyway.

src/linuxA11y.cpp
RDMurray: I'm using accesskit in the swell fork. It adds a live region label to the accessible tree that is updated by this function. I would prefer to do it in OSARA, but haven't fully researched if it is possible and how to send at-spi notifications directly. I can definitely look into it, this was just the easiest way to get it working.

RDMurray:
I suppose the main question is whether to merge this at all? Maybe we should say it isn't supported and shouldn't be discussed on RWP or Whatsapp until it is more stable? OSARA works well, it's swell that still needs work, especially around keyboard handling. 

WDL/wingui controlls aren't accessible at all because they are outside swell and statically compiled into Reaper. I don't know how much that affects, but definitely toolbars and everything in the main window.

I can't see Cockos merging a huge vibe coded a11y layer into swell and wingui, but maybe in the future if I can clean it up. Perhaps they would be open to dynamically linking wingui for Linux builds.

RDMurray:
I'd like to add arm64 support as well. Reaper on Raspberry Pi or similar could be really useful for portable recording setups. Android phones might even be possible with Termux.

src/linuxA11y.cpp
jcsteh: Fair enough. This can potentially be dealt with in a follow-up.

src/archBuild_sconscript
```diff
@@ -38,6 +39,24 @@ sources = [
 	"updateCheck.cpp",
 ]
 
+swellDir = env.Dir("#include/WDL/WDL/swell")
+
+def configureSwell():
+	env.Append(CPPPATH=(swellDir,))
+	env.Append(CPPDEFINES="SWELL_PROVIDED_BY_APP")
+	# We don't want the "lib" prefix for our shared library.
```
jcsteh, line 47: This isn't really related to SWELL, but it is common to both Linux and Mac. You could either move this out and duplicate it, or rename this to configureUnix or configureUnixCommon or something.

jcsteh:
> I suppose the main question is whether to merge this at all?

Yeah, it's definitely a tricky one. As you say, all the OSARA parts work, but most users won't get the distinction between the OSARA parts and the other parts needed to make REAPER accessible. If we have an official Linux build of OSARA, does that send a confusing message and potentially result in a lot of annoying support requests and/or bug reports? Hmm.

I guess we could merge it except for the buildSite changes, but getting download links would be annoying. Perhaps we have it add an additional Linux section at the bottom which just has the Linux build along with instructions and big scary warnings about how it's experimental and might eat your dog?

LeonarddeR:
> I guess we could merge it except for the buildSite changes.

Is there anything against keeping the build changes while just not advertising them to the public?

.github/workflows/build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [windows-2025-vs2026, macos-latest]
+        os: [windows-2025-vs2026, macos-latest, ubuntu-latest]
```
LeonarddeR, line 25: I think there would be real potential in building this for Linux arm64 too, so a raspberry pi can become an accessible REAPER recorder, at least hypothetically.

jcsteh:
When I say buildSite, I'm referring to the buildSite script - i.e. the website builder - which by definition does advertise it to the public. :)

.github/workflows/build.yml
```diff
@@ -330,6 +345,7 @@ jobs:
           version: ${{ needs.build.outputs.version }}
           winUrl: ${{ needs.build.outputs.winInstallerUrl }}
           macUrl: ${{ needs.build.outputs.macInstallerUrl }}
+          linuxUrl: ${{ needs.build.outputs.linuxInstallerUrl }}
```
LeonarddeR, line 348: Or may be @jcsteh just meant to keep this out of the machinery.

sconstruct
```diff
@@ -55,6 +56,27 @@ else: # Mac
 	installer = env.Command("installer/osara_${version}.zip", ["installer/mac/build.sh", "build"],
 		[["$SOURCE", "$version"]])
 	configRc = "build/config.rc"
+else: # Linux
+	if platform.machine() not in ("x86_64", "AMD64"):
+		raise RuntimeError("Linux builds currently support x86_64 only")
```
LeonarddeR, line 61: pretty sure building on ubuntu-24.04-arm will work out of the box, unless you have already tested.

LeonarddeR (commented):
Love it, looking forward to test it later today.
Should the wdl submodule be bumped to your fork? I"m slightly puzzled why this works as it does without changing the submodule pointer.

RDMurray:
> Should the wdl submodule be bumped to your fork? I"m slightly puzzled why this works as it does without changing the submodule pointer.

We only use the submodule for headers, Reaper loads libswell.so at runtime. There's an extension in my WDL fork for updating an accesskit live region, but it uses the existing `SWELL_ExtendedAPI()` to resolve the function at runtime.

PR #1400, opened 2026-06-11
Use lld-link for arm64ec builds

LeonarddeR:
The `__sanitizer_annotate_contiguous_container` undefined symbol that
previously prevented lld-link from being used for arm64ec is resolved in Visual Studio, starting from the June update, since it uses LLVM  22.
We can merge this as soon as CI is at the june update, however this would break for people on Visual Studio <= 2026 june update.

@jcsteh Would you like to keep backwards compatibility or should we merge this as soon as github updated their runner?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1111,7 +1111,7 @@ To build OSARA, you will need:
 	See the note about submodules in the previous section.
 
 #### Windows
-- [Build Tools for Visual Studio 2026](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2026)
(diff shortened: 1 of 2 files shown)


LeonarddeR:
GitHub seems to have updated their Windows runner with the new Visual Studio version that includes Clang22, so this is ready to rock. Also verified on an ARM64 system.
Still awaits decision taking on whether we want to accept breakage with older versions of Visual Studio.

LeonarddeR:
Ugh, this is extremely weird. Looks like GitHub might be load balancing between an older and a newer image.

jcsteh:
Let's merge this. If anyone can't be bothered upgrading VS (this usually includes me 🤣), they can just `scons build/x86_64`.

PR #1402, opened 2026-06-14
FX Parameters dialog: Tweak detection of unnamed and potentially unusable parameters.

### LINKED ISSUE #1401: FX Parameters dialog filtering logic hides valid parameters
The filtering logic for "Include unnamed and potentially unusable parameters" in OSARA's FX Parameters dialog is flawed. It incorrectly hides valid parameters (such as effective CC parameters in ReaControlMIDI) while letting many useless parameters slip through (such as 4-digit # parameters in Kontakt, Control parameters in Waves plugins, purely numeric parameters, etc.).

## Steps to Reproduce
1. Add the ReaControlMIDI plugin to a track in REAPER.
2. In the plugin interface, check "CC Enable" and "CC Raw Mode Enable".
3. At this point, the default values for CC 1 Type, CC 2 Type, and CC 3 Type are CC 7 (Volume), CC 10 (Pan), and Pitch Wheel, respectively.
4. Close the FX window and press P on the track to open the FX Parameters dialog.
5. By default, "Include unnamed and potentially unusable parameters" is unchecked. In this state, CC 7 (Volume) and CC 10 (Pan) are not visible in the list; only Pitch Wheel can be seen.
6. Checking "Include unnamed and potentially unusable parameters" makes CC 7 (Volume) and CC 10 (Pan) appear.

## Other Unusable Parameters That Are Not Filtered
- Kontakt: For example, loading Flutes 2.nki from Cinematic Studio Woodwinds in Kontakt shows parameters starting with "#". Currently, OSARA only filters parameters with up to three digits (e.g., `#123`), but parameters ranging from "`#1000`" to "`#2047`" are not filtered. Similar issues exist with other Kontakt libraries.
- Waves: Plugins like RVerb have parameters starting with "Control" (presumably MIDI controllers) in the FX parameter list. Specific numbers display controller names, such as "Control 64" showing as "Sustain Pedal On/Off". In REAPER's FX window, these parameters are only found under the Param button -> Show track envelope -> MIDI sub-menu, and they are in an unusable (grayed out) state. However, OSARA does not filter them out.
- Keyscape (Spectrasonics): According to community feedback, there are parameters with purely numeric names (e.g., "511") that are not currently filtered. (I don't have this virtual instrument installed, so I'm unsure if there are other purely numeric parameters).
- EZkeys 2 (Toontrack): Community feedback indicates there are parameters starting with "MIDI" followed by numbers, such as "MIDI CC 15|129", which are also not filtered.

## Expected Results
I suggest the following adjustments to OSARA's filtering rules (though we should be cautious that these new rules don't accidentally filter out other useful parameters):
- Do not filter parameters named with "CC" followed by numbers, to prevent hiding valid parameters in plugins like ReaControlMIDI.
- Expand matching range: Extend the matching range for parameters starting with "#" from three digits to four digits (i.e., filter out `#1000-#9999`).
- Filter unusable parameters: Parameters that are mapped as unusable states within REAPER (like the Control parameters in Waves) should not be displayed in the FX Parameters list.
- Add new filtering rules: Introduce filtering for purely numeric parameters (e.g., "511") and parameters starting with "MIDI" followed by numbers (e.g., "MIDI CC 15|129").

jcsteh:
1. Treat any parameters in a section named "MIDI" as unusable. This filters out parameters like "Control 64"" in Waves plugins.
2. Treat parameters named, for example, "#1234" or "123" as unusable. This filters out parameters like "#1000" in Kontakt or "511" in Keyscape.
3. Only treat CC, etc. parameter names as unusable if the parameter is also non-automatable.

Fixes #1401.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -866,8 +866,8 @@ class FxParams: public ParamSource {
 			"|-"
 			// Example: "1234 -"
 			R"(|\d{1,4} -)"
-			// Example: "P123" or "#123"
-			R"(|[P#]\d{3})"
+			// Example: "123" or "P123" or "#1234"
+			R"(|[P#]?\d{3,4})"
 			// Any one of several strings...
 			"|(?:MIDI CC|MIDI Controller|Program Change|CC|Pitch Bend|Pitchbend"
 			"|Aftertouch|Channel Pressure|MIDI State|Poly|Omni|All Notes|All Sound"
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -783,6 +783,7 @@ class FxParams: public ParamSource {
 	bool (*_FormatParamValue)(ReaperObj*, int, int, double, char*, int);
 	bool (*_GetNamedConfigParm)(ReaperObj*, int, const char*, char*, int);
 	bool (*_SetNamedConfigParm)(ReaperObj*, int, const char*, const char*);
+	void (*_GetParamSectionName)(ReaperObj*, int, int, char*, int);
 
 	void initNamedConfigParams();
 
@@ -803,6 +804,8 @@ class FxParams: public ParamSource {
 			(apiPrefix + "_GetNamedConfigParm").c_str());
 		*(void**)&this->_SetNamedConfigParm = plugin_getapi(
 			(apiPrefix + "_SetNamedConfigParm").c_str());
+		*(void**)&this->_GetParamSectionName = plugin_getapi(
+			(apiPrefix + "_GetParamSectionName").c_str());
 		if (fx >= 0) {
 			this->initNamedConfigParams();
 		}
@@ -848,6 +851,14 @@ class FxParams: public ParamSource {
 		if (param < namedCount) {
 			// Named config params aren't FX params; keep them visible.
 			return true;
+		}
+		if (this->_GetParamSectionName) {
+			char section[100];
+			this->_GetParamSectionName(this->obj, this->fx, param, section,
+				sizeof(section));
+			if (strcmp(section, "MIDI") == 0) {
+				return false;
+			}
 		}
 		static const regex RE_UNNAMED_PARAM{
 			"(?:"
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -868,13 +868,6 @@ class FxParams: public ParamSource {
 			R"(|\d{1,4} -)"
 			// Example: "123" or "P123" or "#1234"
 			R"(|[P#]?\d{3,4})"
-			// Any one of several strings...
-			"|(?:MIDI CC|MIDI Controller|Program Change|CC|Pitch Bend|Pitchbend"
-			"|Aftertouch|Channel Pressure|MIDI State|Poly|Omni|All Notes|All Sound"
-			R"(|Local Control|X \(Reserved\)|Internal|Registered Parameter Number)"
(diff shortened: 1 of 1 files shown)


jcsteh:
This doesn't deal with the ReaControlMIDI case yet. I still need more details on how to distinguish these from the useless "CC" parameters in some other plugins.

jcsteh:
I think I've dealt with the ReaControlMIDI case now.

@XHG20033, could you please test the build here with all the cases you mentioned? If you don't have the plugins in question, could you please ask users who reported the issue to you to test it? Thanks.

@ScottChesworth, you may wish to take this for a spin for any tricky cases as well. This is now a kinda merge between your non-automatable approach and the regexp approach.

ScottChesworth:
Nice, liking this more than any of my prior tries. I checked EZKeys as that one was from community feedback. Its MIDI params are still being filtered, format is like "MIDI CC 0|64" and there are a full set of those per MIDI channel all exposed. Given that people would already be familiar with automation to make any use of those, I don't see unchecking a box as being a huge problem here.
Since last I checked, Toontrack have started prepending their own numbers and a period for forced pauses before params. Urgh. No idea who told them that was smart but I'll try to get it taken out.

jcsteh:
@XHG20033, does this latest build fix these other cases you mentioned?

> Keyscape (Spectrasonics): The plugin contains many parameters named with “Spec” followed by numbers (e.g., Spec 0 to Spec 1000). Users report these as unusable parameters, and it is suggested to add a filtering rule for them. Stylus RMX (Spectrasonics): The plugin contains purely numeric parameters (currently observed ranging from 74 to 127) and multiple parameters explicitly named “unnamed”.

XHG20033:
After installing PR1402, I loaded a plugin on a track and pressed P to open the FX Parameters dialog. It prompted:
        
        ---------------------------
        Microsoft Visual C++ Runtime Library
        ---------------------------
        Runtime Error!
        
        Program: C:\Program Files\REAPER (x64)\reaper.exe
        
        
        
        This application has requested the Runtime to terminate it in an unusual way.
        Please contact the application's support team for more information.
        
        
        ---------------------------
        OK   
        ---------------------------
        
        REAPER 7.74,. Windows 11 25H2

jcsteh:
Bah, sorry about that. Try this new build.

XHG20033:
Purely numeric parameters could be filtered starting from single digits, as the VST2 version of Keyscape actually has single-digit numeric parameters. I think that should cover everything else. Huge thumbs up to you for the incredible efficiency!
A friend who helped with the testing suggested whether it might be possible to allow users to add their own custom filtering rules, including exact parameter names (like "unnamed") or regex matching (like spec\d{1,3}). However, I personally feel this might not be the best approach. It's not a technical limitation, but rather the potential for trouble down the line—users who don't fully understand how it works might input bad rules and end up with unexpected parameter filtering behavior.
Moreover, I believe that most of these irregular, unusable parameters are usually listed after the valid ones. They don't really interfere with the operation of the valid parameters, so there's no absolute need to eradicate them all. Just as you mentioned at the very beginning, if we filter too aggressively, we inevitably risk filtering out actually useful ones. After all, this is primarily name-based filtering rather than a definitive, precise judgment.
What do you guys think?

jcsteh:
Thanks for the testing and feedback. Yeah, I think supporting custom filter rules here will complicate both the user interface and the implementation a lot, and I don't think the extra complication is worth the rare cases where it will be useful. I do appreciate the constructive input though.

PR #1404, opened 2026-06-23
Handle sws configured seconds actions

ScottChesworth:


ScottChesworth:
Jamie can you check the refactor commit? These SWS actions needed to be included in move from play cursor config and I couldn't see a cleaner way to do it.

src/reaper_osara.cpp
```diff
@@ -3336,15 +3338,27 @@ map<int, string> POST_COMMAND_MESSAGES = {
 	{40491, _t("all tracks unarmed")}, // Track: Unarm all tracks for recording
 	{42467, _t("all delta solos reset")}, // FX: Clear delta solo for all project FX
 };
-const set<int> MOVE_FROM_PLAY_CURSOR_COMMANDS = {
+// SWS/Xenakios actions which should also move from the play cursor when
+// settings::moveFromPlayCursor is enabled. These only have named command IDs,
+// so they are resolved to numeric IDs and inserted into
+// MOVE_FROM_PLAY_CURSOR_COMMANDS at init (see plugin_register handling below).
+const char* MOVE_FROM_PLAY_CURSOR_CUSTOM_COMMANDS[] = {
```
jcsteh: nit: Please move this below the original set, rather than above it.

src/reaper_osara.cpp
```diff
@@ -3336,15 +3338,27 @@ map<int, string> POST_COMMAND_MESSAGES = {
 	{40491, _t("all tracks unarmed")}, // Track: Unarm all tracks for recording
 	{42467, _t("all delta solos reset")}, // FX: Clear delta solo for all project FX
 };
-const set<int> MOVE_FROM_PLAY_CURSOR_COMMANDS = {
+// SWS/Xenakios actions which should also move from the play cursor when
+// settings::moveFromPlayCursor is enabled. These only have named command IDs,
+// so they are resolved to numeric IDs and inserted into
+// MOVE_FROM_PLAY_CURSOR_COMMANDS at init (see plugin_register handling below).
```
jcsteh: There are probably other ways to do this, but I think this is fine.

src/reaper_osara.cpp
```diff
@@ -3336,15 +3338,27 @@ map<int, string> POST_COMMAND_MESSAGES = {
...
+const char* MOVE_FROM_PLAY_CURSOR_CUSTOM_COMMANDS[] = {
+	"_XENAKIOS_MOVECURRRIGHTCONFSECS", // Xenakios/SWS: Move cursor right configured seconds
+	"_XENAKIOS_MOVECURRLEFTCONFSECS", // Xenakios/SWS: Move cursor left configured seconds
+	nullptr,
```
jcsteh: You should be able to remove this nullptr in combination with the change I'm suggesting below. If it doesn't work, though, this is fine.

src/reaper_osara.cpp
```diff
@@ -3336,15 +3338,27 @@ map<int, string> POST_COMMAND_MESSAGES = {
...
+};
+
+// Not const: the SWS actions above are added at init once their numeric IDs are
```
jcsteh: Obviously, this comment will need to be tweaked (or removed) once you move things around.

src/reaper_osara.cpp
```diff
@@ -6538,6 +6552,13 @@ void delayedInit() {
 		postCommandsMap.insert(make_pair(cmd, POST_CUSTOM_COMMANDS[i].execute));
 	}
 
+	for (int i = 0; MOVE_FROM_PLAY_CURSOR_CUSTOM_COMMANDS[i]; ++i) {
```
jcsteh: I think you can change these two lines to something like this:

```
	for (const char* command : MOVE_FROM_PLAY_CURSOR_CUSTOM_COMMANDS) {
		int cmd = NamedCommandLookup(command);
```

This kind of thing wasn't possible back when I originally wrote this area of the code, which is why we still have the annoying nullptr hack there.

That said, I'm not 100% sure this will work. If the compiler barfs, what you have here is fine.

jcsteh (approved):
Thanks. Approved with suggestions addressed or at least tried.

PR #1405, opened 2026-06-24
Suppress automatic reporting while a ReaScript action runs

chmaha:
**Problem:** Scripts that drive a sequence of actions/API calls currently get OSARA's full automatic narration for every recognized state change along the way — toggle messages, mute/solo/arm/volume/pan reports, etc. This competes with and buries messages a script reports explicitly via `osara_outputMessage`.

**Change:** Adds a new exported ReaScript API, `osara_suppressAutomaticMessages(bool suppress)`, that a script can call to pause OSARA's automatic reporting around a section it wants to handle quietly itself, e.g.:

```lua
reaper.osara_suppressAutomaticMessages(true)
-- ... a bunch of Main_OnCommand calls / track changes the script wants to be quiet about ...
reaper.osara_suppressAutomaticMessages(false)
reaper.osara_outputMessage("Setup complete")
```

Calls nest (each true increments a counter, each false decrements it), so it's safe even if the calling script itself calls other scripts/actions that also toggle suppression. `osara_outputMessage` is never affected by this — explicit messages always get through.

**Design notes:**
- This is purely opt-in. Nothing changes for any existing script or action unless it explicitly calls the new function, so there's no risk to scripts that rely on OSARA's existing automatic narration today.
- Implementation gates the single internal `outputMessage()` function that nearly all of OSARA's own reporting funnels through. `osara_outputMessage` now calls a new `outputMessageFromApi()` that bypasses the suppression gate (but still respects the existing `muteNextMessage` one-shot mute).
- No changes to `handleCommand`/`handleToggleCommand` or any other action-dispatch logic.

**Testing:** Not yet built on Windows/Mac — happy to do so, or for a maintainer to sanity-check the approach first.

### DIFF
--- a/src/exports.cpp
+++ b/src/exports.cpp
@@ -14,7 +14,7 @@
 // The _vararg_ version is needed for ReaScript.
 
 void osara_outputMessage(const char* message) {
-	outputMessage(message);
+	outputMessageFromApi(message);
 }
 void* _vararg_osara_outputMessage(void** args, int nArgs) {
 	osara_outputMessage((const char*)args[0]);
@@ -23,6 +23,14 @@ void* _vararg_osara_outputMessage(void** args, int nArgs) {
 
 bool osara_isShortcutHelpEnabled() {
 	return isShortcutHelpEnabled;
+}
+
+void osara_suppressAutomaticMessages(bool suppress) {
+	setAutomaticMessageSuppression(suppress);
+}
+void* _vararg_osara_suppressAutomaticMessages(void** args, int nArgs) {
+	osara_suppressAutomaticMessages((uintptr_t)args[0] != 0);
+	return nullptr;
 }
 
 void osara_getVersion(char* versionOut, int versionOut_sz) {
@@ -46,6 +54,18 @@ void registerExports(reaper_plugin_info_t* rec) {
 	rec->Register("API_osara_isShortcutHelpEnabled",
 		(void*)osara_isShortcutHelpEnabled);
 	rec->Register("API_osara_outputMessage", (void*)osara_outputMessage);
+	rec->Register("API_osara_suppressAutomaticMessages",
+		(void*)osara_suppressAutomaticMessages);
+	rec->Register("APIvararg_osara_suppressAutomaticMessages",
+		(void*)_vararg_osara_suppressAutomaticMessages);
+	rec->Register("APIdef_osara_suppressAutomaticMessages",
+		(void*)"void\0bool\0suppress\0"
+		"Suppress or resume OSARA's automatic reporting of state changes (e.g. "
+		"track mute/solo/arm, volume, pan, toggle actions).\n"
+		"Call with true before a sequence of actions or API calls you want to "
+		"perform quietly, and false when done. Calls may be nested; make sure "
+		"every true call is matched by a false call.\n"
+		"This has no effect on messages sent explicitly via osara_outputMessage.");
 	rec->Register("APIvararg_osara_getVersion",
 		(void*)_vararg_osara_getVersion);
 	rec->Register("APIdef_osara_getVersion",
--- a/src/osara.h
+++ b/src/osara.h
@@ -278,6 +278,8 @@ extern int lastCommand;
 bool shouldReportTimeMovement() ;
 void outputMessage(const std::string& message, bool interrupt = true);
 void outputMessage(std::ostringstream& message, bool interrupt = true);
+void outputMessageFromApi(const std::string& message, bool interrupt = true);
+void setAutomaticMessageSuppression(bool suppress);
 
 // Call a function or lambda (even a lambda with capture) asynchronously after
 // the specified number of ms. The function must take no parameters and return
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -137,9 +137,14 @@ void _outputMessage(const string& message, bool interrupt) {
 #endif // _WIN32
 
 bool muteNextMessage = false;
+int automaticMessageSuppressionDepth = 0;
+
 void outputMessage(const string& message, bool interrupt) {
 	if(muteNextMessage && isHandlingCommand){
 		muteNextMessage = false;
+		return;
+	}
+	if (automaticMessageSuppressionDepth > 0) {
 		return;
 	}
 	_outputMessage(message, interrupt);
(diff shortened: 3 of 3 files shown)


jcsteh:
On one hand, I understand the use case here. I'm a little less concerned about a toggle like this when provided as a ReaScript API vs an action. With ReaScript, there's a greater assumption that the author has some understanding of how code works and a user can't accidentally engage it.

On the other hand, as Scott noted in email, I'm pretty concerned about the potential for this to be harmful for users, even if entirely unintentionally (which is actually the most likely scenario). All it takes is for a script author to forget to disable the suppression at all, forget to disable the suppression for an early return, or even hit some error state which causes the script to abort such that the suppression isn't disabled. Bam, the user has no accessibility. We can say "but ReaClassical won't do that", "good scripters won't do that", "that would be the script author's fault", etc., but at the end of the day, it still hurts the user and we have a responsibility to mitigate that as much as we can.

I have a counter proposal, but I don't have time to look into it just now. What if we had some way for a ReaScript to say "suppress OSARA output while this ReaScript is running"? Maybe we should even just make it suppress OSARA handling of actions altogether; no point running OSARA code which calculates output if we're just going to throw that output away.

Implementing that would be tricky, but maybe not impossible. OSARA already has a command hook. The question is how we would know when a ReaScript has finished, since we currently return false for any commands we don't handle and we wouldn't know ahead of time whether the script wanted to suppress OSARA output. I guess we could instead run KBD_OnMainActionEx like we do for handlePostCommand, which would mean we'd absolutely know when an action had finished and could disable the suppression. But that might break things for some actions, so I'd prefer to do it only for ReaScripts... but how do we know whether an action id belongs to a ReaScript?

ScottChesworth:
> What if we had some way for a ReaScript to say "suppress OSARA output while this ReaScript is running"? I'd prefer to do it only for ReaScripts... but how do we know whether an action id belongs to a ReaScript?
> 
Looks like those start with _RS then a hash. Is that too fragile though?

ScottChesworth:
> I guess we could instead run KBD_OnMainActionEx like we do for handlePostCommand, which would mean we'd absolutely know when an action had finished and could disable the suppression.
> 
Dangling over the edge of what I'm sure about here but I had a hunch that might not hold for deferred scripts. Something I've run into before is that there's no callback for "deferred script has defo finished". I had a future robot overlord examine the PR and got this summary:

Deferred scripts would break the auto-clear. A script using reaper.defer returns from KBD_OnMainActionEx almost immediately and keeps running later on REAPER's defer loop. OSARA clearing suppression "when the action returns" would clear it TOO EARLY, and the deferred continuation would run with reporting already back on, defeating the purpose.

ScottChesworth:
What about this? More from helpful robot after a few rounds of fragile approaches. It looked hopeful so I've already tested to check reporting does behave as described.

Proposal: self-limiting heartbeat

Make suppression expire on its own unless the script keeps proving it's alive, so silence is always bounded no matter how the script ends, protecting against forgotten release, early return, crash or deferred lifecycle.

- Script re-arms suppression once per defer cycle, "I'm still here, stay quiet". OSARA's Run() checks a small internal grace period (a couple of cycles) then when the re-arms stop, it resumes reporting. During testing resume should be reported, then once solid, optional quiet resume would be useful.
- Individual scripts can't set a timeout to prevent misuse. The grace period is an OSARA constant. Stop re-arming -> speech returns. Abnormal exits fail safely, sidestepping both of the problems Jamie raised. The detect/intercept plumbing is gone.
- Caveat: a loop-less synchronous batch has no heartbeat, so it can't be watchdogged mid-run (OSARA's Run() doesn't tick while REAPER is blocked). Those scripts re-arm per chunk instead or accept a bounded tail.

Here are steps I tested:
1. Enabled "Report markers and regions during playback" in OSARA config.
2. Added project markers along the first 20 seconds of the timeline.
3. Playing the project while a deferred script ran for 15 seconds, OSARA kept reporting markers throughout. This confirms OSARA's Run() DOES tick between defer cycles, so the heartbeat proposal is viable.
4. Re-testing with a script that blocks for 8 seconds in a single cycle, marker reporting stalled when I ran the script and resumed when expected. That confirms OSARA's Run() doesn't tick inside a synchronous batch, illustrating the caveat above.

Thoughts?

jcsteh:
I did think of deferred scripts. However, those wouldn't work with the original proposal here of suppressing output at the start of the script and disabling suppression at the end either. Unless the plan was to call it at the start and end of every cycle?

jcsteh:
The other thing I thought of earlier is having the suppress API set a CallLater with a timeout of 0 which disables suppression. That should mean that the call only runs once REAPER returns to pumping window messages. The problem is that this will break if the script brings up a modal dialog, since modal dialogs spin up a nested message pump and the CallLater will probably run inside that. But now I'm realising this will also break the timeout approach your AI friend described.

jcsteh:
Maybe it's okay to say that you have to call the suppress API again after a modal dialog?  Hmm. There are ways to detect if you're inside a modal dialog on Windows, but I don't know if we can do that reliably cross-platform.

chmaha:
I agree that my current PR is putting too much trust in developers doing the right thing. FYI, ReaClassical contains both ordinary actions and long-running daemons, so any suppression mechanism would need to support both.

Another possible approach is a more explicit “token” model:
```lua
token = reaper.osara_beginSuppression()
reaper.osara_refreshSuppression(token)
reaper.osara_endSuppression(token)
```
Potential benefits (especially if combined with expiry/lease semantics):
- Multiple scripts suppressing independently
- Clear ownership semantics between callers
- Better supports nested usage patterns
- Recovery from crashes or forgotten cleanup when combined with expiry idea
- Avoids global state ambiguity

chmaha:
With the token approach, the modal dialog case in a deferred script still seems problematic. What would define a “safe” timeout before OSARA considers suppression inactive in the absence of refresh calls?

jcsteh:
So the big issue here seems to be detecting modal dialogs. I had a brainwave and realised that modal dialogs usually disable other windows. If we do a CallLater, that will run in the modal dialog loop. But we can detect that a modal is being displayed by using `if (!IsWindowEnabled(mainHwnd))`. In that case, we reschedule the CallLater until that condition is no longer true. So we have a single API suppress call which:

1. Sets a flag.
2. Schedules a CallLater with a timeout of 0 (next tick).
3. The callback function checks if mainHwnd is disabled. If it is, it reschedules the callback for 50ms or so. Otherwise, it clears the suppression flag set in (1).

This needs to be tested on Mac though. I'm not sure if SWELL emulates this window disable case correctly.

I also think we need to re-evaluate the name of the API function, but let's figure out whether this is even feasible first.

PR #1407, opened 2026-06-26
Report fit items to time selection

ScottChesworth:
Report the action we have mapped on Control+Alt+F and other similar ones


ScottChesworth:
Hey Jamie, figured I'd put this up for review because I'm using a tuple for the first time, seemed like an efficient way to store all the things we need to keep track of for reporting these actions.

readme.md
```diff
@@ -149,8 +149,9 @@ Most of these are actions built into REAPER, but a few are useful actions from t
 - Item edit: Shrink left edge of items: Control+. or Control+NumPad6
 - Item edit: Shrink right edge of items: Alt+, or Alt+NumPad4
 - Item edit: Grow right edge of items: Alt+. or Alt+NumPad6
-- Item edit: Move right edge of item to edit cursor: Control+Alt+Shift+.
-- Item edit: Move position of item to edit cursor: Control+Alt+Shift+,
+
```
jcsteh: Why the blank line?

src/reaper_osara.cpp
```diff
@@ -5843,6 +5843,45 @@ void cmdJumpToTime(int command) {
 	}, 50);
 }
 
+void cmdMoveAndFitItemsToTimeSelection(int command) {
+	int selectedItemsCount = CountSelectedMediaItems(nullptr);
+	if (selectedItemsCount == 0) {
+		outputMessage(translate("no selected items"));
+		return;
+	}
+	vector<tuple<double, double, double>> itemData(selectedItemsCount);
```
jcsteh: This doesn't really matter in practice in this case, but strictly speaking, it's not optimal. It's creating a vector with 0-initialised elements and then replacing them. Instead, I would do this:

```
	vector<tuple<double, double, double>> itemData;
	itemData.reserve(selectedItemsCount);
	for (int i = 0; i < selectedItemsCount; ++i) {
		...
		itemData.emplace_back(position, rate, length);
```

src/reaper_osara.cpp
```diff
@@ -5843,6 +5843,45 @@ void cmdJumpToTime(int command) {
...
+	int itemsChanged = 0;
+	for (int i = 0; i < selectedItemsCount; ++i) {
+		MediaItem* item = GetSelectedMediaItem(nullptr, i);
+		double position = GetMediaItemInfo_Value(item, "D_POSITION");
+		double length = GetMediaItemInfo_Value(item, "D_LENGTH");
+		MediaItem_Take* take = GetActiveTake(item);
+		double rate = 0;
+		if (take)
+			rate = GetMediaItemTakeInfo_Value(take, "D_PLAYRATE");
+		if (position != std::get<0>(itemData[i]) // Get position from tuple
```
jcsteh: I have a slight preference for deconstructing the tuple, rather than using std::get. That said, it actually ends up being an extra line of code this way and whether it's more readable is probably debatable, so I don't feel super strongly about it.

```
auto [oldPos, oldRate, oldLen] = itemData[i];
if (position != oldPos ||| ...
```

src/reaper_osara.cpp
```diff
@@ -5843,6 +5843,45 @@ void cmdJumpToTime(int command) {
...
+				|| rate != std::get<1>(itemData[i]) // Get rate from tuple
+				|| length != std::get<2>(itemData[i])) // Get length from tuple
+			++ itemsChanged;
```
jcsteh: nit: Please brace this if block.

jcsteh (approved):
Thanks. Approved with comments considered/addressed.

jcsteh:
Note that I made a typo in my vector comment that I just corrected, so that won't show up in your email if you're just reading there. :)

PR #1409, opened 2026-06-27
Add the ability for ReaScripts to suppress messages from OSARA while they run.

jcsteh:
This is an alternative to #1405. Once osara_suppressMessagesFromOsara() is called, messages are suppressed until the script finishes, taking modal dialogs into account.

This is completely untested at this point.

CC @chmaha.

### DIFF
--- a/src/exports.cpp
+++ b/src/exports.cpp
@@ -2,7 +2,7 @@
  * OSARA: Open Source Accessibility for the REAPER Application
  * Exported API functions code
  * Exports functions to be called by REAPER extensions and scripts.
- * Copyright 2020-2023 James Teh
+ * Copyright 2020-2026 James Teh
  * License: GNU General Public License version 2.0
  */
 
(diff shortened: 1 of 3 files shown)


jcsteh:
Things that need testing here: normal ReaScripts, defer ReaScripts, modal dialogs shown within ReaScripts, Windows, Mac, combinations of the above. I'm especially unsure about modal dialogs on mac.

jcsteh:
Note to self: Add to commit message before merging:
`Co-authored-by: chmaha <reaclassical@gmail.com>`

chmaha:
In initial tests, it seems to be working for normal reascripts and defer scripts. BTW, I'm placing it at the very top of the script outside of the defer cycle and seems to be fine. Is that the intention? Also I seem to be able to place a single line in my announce lua library and given that the file is required at the top of any reascript that outputs custom OSARA messages, it seems to block automatic messages in the original reascript automatically.

jcsteh:
> I'm placing it at the very top of the script outside of the defer cycle and seems to be fine. Is that the intention?

Hmm, no. That suggests that OSARA isn't disabling suppression when the main script exits. I expected you would have to do this at the top of the defer function so it runs for each cycle.

I guess I'll need to write a defer script myself to test this.

PR #1414, opened 2026-06-30
Fix send enumeration with hardware outputs

### LINKED ISSUE #1413: ALT+P, problems in orbital parameters
Hello, I found a small problem. When one track is sent to another track and the hardware is output to an audio card channel at the same time, at this time, in the track parameters of alt plus p, use the up and down cursor to navigate to the focus of the hardware output, reaper it will flash back. I don't know if it is reaper problem, or Osara's problem, because this situation is from version 7.55 reaper at the beginning, and there is also a problem with the report. Normally, it should be the track name transmission volume, such as 2 reverberation transmission volume. As a result, it is now read as 0 transmission volume. Thank you.

ScottChesworth:
Since REAPER 7.75 ("extend various TrackSend APIs to allow accessing 
sends/hardware outputs with UI ordering"), the category 0 (sends) index 
space used by GetSetTrackSendInfo includes hardware outputs, ordered 
before the real sends to match the routing UI. GetTrackNumSends(track, 0), 
however, still returns only the number of real sends. 
blank
OSARA's Track Parameters dialog iterated that index space as 0..sendCount, 
so as soon as a track had a hardware output, OSARA read hardware-output 
slots where it expected sends. Those slots have no destination track, so 
sends were mislabelled (e.g. "0 send volume" instead of "2 send volume"), 
and selecting the destination audio channel parameter passed a null 
destination track to GetSetMediaTrackInfo, crashing REAPER. This was not 
reproducible without a hardware output present, or before 7.75. 
blank
Now we walk the combined category 0 index space (bounded by sends + hardware 
outputs), skip entries with no destination track since those are hardware 
outputs already enumerated separately, and add each real send at its true 
index. On REAPER versions before 7.75, and on tracks with no hardware 
outputs, no entries are skipped and the loop behaves exactly as before, so 
there is no behaviour change there.

I've verified this is stable and consistent on REAPER 7.74, 7.75 and 7.76.

Fixes #1413

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1605,13 +1605,32 @@ class TrackParams: public ReaperObjParamSource {
 	template<int category>
 	void addSendParams(const char* trackParam) {
 		int count = GetTrackNumSends(track, category);
+		// Since REAPER 7.75, the category 0 (sends) index space used by
+		// GetSetTrackSendInfo also includes hardware outputs, ordered before the
+		// real sends to match the routing UI. GetTrackNumSends(track, 0), however,
+		// still returns only the number of real sends. We therefore walk the index
+		// space, skipping hardware outputs (which have no destination track and are
+		// enumerated separately as category 1 below), until we've found every send.
+		// On REAPER versions before 7.75, category 0 contains no hardware outputs,
+		// so this behaves like a simple 0..count iteration.
+		int maxIndex = count;
+		if constexpr (category == 0) {
+			maxIndex += GetTrackNumSends(track, 1);
+		}
 		string lastTarget;
 		int sameTargetCount = 1;
-		for (int i = 0; i < count; ++i) {
+		for (int i = 0, found = 0; i < maxIndex && found < count; ++i) {
 			string target;
 			if (trackParam) {
 				// Send or receive.
 				MediaTrack* sendTrack = (MediaTrack*)GetSetTrackSendInfo(this->track, category, i, trackParam, nullptr);
+				if constexpr (category == 0) {
+					if (!sendTrack) {
+						// A hardware output occupying a slot in the sends index space.
+						// Skip it here; it is enumerated as a hardware output below.
+						continue;
+					}
+				}
 				auto trackNum = (int)(size_t)GetSetMediaTrackInfo(sendTrack,
 					"IP_TRACKNUMBER", nullptr);
 				char* trackName = (char*)GetSetMediaTrackInfo(sendTrack, "P_NAME", nullptr);
@@ -1723,6 +1742,7 @@ class TrackParams: public ReaperObjParamSource {
 			this->params.push_back(make_unique<TrackSendParamProvider>(
 				paramName(translate("send type")), this->track, category, i, "I_SENDMODE",
 				SendTypeParam::make));
+			++found;
 		}
 	}
 
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -1435,11 +1435,6 @@ class AudioChannelParam:  public ReaperObjParam {
 	private:
 	Param::AfterOption addChannels(int count) {
 		MediaTrack* track = this->getTargetTrack();
-		if (!track) {
-			// No target track (e.g. a hardware output, or a send REAPER reports with
-			// no destination track). There's nothing to add channels to.
-			return Param::AfterOption::nothing;
-		}
 		int channels = *(int*)GetSetMediaTrackInfo(track, "I_NCHAN", nullptr);
 		channels += count;
 		GetSetMediaTrackInfo(track, "I_NCHAN", (void*)&channels);
@@ -1453,21 +1448,6 @@ class SourceAudioChannelParam: public AudioChannelParam {
 			AudioChannelParam(provider) {
 		options.push_back({translate_ctxt("audio channel", "none"), -1});
 		MediaTrack* srcTrack = this->getTargetTrack();
-		if (!srcTrack) {
-			// No source track. We don't know how many source channels there are, so
-			// just expose the current setting alongside "none".
-			int src = *(int*)provider.getSetValue(nullptr);
-			if (src != -1) {
-				if (src & MONO_FLAG) {
-					this->options.push_back({fmt::format("{}", (src & ~MONO_FLAG) + 1), src});
-				} else {
-					// Multi-channel, but we don't know how many.
-					this->options.push_back({fmt::format("{}-", src + 1), src});
-				}
-			}
-			this->finishOptions();
-			return;
-		}
 		int channels = *(int*)GetSetMediaTrackInfo(srcTrack, "I_NCHAN", nullptr);
 		this->addMonoOptions(channels);
 		this->addStereoOptions(channels);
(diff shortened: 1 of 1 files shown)


yyqh01:
Hello, this version of Osara, when navigating in the Track Parameters dialog box, it does not flash back reaper yeah, but, uh, in Scripts, in the script, the script related to obtaining the track written by lua will report an error when it is run. Bad argument #1 to'GetMediaTrackInfo Value' (MediaTrack expected)

jcsteh:
I'm really surprised Cockos made a backwards incompatible API change like this. Hmm.

src/paramsUi.cpp
```diff
@@ -1435,6 +1435,11 @@ class AudioChannelParam:  public ReaperObjParam {
 	private:
 	Param::AfterOption addChannels(int count) {
 		MediaTrack* track = this->getTargetTrack();
+		if (!track) {
```
jcsteh: Do we still need this? As I understand it, your later code means that we don't add hardware outputs as sends any more, so this should never happen. Or is there some way you can get a send without a destination that isn't a hardware output?

src/paramsUi.cpp
```diff
@@ -1448,6 +1453,21 @@ class SourceAudioChannelParam: public AudioChannelParam {
 			AudioChannelParam(provider) {
 		options.push_back({translate_ctxt("audio channel", "none"), -1});
 		MediaTrack* srcTrack = this->getTargetTrack();
+		if (!srcTrack) {
```
jcsteh: As above.

src/paramsUi.cpp
```diff
@@ -1473,17 +1493,18 @@ class DestAudioChannelParam:  public AudioChannelParam {
...
 		int srcChans = *(int*)sendProv.getSetValue("I_SRCCHAN", nullptr) >> 10;
 		if (srcChans == 0) {
 			srcChans = 2;
 		} else if (srcChans > 1) {
 			srcChans *= 2;
 		}
-		if (srcChans == -1) {
-			// If no source audio channel is set, we don't know how many destination
-			// channels there are. Just expose the current setting.
+		if (!dstTrack || srcChans == -1) {
```
jcsteh: As above.

yyqh01:
So this is reaper is there an API problem for? Also hope to help feedback, thank you.

src/paramsUi.cpp
```diff
@@ -1583,13 +1605,32 @@ class TrackParams: public ReaperObjParamSource {
...
+		// GetSetTrackSendInfo also includes hardware outputs, ordered before the
+		// real sends to match the routing UI. GetTrackNumSends(track, 0), however,
+		// still returns only the number of real sends. We therefore walk the index
+		// space, skipping hardware outputs (which have no destination track and are
+		// enumerated separately as category 1 below), until we've found every send.
+		// On REAPER versions before 7.75, category 0 contains no hardware outputs,
+		// so this behaves like a simple 0..count iteration.
+		int maxIndex = count;
+		if constexpr (category == 0) {
+			maxIndex += GetTrackNumSends(track, 1);
```
jcsteh: I wonder if a slightly cleaner way to do this could be to count the number of hardware outputs, check if the first send is a hardware output, and if so, set the start index to the number of hardware outputs and increase the count accordingly. Something like this:

```
		int i = 0;
		if constexpr (category == 0) {
			const int hwCount = GetTrackNumSends(track, 1);
			if (hwCount > 0 && !GetSetTrackSendInfo(this->track, category, 0, trackParam, nullptr) {
				// REAPER >= 7.75 includes hardware outputs in send category 0. We want to start at the first send.
				i += hwCount;
				count += hwCount;
			}
			...
			for (; i < count; ++i) {
```

I don't feel strongly about this though. Thoughts?

yyqh01 (changes requested):
12

ScottChesworth:
> I'm really surprised Cockos made a backwards incompatible API change like this. Hmm.

I didn't think it was backward incompatible based on testing in situ. Where would the rub be.

jcsteh:
It's backward incompatible precisely because it breaks any code that isn't updated to compensate for it. If it were backward compatible, the code wouldn't need to be updated. Anyway, it's academic because we have a fix, but it just really surprises me; they don't normally make changes which break existing code that was using the previous API the way it was intended.

src/paramsUi.cpp
ScottChesworth: Ah yeah, those guards are dead code now. Probably should've removed those before submitting. Thanks for being patient.

src/paramsUi.cpp
ScottChesworth: I've kept the per-entry skip. Your approach is cleaner but - as I understand it with the help of my token stooge - less defensive. Probing index 0 and offsetting assumes hardware outputs are always contiguous at the front, which matches what we see today, but it's empirical rather than documented. seeing as it's pretty cheap to belt-and-brace against future changes, we may as well do it.

jcsteh: I'd counter-argue that although this is more defensive, it isn't really future proof. If Cockos can shove hardware outputs at the front, they could just as well shove something else in there that isn't a hardware output but also doesn't have a track for some reason, in which case we'd have to rewrite this anyway. Defensiveness can hide bugs and make them hard to track down, which can be worse than just letting them show up in a more natural way. But as I said, I don't feel strongly enough about it to force the point; there are downsides to both approaches for sure. IMO, the real problem is that this API is just really ugly now: it doesn't make sense to have a function which counts something and then have a similarly named function where the indexes don't actually correspond to the count.

src/paramsUi.cpp
```diff
@@ -1583,13 +1583,34 @@ class TrackParams: public ReaperObjParamSource {
...
-		for (int i = 0; i < count; ++i) {
+		for (int i = 0, found = 0; i < maxIndex && found < count; ++i) {
 			string target;
 			if (trackParam) {
 				// Send or receive.
 				MediaTrack* sendTrack = (MediaTrack*)GetSetTrackSendInfo(this->track, category, i, trackParam, nullptr);
+				if constexpr (category == 0) {
+					if (!sendTrack) {
+						// A hardware output occupying a slot in the sends index space.
+						// Skip it here; it is enumerated as a hardware output below.
```
jcsteh: nit: "below" makes it look like we're going to do it below in this function, but it's really an entirely separate call to this function later after we return from this call.

```suggestion
						// Skip it here; it is enumerated as a hardware output in a subsequent call to this function.
```

jcsteh (approved):
Thanks.

jcsteh:
I'll let you merge this. I don't have time just now to come up with a sensible commit message, LOL. Thanks for fixing it.

src/paramsUi.cpp
ScottChesworth: Gotcha. Electric minion oversold the defensive rationale. I'm gonna do your suggestion seeing as that'll be easier to troubleshoot issues in future.

jcsteh:
I think this API change was reverted in REAPER 7.77. Either it wasn't intentional in the first place or they realised it broke too many things. The question is when we want to revert this in OSARA. I guess it might make sense to wait a month or so in case folks updated to 7.76 and don't update to 7.77 immediately. I do think we should revert it at some point; it isn't doing any harm, but it is kinda confusing and dead code is best avoided.

yyqh01:
-- 对象闪避 - 最终修复版

ScottChesworth:
> I guess it might make sense to wait a month or so... 

Agreed. When we revert it needs a clear change log entry flagging 7.76 as a problematic release.

PR #1416, opened 2026-07-05
Make paste/duplicate track and item added messages fully translatable

### LINKED ISSUE #1410: String concatenation for "added" causes grammar and context issues in Spanish translation
Hi, OSARA Team!
I’m reaching out with another translation request (I know, I know, but I’m trying to keep OSARA as polished as possible for the Spanish users! Haha). Yes, I know perfectly well that this standalone added string has been there for years, but now that I've seen we've advanced so much more with the arrival of new translation strings and improvements, I decided to finally report it and open this new issue, hoping it can be resolved.
Right now, OSARA concatenates the added string at the end of dynamic reports, like {} track + added, or {} tracks + {} items + added.
In English, "added" works perfectly for everything, but in Spanish, we have a double issue:
1.
Context/False Information: When a user copies and pastes tracks and/or items using Ctrl+V, we temporarily translated added as "desde el portapapeles" (from clipboard) so NVDA would make sense (e.g., "1 track from clipboard" or combined as "2 tracks 2 items from clipboard"). However, we realized that when a user duplicates tracks or items at the same time (Alt+D or Ctrl+D), OSARA triggers the exact same concatenated string. This results in NVDA saying "2 tracks 2 items from clipboard", which is false information since the clipboard wasn't used for duplication.
2.
Gender Agreement: If we change the string back to the literal translation "agregado" (masculine added), it clashes with "track" (pista, feminine) and NVDA would say "1 pista agregado". The exact same happens in reverse: if we use "agregada" (feminine), it clashes with "item" (ítem, masculine) and NVDA would say "1 ítem agregada" or "2 ítems agregadas".
Taking advantage of this, I wanted to ask: is the added string used to report the creation of other things besides items and tracks?
My Request:
Is it possible to stop using the standalone added string and instead provide full format strings for these actions? Something like:
{} track added
{} tracks added
{} item added
{} items added
{} tracks, {} items added
And just to bother you a bit more (I hope it's not too much to ask! haha), wouldn't it be better to also create specific strings that report "duplicated" exclusively for when tracks or items are duplicated, separating them from the add/paste action?
This would allow us to perfectly adjust the gender, plural, and context rules for each specific scenario directly in Crowdin.
Let me know if this is feasible or if it’s too complex to implement at this stage. Thanks for your infinite patience with me and for all your help!
Hugs!

munzzyy:
Fixes #1410.

When pasting/duplicating produces both new tracks and new items, `cmdPaste` built the message by gluing together two bare counted phrases ("2 tracks" + "3 items") and then tacking a single shared "added" on the end. As ElCrackLoko pointed out in the issue, that doesn't work for languages with grammatical gender (Spanish "agregado"/"agregada" can't agree with both a masculine and a feminine noun in the same word) or more complex plural systems.

This applies the fix James and Scott settled on in the thread: each part becomes its own complete, translatable phrase ("2 tracks added", "3 items added"), joined with ", " when both are present. So in English you now get "1 track added, 3 items added" instead of "1 track 3 items added". This is the same join pattern already used for the combined mute/solo/arm/phase messages elsewhere in this file, so it should translate the same way those do.

Only touches `cmdPaste` - the automation item/envelope point/take branches further down and `cmdSplitItems` (which only ever reports items, never tracks) weren't affected by this bug and are untouched.

I don't have the Reaper SDK or the build submodules set up, so I wasn't able to actually compile this. I checked it carefully against the existing patterns in the file (the mute/solo/arm code does exactly this two-phrase-joined-by-comma thing already) and it only touches string literals and one separator character, but a build/smoke test on your end would be good before merging.

### DIFF
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4196,7 +4196,7 @@ void cmdPaste(int command) {
 	Main_OnCommand(command, 0);
 	int added;
 	// We want to report both tracks and items if both got added. Each part is
-	// a full, independently translatable phrase (e.g. "1 track added, 3 items
+	// a full, independently translatable phrase (e.g. "1 track added", "3 items
 	// added") because concatenating counted nouns doesn't work for languages
 	// with grammatical gender or complex plural forms; see #1410.
 	ostringstream s;
--- a/src/reaper_osara.cpp
+++ b/src/reaper_osara.cpp
@@ -4195,27 +4195,25 @@ void cmdPaste(int command) {
 	}
 	Main_OnCommand(command, 0);
 	int added;
-	// We want to report both tracks and items if both got added; e.g.
-	// "1 track 2 items added".
+	// We want to report both tracks and items if both got added. Each part is
+	// a full, independently translatable phrase (e.g. "1 track added, 3 items
+	// added") because concatenating counted nouns doesn't work for languages
+	// with grammatical gender or complex plural forms; see #1410.
 	ostringstream s;
 	if ((added = CountTracks(0) - oldTracks) > 0) {
-		// Translators: Reported when tracks are added. Other things might be added
-		// at the same time (e.g. items), so other messages may surround this.
-		// {} will be replaced with the number of tracks; e.g. "2 tracks".
-		s << format(translate_plural("{} track", "{} tracks", added), added);
+		// Translators: Reported when tracks are added. {} will be replaced with
+		// the number of tracks; e.g. "2 tracks added".
+		s << format(translate_plural("{} track added", "{} tracks added", added), added);
 	}
 	if ((added = CountMediaItems(0) - oldItems) > 0) {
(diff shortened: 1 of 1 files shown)


src/reaper_osara.cpp
```diff
@@ -4195,27 +4195,25 @@ void cmdPaste(int command) {
 	}
 	Main_OnCommand(command, 0);
 	int added;
-	// We want to report both tracks and items if both got added; e.g.
-	// "1 track 2 items added".
+	// We want to report both tracks and items if both got added. Each part is
+	// a full, independently translatable phrase (e.g. "1 track added, 3 items
```
jcsteh: nit: Clarify this comment slightly.
```suggestion
	// a full, independently translatable phrase (e.g. "1 track added", "3 items
```

jcsteh (approved):
Thank you.

PR #1418, opened 2026-07-12
Parameters dialog: When adjusting the slider, retry checking for a value change if the value doesn't change immediately.

jcsteh:
This is an alternative to #1227. It implements the idea I proposed in https://github.com/jcsteh/osara/pull/1227#issuecomment-4912240134.
@ScottChesworth , I'd be interested to get your thoughts on this with various obscure plugins.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -314,6 +314,7 @@ class ParamsDialog {
 	HWND prevFocus;
 	bool isDestroying = false;
 	bool suppressValueChangeReport = false;
+	CallLater valChangeLater;
 
 	void updateValueText() {
 		if (this->valText.empty()) {
@@ -357,6 +358,7 @@ class ParamsDialog {
 	}
 
 	void onSliderChange(double newVal) {
+		this->valChangeLater.cancel();
 		if (newVal == this->val
 				|| newVal < this->param->min || newVal > this->param->max) {
 			return;
@@ -365,7 +367,6 @@ class ParamsDialog {
 		if (newVal < val) {
 			step = -step;
 		}
-		this->val = newVal;
 
 		// If the value text (if any) doesn't change, the value change is insignificant.
 		// Snap to the next change in value text.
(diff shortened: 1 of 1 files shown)


jcsteh:
If this works, I'd like to get it merged and then pursue getting rid of the various value caches we have in this code, as I think they confuse things and probably don't have that much of a performance benefit. I don't want to do both at once, though, because that introduces too many variables regarding potential breakage.

ScottChesworth:
Nice! It solves issues I was seeing with sliders in Kontakt and KK. There's only one outlier situation I've found here where sliders are still obnoxious, but it's the same behaviour across all of the products I've got from one company, so I think that'll be something for them to tighten up, not us.
I think it's helping somewhat with older VST and VSTI sliders that ideally should be stepped, paging through those seems to find the next change in less presses now. Gonna pull the branch and see if I can find any tweaks to get that working in a single press per step. It's already a net improvement IMO, so if you're happy with this and ready to merge, I can follow up if my tinkering finds anything.

jcsteh:
> There's only one outlier situation I've found here where sliders are still obnoxious

Can you describe the obnoxious behaviour?

> I think it's helping somewhat with older VST and VSTI sliders that ideally should be stepped, paging through those seems to find the next change in less presses now.

Perhaps those plugins take a while to update the value or snap the value when you set it.

> Gonna pull the branch and see if I can find any tweaks to get that working in a single press per step.

The problem with those plugins is that they only support formatting the value for the current value, not for an arbitrary value. That means OSARA can't find the next snap value by continually incrementing a temporary value and formatting the temporary value until it gets a different string. The only way to fix that would probably be to actually set the value and then format it once set, rather than using a temporary value. But I imagine that would get really, really expensive and slow, not to mention creating a crap tonne of envelope points if you're automating. I guess maybe we can revert to that behaviour for plugins which appear to return a formatted value for the current value but not other values. There's also the question of how long we wait for the value to change each time we set it; I don't think we want to wait 300ms before trying the next value if we're incrementing 100 times.

ScottChesworth:
> Can you describe the obnoxious behaviour?
> 
Try a library from Impact Soundworks if you've got any. Their sliders won't budge. Best I can do here is Home or End will move them, but no feedback without refocusing the slider, and even then sometimes that lies. Seems to be the same in Kontakt or KK.

Issue #558, opened 2021-07-19
Is it possible to access the Send Envelopes more easily?


Lo-lo78:
Hello,
with the object navigator of NVDA it is possible, with a little attention, to arm the Send Envelopes.
The problem is that when you mix this procedure slows down the work.
Right now I do "shift + l", NVDA + 8 numpad, NVDA + 2 numpad, NVDA + 6 numpad and look for the Send Envelopes.
Is it possible to capture these automations and make a dialog box where the Send Envelopes are, where it is possible to arm or disarm them?
We with OSARA can arm the parameters FX, volume, pan and so on ...
But without NVDA the Send Envelopes cannot be armed.
I thought, for example:
track 1 with a send in track 2;
I select track 1 and with a shortcut linked to an action, a dialog box opens with all the Send Envelopes of the selected track that I can manage more quickly.

1 snare: combo box, "down arrow":
Track 2 "reverb" Send Volume Check box not activated 26 of 42
Track 2 "reverb" Send Pan Check box not activated 29 of 42
Track 2 "reverb" Send Mute Check box not activated 32 of 42

ok button

Obviously if there are two Sends from track 1 to track two it is logical that there will be two combo boxes.
It is also obvious that if there are multiple Sends from track 1 where you go into track 2 and track 3 there will be combo boxes describing the source track and the destination track.
Thanks James!

ScottChesworth:
I'd prefer to see Cockos improve keyboard navigation in Reaper than for OSARA to add another custom dialog. Do you agree, @jcsteh?

Lo-lo78:
Yes, Scott, actually if there were APIs on Reaper for that it would be 
ok. Also the FX Parameter dialog is from OSARA, but as you say Cockos 
better make send envelopes more accessible.

Greetings!

jcsteh:
It would be nice if Cockos improved keyboard navigation in the envelopes dialog, yes. That said, if I were going to do this in OSARA, rather than a new dialog, I'd do it in the Parameters dialog similar to my proposal for envelopes for FX parameters (#533). For example, you'd open the track parameters dialog, move to the send volume parameter and press the Show envelope button.

Lo-lo78:
Hi James, it's a very good way, for example:

2 reverb send volume; Value: progress slider 0.00dB;

Edit Selected 0.00dB;

Arm envelope: Value: progress slider off / on.

! 👍

Lo-lo78:
While you are there you can do this for all parameters in the Track 
[...]

PR #1419, opened 2026-07-12
Parameters dialog: For send volume/pan/mute parameters, add a context menu option to show/hide envelope.

jcsteh:
This has come up before in #558. We didn't end up doing it because REAPER's Envelopes dialog became accessible. Even so, this still grinds my gears every time I have to do it. To create a send volume envelope, the workflow goes something like this:

1. Press shift+l to open Track Envelopes.
2. Tab once to the filter text box.
3. Type: send
4. Tab twice to the envelope list.
5. Down arrow to the one you want. Alternatively, you could press t I guess. This gets more fiddly the more sends you have.
6. Open the context menu.
7. Down arrow to Active and press enter. Alternatively, type a and then press enter.

If we support this in the OSARA Parameters dialog, the workflow goes like this:

1. Alt+p.
2. First letter nav to the send. For example, if you have a send to track 2, just press 2.
3. Open the context menu.
4. Press e.

It's not really that much simpler, but it feels smoother or more intuitive to me for some reason.

@ScottChesworth, thoughts? Am I being ridiculous?

FX parameter envelopes don't bother me any more because moving to a parameter is enough to touch it, so you can just control+alt+l and you're done, no slider wiggling required.

### DIFF
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -312,7 +312,7 @@ class ParamsDialog {
 	double val;
 	string valText;
 	HWND prevFocus;
-	bool isDestroying = false;
+	bool shouldAllowDeactivate = false;
 	bool suppressValueChangeReport = false;
 
 	void updateValueText() {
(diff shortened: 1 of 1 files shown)


jcsteh:
@ScottChesworth, ping for your thoughts on this when you get a sec.

ScottChesworth:
Oh sorry, thought I'd already replied to this. I like it!

PR #1420, opened 2026-07-13
Parameters dialog: If a parameter only supports formatting the current value, try setting it and then formatting the value.

jcsteh:
This makes it possible to snap to the next or previous significant value with a single arrow key press with some older plug-ins.

@ScottChesworth, give this a try with your plug-ins that won't snap properly. But also, try it with other obscure plug-ins too, because I can't guarantee I haven't regressed anything, though I did take it for a spin with quite a few plug-ins here.

### DIFF
--- a/src/osara.h
+++ b/src/osara.h
@@ -226,6 +226,9 @@
 #define REAPERAPI_WANT_TimeMap_GetTimeSigAtTime
 #define REAPERAPI_WANT_GetSetProjectGrid
 #define REAPERAPI_WANT_GetTrackMIDINoteNameEx
+#ifdef LOGGING
+#define REAPERAPI_WANT_ShowConsoleMsg
+#endif
 
 #include <reaper/reaper_plugin.h>
 #include <reaper/reaper_plugin_functions.h>
@@ -379,5 +382,12 @@ IReaperControlSurface* createSurface();
 extern bool selectedEnvelopeIsTake;
 // exports.cpp
 void registerExports(reaper_plugin_info_t* rec);
+
+#ifdef LOGGING
+#define dbg(msg, ...) \
+	ShowConsoleMsg(fmt::format("osara " msg "\n", __VA_ARGS__).c_str())
+#else
+# define dbg(msg)
+#endif
 
 #endif
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -359,32 +359,67 @@ class ParamsDialog {
 
 	void onSliderChange(double newVal) {
 		this->valChangeLater.cancel();
-		if (newVal == this->val
-				|| newVal < this->param->min || newVal > this->param->max) {
-			return;
-		}
 		double step = this->param->step;
-		if (newVal < val) {
+		if (newVal < this->val) {
 			step = -step;
+		}
+		if (step < 0 && newVal < this->param->min) {
+			newVal = this->param->min;
+		} else if (step > 0 && newVal > this->param->max) {
+			newVal = this->param->max;
 		}
 
 		// If the value text (if any) doesn't change, the value change is insignificant.
 		// Snap to the next change in value text.
 		// Continually adding to a float accumulates inaccuracy, so multiply by the
 		// number of steps each iteration instead.
+		const string origText = this->param->getValueText(this->val);
+		const string minText = this->param->getValueText(this->param->min);
+		const string maxText = this->param->getValueText(this->param->max);
+		const bool canOnlyFormatCurrent = !origText.empty() &&
+			// Some parameters return an empty string when you query the formatted text
+			// for a value which isn't the current.
+			(minText.empty() || maxText.empty() ||
+			// Others return the current value's formatted text for all values.
+			(origText == minText && origText == maxText));
+		dbg(
+			"params slider: origVal {} origText '{}' newVal {} step {} "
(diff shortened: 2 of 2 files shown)


jcsteh:
Note that this won't work with plug-ins that both fail to format non-current values and also take a while to reflect a parameter change. Supporting both would mean waiting for the value to change each time we increment or decrement while trying to find the next value, which could be a thousand iterations. If we wait even 30 ms for the value to change each iteration, that's 30 seconds in total. So um, let's not.

ScottChesworth:
> @ScottChesworth, give this a try with your plug-ins that won't snap properly.
> 

Nice one! Everything I've got here to test it with feels like VST3. I've asked for suggestions of more plug-ins to check in the WhatsApp group, will report back when I've tried more.

jcsteh:
Anyone got a slow machine they can test one of these plugins on with this build? Repeatedly spamming value sets like this is pretty horrible, so I'm slightly concerned it'll crawl on slow machines. That said, I'm not seeing any noticeable perf hit on a fast machine, which hopefully means it shouldn't be too bad on a slow one.

jcsteh:
I looked into why this doesn't help with Impact Sound Works plugins. The only ones I have here are Shreddage, but I can reproduce it well enough. It strikes me as particularly odd because they're Kontakt instruments and I don't see this in any other Kontakt instrument. Do you, @ScottChesworth ?

These plugins seem to suffer from three problems:

1. They can't format values except for the current value.
2. Worse, instead of just not supporting it so that REAPER returns an empty string, they say they support it... and then just return the same text regardless of what numeric value you specify. That means the code in this PR never triggers. There's no way we can tell the difference between the snapping case, where multiple numeric values map to the same formatted text, and this broken case, where every numeric value maps to the same formatted text until you change the value.
3. Worse still, they seem to take some time before they update the formatted value to the correct text for the newly set value. We don't have any way of knowing that we need to wait for that or whether it's just the snapping case.

Unfortunately, I don't think there's anything I can do for these. They're just broken in too many ways.

What's interesting is that these seem to work with NKS for me. I guess NKS doesn't have to worry about finding the next snap value; it just clicks over when the knob is turned far enough.

ScottChesworth:
Have you got any Valhalla DSP plug-ins there Jamie? Sliders in their VST3s are hosed in this PR, also broken in latest snapshot. Oops. Didn't notice because I'm mostly still running those as VST. They offer a couple of freebies if you've got time to investigate, or failing that, any pointers on how to troubleshoot appreciated.

jcsteh:
Take them for a spin with this build. It'll dump a bunch of crap to the ReaScript Console. Please gimme that.

ScottChesworth:
[VST3 broken Shift slider Valhalla DSP FreqEcho.txt](https://github.com/user-attachments/files/30165091/VST3.broken.Shift.slider.Valhalla.DSP.FreqEcho.txt)
[VST working Shift slider, arrows page keys home and end.txt](https://github.com/user-attachments/files/30165093/VST.working.Shift.slider.arrows.page.keys.home.and.end.txt)
Logs attached. Hopefully they help. Holler if you need more.

jcsteh:
Are you sure these logs aren't truncated at the start? There should be a line containing "origVal" and "origText", but the first thing I see is:

> osara params slider: newVal 0.755 testText '-1000.00 Hz'

jcsteh:
Regardless, I don't understand how this could have worked with the old code. This plugin does not support formatting values other than the current. It's similar to Impact Sound Works products. The VST2 version at least returns an empty string, so OSARA can detect it. The VST3 version doesn't.

Are you absolutely certain both versions were totally fine with the old code?

ScottChesworth:
> Are you sure these logs aren't truncated at the start?
> 
Zarqing fardbarx, guess the log scrolled and I forgot to check.

> Are you absolutely certain both versions were totally fine with the old code?
> 
Totally fine would be overegging the pudding a bit, they used to move in tiny steps and it took a bunch of presses before you'd hear an updated report, so they were only usable with page keys, but yes, VST and VST3 did used to be workable the same way. Shall I make a script to handle logging so you can see what happened with an older snapshot?

jcsteh:
Oh, I think I see why it might have been different. That's... annoying. I'd still like a full log from the VST3 version if you can get it.

So VST2 now works as you'd like, no silent steps?

ScottChesworth:
> I'd still like a full log from the VST3 version if you can get it.
> 
Tried again more carefully and the first line in my console is still the format I uploaded earlier. Does the logging definitely output what you're expecting at your end?

> So VST2 now works as you'd like, no silent steps?
>
Way fewer silent steps. They're not entirely gone, but I'm only seeing them in a handful of clearly tricky cases now. I'd really like to keep the big win this implementation provides (properly making continuous sliders feel stepped in loads of older VST plug-ins that will never get updates). It's a big net UX improvement.

jcsteh:
Hmm, I wonder if the ReaScript console drops earlier messages if there are too many. That's... unhelpful.

jcsteh:
Can you try setting the value to, say, 0.9 with the text box, then pressing right arrow on the slider? That should hopefully give us less steps to deal with but still give me the data I need.

ScottChesworth:
Got it
[Broken continuous slider in VST3 Valhalla FreqEcho.txt](https://github.com/user-attachments/files/30175876/Broken.continuous.slider.in.VST3.Valhalla.FreqEcho.txt)

PR #1427, opened 2026-07-24
Provide contextual feedback in 3 commonplace scenarios where an item can't be split

### LINKED ISSUE #1384: Could OSARA provide more info about being unable to split items?
I've noticed some people are confused when they hit A and hear 0 items added. After many investigations, often that report seems to be happening in one of three commonplace causes:
1. There's no track selected, hence an item can't be split on selected track at cursor.
2. There's not an item on the selected track at the cursor position to split.
3. There's already a split in the item on the selected track at this cursor position.
All of these situations can be detected reliably, so I'm thinking of adding some additional feedback for clarity. Suggestions of reports that would strike a good balance between informational and brevity would be appreciated.

ScottChesworth:
Closes #1384


src/reaper_osara.cpp
```diff
@@ -4148,9 +4148,40 @@ void cmdRedo(int command) {
...
 	int oldCount = CountMediaItems(nullptr);
 	Main_OnCommand(command, 0);
 	int added = CountMediaItems(nullptr) - oldCount;
+	if (!added) {
+		double curPos = GetCursorPosition();
+		bool canSplit = false;
+		bool alreadyOnSplit = false;
+		for (int i = 0; i < CountSelectedTracks2(nullptr, false); ++i) {
+			MediaTrack* track = GetSelectedTrack2(nullptr, i, false);
+			for (int j = 0; j < CountTrackMediaItems(track); ++j) {
```
jcsteh: Let's at least use `t` (for track) and `i` (for item) here. i and j are nondescript.

src/reaper_osara.cpp
```diff
@@ -4148,9 +4148,40 @@ void cmdRedo(int command) {
...
+				MediaItem* item = GetTrackMediaItem(track, j);
+				double position = GetMediaItemInfo_Value(item, "D_POSITION");
+				double length = GetMediaItemInfo_Value(item, "D_LENGTH");
+				if (curPos > position && curPos < (position + length))
+					canSplit = true;
+				else if (position == curPos || curPos == (position + length))
+					alreadyOnSplit = true;
```
jcsteh, line 4173: nit: Please brace the bodies of both these ifs.

src/reaper_osara.cpp
```diff
@@ -4148,9 +4148,40 @@ void cmdRedo(int command) {
...
+			}
+		}
+		if (!canSplit && !alreadyOnSplit) {
+			// Translators: Reported when there is not an item under the edit cursor to be split.
+			outputMessage(translate("no item at cursor"));
+			return;
+		}
+		else if (!canSplit && alreadyOnSplit) {
```
jcsteh: nit: `else` after return is redundant. Drop the `else` to make it just an `if`.

jcsteh (approved):
Thanks. Just some nits.

PR #1429, opened 2026-07-28
Parameters dialog: Use a tree instead of a combo box.

jcsteh:


src/paramsUi.cpp
```diff
@@ -642,17 +661,13 @@ class ParamsDialog {
 			) {
 				return -666; // Force to main window.
 			}
-			if (msg->hwnd == dialog->paramCombo && msg->wParam == VK_SPACE) {
-				// In the combo box, we also pass space to the main section.
```
jcsteh, line 678: @ScottChesworth This raises an interesting question. Tree views allow you to do proper type ahead navigation, which is great. But that includes space. That's useful for parameters which contain spaces. On the other hand, it means that pressing space will no longer play/stop when focused in the parameter tree. Is that what we want?

jcsteh:
This could do with some testing, especially on Mac.

src/paramsUi.cpp
ScottChesworth: > pressing space will no longer play/stop when focused in the parameter tree. Is that what we want?
> 
I think it's fine. Typeahead is a navigation win, it's the sliders that need to pass Space so you can hear what you're adjusting in situ.

ScottChesworth:
Tagging some Mac users for testing: @GarthHum , @mulcahy1000 , @pitermach, @tbdalgaard, @vick08. Feedback from others is welcome of course.

Please keep all feedback/requests for support posted here in these comments so everything is trackable and details don't get missed.

pitermach:
Navigation is working well on the 27 beta. The tree is a million times faster than the popup menu ever was, filtering is also working as expected.

There is a regression when using CTRL+Tab/CTRL+Shift+Tab though which I suspect might also occur on the Windows version. The active parameter does change which you can see by looking at what's selected in the tree, but speech always reads the name of the first parameter of the plugin. So in the case of Surge which is what I tested on it always says "m1"

jcsteh:
Regarding control+tab, is this pressing control+tab while focused on the slider or the tree? I can't reproduce this in either case on Windows, which is... weird.

pitermach:
It happens with both, though I didn't test carefully enough and it turns out the issue is slightly different. The value slider always controls the first parameter, so even if you do change the selection either by selecting a different item in the table or using CTRL+Tab (which correctly highlights the next/previous parameter in the list as well) you're always controlling the first parameter.

jcsteh:
Okay. I think Mac's tree widget (or at least SWELL's usage of it) might not apply the selection immediately, whereas it does on Windows. I think I have an idea how to fix it, but it'll definitely need re-testing once I do, as I can't validate the fix here.

pitermach:
It now works well in build 464, but not 465.

jcsteh:
Err... that's weird. Same issue in 465 as the original build? Damn it I hate supporting mac.

jcsteh:
I can't understand why that would be the case. The part of the code relevant to that hasn't changed. aldskfjasld;kfajds

pitermach:
Hm. Well it turned out 464 wasn't working perfectly either - CTRL+Tab and Shift+Tab work, but selecting anything in the table always controlled the first parameter. In 465 CTRL+Tab also stopped working.

Because it's getting late and I wanted to hopefully give you something more helpful to work with I recorded a comparison on how it works in 464 vs 465, think  I didn't miss any details this time.
[osara bug.mp3](https://github.com/user-attachments/files/30565395/osara.bug.mp3)

Fwiw I seriously appreciate the effort put in to the Mac support.

jcsteh:
Hmm, that mp3 file seems to be rendered with speed boost or something? 😕 Anyway...

I added a bunch of logging. Could you please:

1. Edit reaper-extstate.ini in your REAPER data folder.
2. In the [osara] section, add: logging=1
3. Ensure you restart REAPER.
4. Select items in the tree, use control+tab, etc. Try to keep this fairly minimal, just reproducing the problems, so I have less log output to work through. :)

jcsteh:
Oh, and attach the log file in $TMP/osara.log. Thanks.

pitermach:
You'd think I'd remember that rendering with the playrate increased would keep the increased playrate in the resulting audio file, sorry about that.

Log is below. This was done with SurgeXT, I first down arrowed through the table, then put myself on the slider and CTRL+Tabbed a few times. Throughout all this the speech and slider was showing me the "m1" parameter rather than any of the later ones. I'm also going to send this to a few other people who use Mac regularly that aren't on the 27 beta, in case this is a factor, though other trees in the program including the preferences or FX browsers behave normally.


[osara.log](https://github.com/user-attachments/files/30596571/osara.log)

jcsteh:
Thanks; that was very enlightening. Try the next build.

pitermach:
Happy to report that did it, now both moving in the tree and using CTRL+Tab is working as expected.

One of the commits mentioned that parameter categories should appear as seperate branches, is that something which will be added later or should already be working? Because with this most recent build on both Mac and Windows in Surge I'm still seeing a flat tree, where if I check IE the show envelopes submenu under the param button in the FX dialog Surge is making extensive use of categories which appear as submenus there. Also, in the case of sends, they correctly appear as a branches on Windows containing the different send parameters, while on Mac I just see a "sends" item which doesn't expand to show anything. CTRL+Tabbing also doesn't move past the mute parameter for the send.

Heading off for the day but will test any new builds tomorrow if needed.

jcsteh:
I haven't implemented FX sections yet. Just trying to get sends working first, then I'll tackle FX parameters in a separate PR.

Damn it, I don't know why sends aren't working on Mac. That kinda defeats the whole point of this work. 😩

pitermach:
OK just caught this, and again happy to say that did it, now the sends show up and can be adjusted.

jcsteh:
I changed my mind: I added FX section name support here too. Let me know how that goes for you.

jcsteh:
@ScottChesworth , I think we're just about ready to merge this. All happy on your end?

pitermach:
Working great here on Mac, now the categories in Surge show up.

ScottChesworth:
> @ScottChesworth , I think we're just about ready to merge this. All happy on your end?
> 
Can we wait for a few days so a JAWS scripts update can land alongside it?

jcsteh:
Urgh, if it isn't Mac, it's JAWS. 🤦‍♂️ Sure. :)

vick08:
I know you guys know what you're doing, but is the reason for this enhancement described anywhere? I see this request says "no description" for some reason? Sorry if I missed something... I'd just love to understand why this change...

jcsteh:
The primary reason is that it allows for easier navigation of send parameters. If you have more than one send, it can be quite tedious to get from one send volume to another send volume, for example. It apparently also makes navigation of the parameter list much easier on mac in general; see #534. Finally, it also means that we can categorise parameters into sections for plug-ins which report section names for their parameters such as Surge.

ScottChesworth:
> Urgh, if it isn't Mac, it's JAWS.
> 
And JAWS is being stubborn on this one. Just checking in to let you know I haven't forgotten, testing and feedback is happening.

jcsteh:
No changes here, just tidying up the branch ready for merge when JAWS stops being an obstacle.

ScottChesworth:
Jim's close with JAWS now. I'm having him do a last bit of persuasion to get it reading the input and output channels nicer while the changes are top of mind. Guessing it'll be good to go sometime over the weekend.

jcsteh:
Err, why does he need to do any persuading at all? That should just work. All of this should just work out of the box.

ScottChesworth:
It should, but it doesn't. Last issue I've asked him to look at is JAWS splurting param numbers in a confusing way when control+tabbing through the new in/out channels.

jcsteh:
Just double checking: we definitely don't have any bugs like that with other screen readers? I'm not sure whether you mean the send channels or the FX pins, but the latter are in a different PR. I don't think I tested control+tab properly with the FX pins; should probably do that.

ScottChesworth:
Yeah I know FX pins are in a separate PR, figured I'd check those as well while there's scripting happening in the dialog, turned  out it did need some extra help to report those with the same style Snowman uses.
I've already checked Control+Tabbing in both PR's, both working fine.

jcsteh:
@ScottChesworth, would love your thoughts on all three of these latest commits.

@pitermach, does having the value in the tree behave on Mac? Because the value is added lazily when the item is first selected, it's possible it will cause double speaking of the parameter name. If that's the case, that's a show stopper and I will revert this.

ScottChesworth:
I can already tell it'll break the new JAWS scripting, you maniac LOL! Might negate the need for that dialog though. Will check soon.

jcsteh:
I got bored waiting for JAWS to stop blocking me, so I figured I may as well make it more complicated. 🤣

pitermach:
It's actually behaving itself, not seeing any double speech, the list also displays fine in braille. For what it's worth, I really like this change.

This is almost certainly going to need more tweaks from the JAWS scripts side though, because if I remember right its scripts made the parameter list behave like this from the start - IE reading the parameter values as you moved through the list.

ScottChesworth:
To be fair, scripts have been done for a couple days. I didn't push go yet because the last couple days have been busy, wouldn't have been able to watch the groups like an insomniac hawk until tomorrow.

jcsteh:
LOL. If you like, I can move these bits into a separate PR. I just figured that if we were blocked, I may as well make more progress here. But if we're not, we can push what we have and do the rest later.

jcsteh:
Surely the number stripping changes won't break JAWS?

pitermach:
I guess that would depend on how the scripts behave - IE did they already do stripping like this on their own? Perhaps in the worst case someone might get double value speaking which wouldn't be the end of the world, or it could get more confused.

It would probably be good to have someone with the pre-release scripts try this and see what happens, I don't have access to JAWS here right now

ScottChesworth:
Yeah I think it strips based on finding the parens. Either way, keep it in this PR, I'll check JAWS at factory defaults and if there's parity with NVDA/VO now, Snowman can remove his handling.

jcsteh:
Hmm, I guess the safest thing to do if we don't want to be blocked longer on JAWS for the tree stuff is to merge the PR as it was before i started faffing with it today (i.e. kill the last 3 commits), then spin up separate PRs for the other stuff. That's normally what I would have done, but it's a bit awkward because this PR is blocking several other changes.

ScottChesworth:
Keep it all in this PR. I checked with JAWS at default settings, reporting is exactly the same as NVDA. With Snowman's current and prerelease scripts the param value is appended twice, no serious breakage that I could find. I'll send him the test build from today and suggest dropping the scripting in the params dialog in his next update. I'll report back as soon as I get confirmation, and while you wait, I've found something else annoying for ya :)

ScottChesworth:
It's possible to get stuck in the tab order.
1. Hit Alt+P on a track or item.
2. Control+Tab and Control+Shift+Tab through params with focus in either the tree or on the sliders, then tab away, that all works as expected.
3. Now try that from an edit field instead. Jumping through params works, but when you try to tab away, you'll just get a Windows ding sound. At this point Escape doesn't work to close the dialog either, has to be Alt+F4.

jcsteh:
Hmm. Is this only if you control+tab to a toggle (like mute) while the text box is focused? I can't reproduce it otherwise. That happens because the text box gets disabled when you select a toggle. However, this isn't new. It also happens on master. Unless you're seeing a different issue to what I'm seeing?

jcsteh:
This is going to be tricky to fix here because the FX pin parameter pull request changes stuff around that area of code, so we'll end up with annoying merge conflicts. I guess I could merge that in here too, but... oof. This is what happens when stuff like this isn't merged for a while. :)

jcsteh:
Screw it, I'll just pull the FX pin stuff in here. This is already a kitchen sink, why not make it an entire kitchen...

jcsteh:
Bah, pin stuff conflicts with the changes I just made here today anyway.

ScottChesworth:
> Is this only if you control+tab to a toggle (like mute)
>
Ah that explains it. Sorry, I wrote STR before understanding what happens when a toggle gets focus. It is on master too of course but I've never noticed it until I went looking for ways to make JAWS mad today.

jcsteh:
If it's any consolation, I saw that bug for the first time while I was developing the tree support and never could work out how to reproduce it, so just put it down to Zarquan taking a juju-flop on me at the time.

MooSteinebach:
A little thing to consider:
If there's a checkBox in the parameters list, it's displayed as 0 or 1. I suggest to display it as "on" and "off" for better readability.
Example: Reacomp, auto release

jcsteh:
That's not always true. It depends on the plugin and only the plugin can specify this unfortunately.

jcsteh:
This one needed to be rebase merged. It's too complex to be a single commit on master.

ScottChesworth:
Oops! Looks like it built though. What did I break?

jcsteh:
I'm reverting it now. Nothing is broken; massive commits like that are just hard to reason about when you're trying to figure out how something was implemented or track down a problem.

ScottChesworth:
Oh so rebase merge keeps the individual commits visible. Got it. Will remember that if we end up with another biggun.

jcsteh:
Okay, all sorted. I've manually re-merged it as separate commits. I had to force push master though, so if you pulled it, you may need to reset your copy of master.

jcsteh:
Yeah. Rebase merge is only a good idea though if the branch has a very clean commit history; i.e. work-in-progress commits squashed out, commit messages tidy, etc. I did that with this one, but it takes a bit of work.

jcsteh:
That is, I rewrote the history several times when I cleaned things up. That's why you kept seeing force pushes here.

PR #1431, opened 2026-07-31
If a parameter only supports formatting the current value, don't try setting it and then formatting the value.

jcsteh:
This reverts the main behavioural change in #1420. While nice with some plug-ins, it was causing highly erratic behaviour with others.


jcsteh:
@ScottChesworth We shouldn't lose the gains in #1418 here; i.e. values shouldn't bounce about in some cases. Also, this should keep the Valhalla fix from #1420. However, both of those need to be verified.

ScottChesworth:
Valhalla VST3 full range sliders are holding up. Step detection now feels the same as VST2 and I miss it already! Is there any chance you'd be open to a "Detect stepped settings (best oeffort)" checkbox in the params dialog? IMO it worked well in enough cases to justify being on by default, we can document that there are still some rough edges, try disabling it if you can't move sliders as expected.

jcsteh:
The problem is that we already "Detect stepped settings"; we have for years. It just doesn't work for plug-ins which don't support formatting values other than the current. It'd have to be "Detect stepped settings for plug-ins which don't support returning friendly values for any value other than the current", which is a ridiculously confusing name. We could make it "Detect stepped settings for broken plug-ins", but that's ambiguous because there are two different types of brokenness here, one of which this fixes, one of which it breaks.

> Step detection now feels the same as VST2 and I miss it already!

I don't follow this. Do you mean the VST2 version of the same plug-in? The code in #1420 should have made step detection work there too, not just for VST3. Hmm.

jcsteh:
What about "Try setting value to detect steps"?

jcsteh:
@ScottChesworth That last commit should at least make value reporting more reliable in Impact Sound Works libraries, even if the slider is oddly jumpy. Also, unchecking the check box will mean that you can at least page up/page down through those ISW parameters.

jcsteh:
I think this plug-in parameters stuff is going to be the stuff of my nightmares from now on. If I have to touch this too much more, I think OSARA will need to find a new lead developer. :)

ScottChesworth:
This is great! seems ready to merge here.

ScottChesworth:
> I think this plug-in parameters stuff is going to be the stuff of my nightmares from now on. 
> 
I feel similarly about supporting Linux whenever that's a goer. It'll be time to take the money and run.

PR #1432, opened 2026-08-01
FX Parameters dialog: Support FX input and output channel mappings.

### LINKED ISSUE #114: Plugin Pin Connector
Hi, 

We've already spoken about this too but just thought I'd file it officially. Is there some way, either by talking to Cockos or something you can do on Osara's end, to make access to channel routing for effects plug-ins less laborious. 

Right now you have to open and close the context menu every time you want to enable or disable a channel send or receive. It must take significantly longer than it takes sighted people and is easy to make mistakes. 

Something like a tree view would be ideal, with collapsable channels and the ability to check and uncheck the ones you do and don't want. 

Thanks a lot, 

Justin

jcsteh:
There are now "inputs" and "outputs" branches in the tree after all other parameters. Each input/output is its own item inside the appropriate branch. For each input/output, you enter the numbers of the channels to map in the text box, separated by spaces.

Fixes #114.


jcsteh:
Merged into #1429.

PR #1433, opened 2026-08-02
Add OSARA: Merge OSARA key map action.

jcsteh:
This is very much work in progress and has only had very preliminary testing. Use at your own risk; make sure you back up your key map first.

This is also available from the Extensions menu -> OSARA -> Merge OSARA key map. It compares your key map against OSARA's bundled key map. Any new custom actions and script entries are added without asking. If there are any new key bindings that do not conflict, they are added without asking. If there are conflicts, a dialog is displayed with a checkable list view, with all items unchecked by default. For each conflict, the list includes the key, the action in OSARA's key map and the action in your key map. Checking the item accepts OSARA's action. Once you choose to proceed, the merged key map is written, you are asked to restart REAPER and REAPER exits. The original key map is backed up as reaper-kb.ini.osara-merge-backup just in case.


jcsteh:
An open question: there are certain keys like "y" which have a default REAPER binding but which the OSARA key map overrides so that they do nothing. These are currently listed as "Disabled default". Is this understandable or is there better wording?

jcsteh:
Mac strikes again.

jcsteh:
Zarq it, SWELL list views don't support check boxes.

ScottChesworth:
> SWELL list views don't support check boxes.
> 
Is that why state hasn't been possible to report in Prefs, General, Import/Export Configuration?

ScottChesworth:
Thanks heaps for working on this, looking forward to stress testing. BTW now it exists you're gonna have to stay as lead dev forever and ever to stop me from renaming it the Key Map Coagulator.

jcsteh:
> Is that why state hasn't been possible to report in Prefs, General, Import/Export Configuration?

I had that thought too, but couldn't remember which list it was. Yes, I'm fairly certain this is why. SWELL doesn't support LVS_EX_CHECKBOXES, so REAPER can't use it, and so the check boxes aren't accessible even on Windows.

pitermach:
Because I wanted a solution for preserving keymap changes easier I was very curious to try this change, and I'm really sorry about this, but it looks like it's not working on Mac 🫢

What I did:

- In the actions dialog, replace OSARA"s binding of O from the tempo markers back to the LFO generator, also tried binding space to play/pause in a second test.
- - Run the merge action

On Windows this correctly reports the merge conflict with O and lets me decide what to do with it. On Mac, OSARA always reports no conflicts and happily overrides the O binding back to the tempo markers. It didn't report a conflict when I reassigned space as well. I also tried restarting reaper before running the action, in case the ini file gets saved only when Reaper exits, but this didn't make a difference and on Windows the merge action picks up the conflict right away.

jcsteh:
Arrrgh! Thanks for testing though.

> On Mac, OSARA always reports no conflicts

Did you get the message "No key binding conflicts were found. {} key bindings will be added. Do you want to merge the OSARA key map?" Did it correctly report that some number of bindings would be added?

Can you please attach your reaper-kb.ini?

pitermach:
Yes, I did get the message, it wanted to import 700 ish commands and the import itself did work, it just overrode my assignment.

Here's the .ini file, it has O reassigned to the LFO generator and Space to play/pause.
[reaper-kb.ini.zip](https://github.com/user-attachments/files/30704334/reaper-kb.ini.zip)

jcsteh:
Thanks. Very perplexing. Please remove any existing osara.log and try the next build with logging. You don't need to say yes to the merge; it's enough just to run the action and say no when it asks if you want to proceed. Then provide the log here. Thanks.

pitermach:
Here you go.
[osara.log](https://github.com/user-attachments/files/30734736/osara.log)

jcsteh:
Thanks. Please try the next build.

pitermach:
Working perfectly now. The conflict dialog came up, VoiceOver read the state of each item and space correctly toggled an option. I intentionally told it to replace the play/pause binding back to the one for OSARA while leaving my assignment for the LFO generator and it did what I asked.

jcsteh:
In this PR, I want to just focus on getting the initial merge tool right, which should hopefully be done now minus wider testing. However, moving forward from here:

1. We can invent some way for OSARA to know when its key map has been updated since the last time the user merged. We can probably just use a version number like we do for optimal REAPER config.
2. When REAPER starts, OSARA can detect the update and offer to merge, again similar to optimal config.
3. We can remove key map overwriting from the installers.
4. This paves the way for OSARA to be distributed via a ReaPack repository for those who want that, since it only needs to distribute files and we don't need to run stuff outside of REAPER.
5. We can potentially look at supporting an extra layer of keys via the alt-1 section, etc. The names for those sections are stored outside the key map, but the merge tool can be taught to update those as well.

Issue #491, opened 2021-03-04
OSARA translation crash REAPER


Aleskis:
When pressing shift+F8 shift+F9 and shift+F10, translations corresponding to item mono mode crash REAPER. I see nothing about syntax error.
I created and attached a .zip with the french langpack and the .po file. Can someone help me figure out what's wrong?
Anyway, I find it very dangerous that a simple translation can crash the whole REAPER without a preamble. I don't know if in terms of code, there's something that can be done to protect the execution of the program, but if so, let's do it because if it happens to me, it will happen to others.
[OSARA French Translation.zip](https://github.com/jcsteh/osara/files/6084960/OSARA.French.Translation.zip)

ScottChesworth:
First step, can any of our other French-speaking users reproduce this?

windrofy:
I tested that the files provided by your compressed package did not crash.
osara version is osara_2021.1pre-483,22e726e4.
reaper version is reaper624_x64-install.

jcsteh:
As I [explained to you on Twitter](https://twitter.com/jcsteh/status/1366834245956562945):

> The problem is, for example, "set {count} takes to {mode}". {count} and {mode} need to be left untranslated.

So, this string:

set {count} take to {mode}

has been translated as:

{Comptage} de prise réglé sur {mode}

You've translated the "count" in the braces, but that needs to be left alone. As the translator comment explains:

> {count} will be replaced with the number of takes affected.

Anything in braces is a placeholder and needs to be left alone, though you can move it wherever you need to. The translators comment always explains what the placeholders mean.

Regarding your other questions:

> I see nothing about syntax error.

I see you haven't uploaded your translation to Crowdin. Crowdin has a QA feature which will detect errors like this.

> Anyway, I find it very dangerous that a simple translation can crash the whole REAPER without a preamble. I don't know if in terms of code, there's something that can be done to protect the execution of the program, but if so, let's do it because if it happens to me, it will happen to others.

Protecting against this is tricky. After hours of work, I think I've come up with a way to prevent the crash, but all you're going to get is a message like "error in translated string". I can't provide anything more specific than that.

I can probably set up some automation to detect bad translated strings while syncing them from Crowdin and reject the translation, though the question is how to notify translators at that point.

Aleskis:
Thank you James, didn't saw comment translation on Poedit. Anyway, I corrected the translation and it works. Thanks also for the work to try to exclude translations that could crash REAPER.

jcsteh:
[...]

PR #1437, opened 2026-08-05
Strip the directory (if any) from the REAPER lang pack when determining the OSARA translation to use.

### LINKED ISSUE #1435: langpack detection for Osara's translation
If you download a langpack from Reaper's download page, they are named like
deutsch.reaperlangpack
for e.g. german.
Osara of course searches for de_DE.
So you have reapers GUI in german, but not Osara's.
What about just a table wich matches those strings like
deutsch=de_DE
for the most common languages?

jcsteh:
Normally, the REAPER lang pack is just specified as a file name. However, in some cases, it is a full path, which previously prevented OSARA from matching it against the correct OSARA language. Now, we ensure we're matching against the file name only.

Fixes #1435.

### DIFF
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -79,8 +79,13 @@ void initTranslation() {
 	}
 	// We can't use std::filesystem::path because it isn't supported until MacOS 10.15. Grrr!
 	string name(langpack);
+	// REAPER normally gives us a file name, but may return the full path.
+	const auto dirPos = name.find_last_of("\\/");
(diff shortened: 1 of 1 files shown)


jcsteh:
@MooSteinebach A build should appear here a few minutes after I write this. Can you please test this build and let me know whether the problem is fixed? Thanks.

MooSteinebach:
Tested #491, all works fine. Problem fixed!

PR #1439, opened 2026-08-06
Fix loading of OSARA translation when REAPER is set to prompt for language on load.

### LINKED ISSUE #1204: Problem with loading translation when prompt on load option is selected
Hi,
The translation of Osara does not load if prompt on load option is selected in preferences.
Thanks.

jcsteh:
Fixes #1204.

Completely untested at this stage. If anyone wants to take this for a spin, that'd be great. CC @reaperaccessible, @MooSteinebach.

### DIFF
--- a/src/translation.cpp
+++ b/src/translation.cpp
@@ -70,10 +70,20 @@ static istringstream filterPoAmpersands(istream& input) {
 void initTranslation() {
 	// Figure out which file name to load. We base it on the REAPER language
 	// pack.
-	char langpack[200];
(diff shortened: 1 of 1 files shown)


MooSteinebach:
works perfect, Osara always loads the correct translation!
Tested:
- englisch, load once and load always
- german, load once and load always
- with and without a path for langpack= in reaper.ini, 
for all possible combinations.

jcsteh:
Thank you for testing.

Issue #1436, opened 2026-08-04
update german translation


MooSteinebach:
Hi James,
I'd like to update the german translation for Osara.
Crowdin user: MooSteinebach
Language: german
Reaper langpack: de_DE or german

jcsteh:
Thanks for your offer of help. I prefer to get agreement from existing translators for a language before I add new ones, as otherwise, we can end up with disputes between translators.

@michaelvogt33, are you happy for @MooSteinebach to update the German translation?

MooSteinebach:
@jcsteh:
Do you think Michael will get back to you shortly?

PR #1440, opened 2026-08-08
Make the decimal point translatable.

### LINKED ISSUE #1438: Make separator character translatable when reporting numbers
This request came from @MooSteinebach in the REAPER Access WhatsApp group:

In germany and some other countries, the comma is used as the decimal point. When OSARA always uses period, at the moment it speaks times like
two minutes, fortysixthousandsevenhundrettwentyfive. seconds

Moo asked whether we could add an option to swap comma and period in time announcements, but I think it would be better to make the separator character translatable instead?

jcsteh:
In English, the decimal point is ".". However, in some other languages, the decimal point is different; e.g. "," in German. Using the incorrect decimal point can cause confusing pronunciation of numbers. To deal with this, make the decimal point translatable. Note that REAPER itself still always uses the English decimal point. Therefore, any values that are formatted by REAPER or which might be passed to REAPER later still use the English decimal point; e.g. the text box in the Parameters dialog.
Fixes #1438.

### DIFF
--- a/src/controlSurface.cpp
+++ b/src/controlSurface.cpp
@@ -136,8 +136,7 @@ class Surface: public IReaperControlSurface {
 			s << translate("volume") << " ";
 			this->lastParam = PARAM_VOLUME;
 		}
-		s << fixed << setprecision(2);
-		s << VAL2DB(volume);
+		s << formatDouble(VAL2DB(volume), 2);
 		outputMessage(s);
 	}
 
--- a/src/osara.h
+++ b/src/osara.h
@@ -350,7 +350,13 @@ const char* getActionName(int command, KbdSectionInfo* section=nullptr, bool ski
 
 bool isTrackSelected(MediaTrack* track);
 bool isTrackArmed(MediaTrack* track);
-std::string formatDouble(double d, int precision, bool plus=false);
+
+// Format a double d to precision decimal places.
+// If plus is true, a "+" prefix will be included for a positive number.
+// If stripZeros is true, trailing zeroes will be removed.
+std::string formatDouble(double d, int precision, bool plus=false,
+	bool stripZeros=true);
+
 MediaItem* getItemWithFocus();
 
 #ifdef _WIN32
--- a/src/paramsUi.cpp
+++ b/src/paramsUi.cpp
@@ -321,9 +321,7 @@ class ParamsDialog {
 			// Fall back to a percentage.
 			double percent = (this->val - this->param->min)
 				/ (this->param->max - this->param->min) * 100;
-			ostringstream s;
-			s << fixed << setprecision(1) << percent << "%";
-			this->valText = s.str();
+			this->valText = formatDouble(percent, 1) + "%";
 		}
 #ifdef _WIN32
 		// Set the slider's accessible value to this text.
--- a/src/peakWatcher.cpp
+++ b/src/peakWatcher.cpp
@@ -395,7 +395,6 @@ bool isWatchingMultipleValues() {
 
 void CALLBACK tick(HWND hwnd, UINT msg, UINT_PTR event, DWORD time) {
 	ostringstream s;
-	s << fixed << setprecision(1);
 	const bool multiple = isWatchingMultipleValues();
 	auto& projWatchers = watchers[currentProject()];
 	for (int w = 0; w < NUM_WATCHERS; ++w) {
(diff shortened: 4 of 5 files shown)


jcsteh:
@MooSteinebach Give this a try. You'll need to add this to OSARA's locale/de_DE.po:

```
#. The decimal point symbol.
msgctxt "decimal"
msgid "."
msgstr ","
```

I can't commit that change because it will just be overridden next time translations are updated on Crowdin. So, we'll need to get #1436 sorted before we can fully fix this for German.

MooSteinebach:
Works absolutely correct with german translations loaded.
Tested:
- speaking time in project (ctrl+shift+j) - correct, two minutes, 26 comma 426 Milliseconds
- speaking position while scrubbing - correct
- values of textboxes in "jump to time", "item position", "Parameter dialog" are spoken correctly as e.g. "one period 000"
- moving item to a new position (ctrl+alt+shift+comma) - correct

If you scrub by 10 ms through the project you get e.g.
0 minutes, 0 comma 09 seconds
0 minutes, 0 comma 1 seconds?
0 minutes, 0 comma 11 seconds
I guess trailing zeros are being cut off, so it maybe confusing not to get
09 ms, 10 ms, 11 ms
but
09 ms, 1 ms, 11 ms

I would say getting results like
2 minutes, 26 comma 400 seconds
would be absolutely fine.

jcsteh:
> 09 ms, 1 ms, 11 ms

I'm not sure I entirely agree. It's not 1ms; it's 0.1 seconds. Those two things are quite different. I thought the whole point here was to make the pronunciation of decimal numbers more correct, so this shouldn't be confusing.

That said, it's a change to the behaviour we had before, so I guess it's simpler just to go back to what we had there.

jcsteh:
Next build restores the trailing zeros for fractional seconds.

MooSteinebach:
Sorry, May I did'nt explain it correctly.
Take an empty project, press ctrl+j and enter
0:02.100
Before build 498 Osara said:
0 minutes, twothousandonehundret seconds, or in digits
0 minutes, 2.100 seconds.
With build 500, same procedure, the speech output is:
0 minutes, 2.1 seconds, without the trailing zeros.
The correct output, after changing the de_DE-po file, should be
0 minutes, 2,100 seconds, instead of 
0 minutes, 2,1 seconds.
That was, what I'm trying to say. So the routine making the decimal point translatable cuts of the zeros at the end of the string.

jcsteh:
I understand that. My point is that 2.1 and 2.100 (assuming the English decimal point) are the same number. That is basic decimal math. This is the same reason we just say 0dB instead of 0.000dB or 6.5dB instead of 6.500dB.

In any case, the last build doesn't strip trailing zeros for seconds any more. Let me know if it works as you expect.

MooSteinebach:
The decimal point is translated correctly with build 500.
Tested with time announcements as well as parameter dialog and adjusting volume of tracks by pressing alt-up/down arrow.
What I want to mention is:
in the last official build, as in all builds before, the following happened (talking about period as the decimal point):
If you have an empty project and press page down once, Osara said:
0 minutes, 2.000 seconds
In build 500, same test, Osara says:
0 minutes, 2 seconds
Both are of course correct, I only want to say that users my become confuseed, if the trailing zeros in a time announcement are no longer there.
Build 500 still strips the trailing zeros.
But, the trailing zeros were only spoken when announcing the time, e.g. 0 minutes, 2.000 seconds, when adjusting the volume of a track with alt+up/down arrow, the output was e.g.
-0.95
-1
So, in my opinion, if the time announcement is like in build 500
0 minutes, 2 seconds
that would be okay and even more consistent.

jcsteh:
Bah, sorry, I screwed up build 500. Try the next one, which should prevent stripping trailing zeros for seconds.

MooSteinebach:
Yep! Build 501 works absolutely as expected!
Tested with german langPack loaded:
- time announcement (ctrl+shift+j): 0 minutes, 2 comma 010 seconds
- adjusting track volume (alt+up/down arrow): -0,9, -0,95, -1 - as expected
- jump to time (ctrl+j): the contents of the textfield is spoken as 0:00.000 as expected
- routing parameters (i): contents of text fields spoken correctly as -4.50 as expected
- FX-parameter dialog: sliders and edit boxes use still decimal point (period) for notation, which is absolutely correct here.

In my opinion you've got it!

